PR #20, opened 2017-04-24
Cpp library gluecode

gantech:
This pull request adds a FAST - C++ interface and a sample driver program to illustrate the use of this class.  This mostly does not modify the existing code except for some additions to the InitOut data of Aerodyn and ElastoDyn modules.


sayerhs:
@gantech Can you merge with `dev` and then resubmit? Right now there are conflicts with merge.

gantech:
Ok. Will try. I think this is due to the fact that I forked off master initially and master being ahead of dev. This pull request merges cleanly with master.

sayerhs:
`master` and `dev` were synced, so the only conflict is in the `beamdyn/CmakeLists.txt` should be a minor fix.

sayerhs:
@gantech Did you fix CMake rules for `beamdynlib`? It seems to have broken some rules for export.

gantech:
Thanks @sayerhs. This is odd. I only resolved the conflicts I saw on github. 

@michaelasprague Could I add to the documentation in another pull request. I'm focused on getting the capability to work with Nalu to run OWEZ soon.

Also.. I have a few more mods based on a discussion with Matt Churchfield yesterday. I'll get those in today as well. Then it should be ready to merge.

PR #23, opened 2017-05-04
Regression Test

rafaelmudafort:
This pull request adds a simple regression test borrowing heavily from @jrood-nrel's Nalu regression test and @gantech's fastCPPtests.

The test configuration is handled with CTest, so the test is only built if the appropriate CMake flag, `ENABLE_TESTS`, is `ON`. 

The regression test itself is driven by `reg_tests/executeRegressionTestCase.py` which runs the OpenFAST test case and compares the binary output to the "gold standard" with `reg_tests/pass_fail.py`. The pass/fail criteria is a comparison of the variance of each channel in the binary output file with a predetermined tolerance. The tolerance is automatically picked by CMake based on machine and compiler type. 

The test data is included in a submodule called RegressionTestData which based on the CertTest directory in FAST v8. This repository has two subdirectories: `inputs` and `outputs`. The contents of `inputs` are used in CTest configuration and the contents of `outputs` are the "gold standard" solutions. There is also a python script `runCertTestsLocally.py` which executes all of the cases in `inputs` in a new directory `outputs-local`.

To run the test from a fresh clone:
`git clone https://github.com/rafaelmudafort/openfast.git`
`cd openfast`
`git checkout dev`
`git submodule update --init --recursive`
`mkdir build`
`cd build`
`ccmake .. # or otherwise configure cmake flags`
-- hit c, then e
-- turn on the test boolean, configure the install prefix, etc
-- hit c, then e
-- again hit c, then e 
-- hit g
`make`
`make test`

Improvements to come:
- determine appropriate machine/compiler specific tolerances
- add support in ctest for selecting a target reference directory
  - this would allow a developer to generate their own "gold standards" in case of new feature implementation or improvement
- cron job & CDash


rafaelmudafort:
@ghaymanNREL Yes, the submodule should be merged into the openfast repo. However, the gold standards currently in my RegressionTestData repo were generated on my local computer. I now have access to peregrine so I can create the results there. After getting those, I'll update RegressionTestData so it can be moved to openfast and update .gitmodules.

rafaelmudafort:
@ghaymanNREL I've updated the `dev` branch on my openfast fork so that it now includes the `openfast/r-test` repository as a submodule rather than my `rafaelmudafort/r-test` repository

sayerhs:
@rafaelmudafort What architecture/compiler are the _gold_ files generated for?

rafaelmudafort:
@sayerhs the gold standard files in the master and develop branch in `r-test` are macos-gnu. There is a branch called `feature/goldstandard-expansion` which currently contains gold files for macos-gnu and rhel-intel, if you're interested. I'm working on a corresponding openfast branch to make use of various gold standard options in the regression test.

PR #28, opened 2017-05-15
Bug fixes to C++ API

gantech:
Bug fixes related to setting of velocity at the actuator force nodes, interpolation of velocity from actuator force to Aerodyn nodes and including the turbine base location in the coordinates returned to the driver program.


gantech:
Thanks. I have updated the code to use GetNewUnit() in the latest commit.

michaelasprague:
@ghaymanNREL - any objections to merging?

PR #37, opened 2017-07-26
fix compiler warnings regarding too many arguments for c print statements

rafmudaf:
The `fprintf(...` statements in the fast registry module generator have too many arguments. In most cases, no arguments were needed but `r->name` was given.

### DIFF
--- a/modules-local/fast-registry/src/gen_module_files.c
+++ b/modules-local/fast-registry/src/gen_module_files.c
@@ -419,13 +419,13 @@ gen_pack( FILE * fp, const node_t * ModName, char * inout, char *inoutlong )
     if (has_deferred_dim(r, 0)){
          // store whether the data type is allocated and the bounds of each dimension
   fprintf(fp, "  IF ( .NOT. %s(InData%%%s) ) THEN\n", assoc_or_allocated(r), r->name);
-  fprintf(fp, "    IntKiBuf( Int_Xferred ) = 0\n", r->name); // not allocated
-  fprintf(fp, "    Int_Xferred = Int_Xferred + 1\n", r->name);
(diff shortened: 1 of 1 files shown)


rafaelmudafort:
I think the input argument in these lines was a mistake. This is one of the modified blocks with the lines in question denoted by `<-----`:
```
fprintf(fp, "  Re_Xferred  = 1\n");
fprintf(fp, "  Db_Xferred  = 1\n");
fprintf(fp, "  Int_Xferred = 1\n\n");

// Pack data
for ( r = q->fields ; r ; r = r->next ) {
  if (has_deferred_dim(r, 0)){
    // store whether the data type is allocated and the bounds of each dimension
    fprintf(fp, "  IF ( .NOT. %s(InData%%%s) ) THEN\n", assoc_or_allocated(r), r->name);
    fprintf(fp, "    IntKiBuf( Int_Xferred ) = 0\n", r->name); // not allocated   <----- 
    fprintf(fp, "    Int_Xferred = Int_Xferred + 1\n", r->name);   <-----
```
In this case, the generated code is using a variable `Int_Xferred` which was initialized just above the `fprintf` lines I modified. 

In another block
```
fprintf(fp, "      IF(ALLOCATED(Re_Buf )) DEALLOCATE(Re_Buf )\n", r->name);
fprintf(fp, "      IF(ALLOCATED(Db_Buf )) DEALLOCATE(Db_Buf )\n", r->name);
fprintf(fp, "      IF(ALLOCATED(Int_Buf)) DEALLOCATE(Int_Buf)\n", r->name);
```
these arrays are to be deallocated, so the `r->name` argument doesn't have a place here.

It seems like someone copied and modified another line but forgot to remove the extra argument.

bjonkman:
@rafaelmudafort is correct. Those are extraneous arguments.

PR #45, opened 2017-09-11
Updated query of git version string, Updated vs-build contents

ghaymanNREL:
This includes the relocation of RANLUX.f90 from the icefloe module to the nwtc_library so that other modules can point to this code.


rafaelmudafort:
- is it manageable to change the executable name from the vs build to `openfast.exe`?
- the visual studio generated build files should be gitignored so that simply building the executable doesn't cause the version to be "dirty"
```
(C:\Users\rmudafor\AppData\Local\Continuum\Anaconda3) C:\Users\rmudafor\Desktop\openfast>git st
HEAD detached at greg/dev
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        modified:   vs-build/MAPlib/MAP_dll.vcxproj
        modified:   vs-build/Registry/FAST_Registry.vcxproj

Untracked files:
  (use "git add <file>..." to include in what will be committed)

        vs-build/FAST/FAST.VC.VC.opendb
        vs-build/FAST/FAST.VC.db
        vs-build/Registry/Debug_Double/

no changes added to commit (use "git add" and/or "git commit -a")
```

PR #46, opened 2017-09-13
ServoDyn and glue code bug fixes

bjonkman:
- fixed problem with constant ElecPwr and GenTrq sent to Bladed DLL controller
- fixed error check that occurred before all information was obtained


michaelasprague:
@rafaelmudafort , please comment on this pull request; what happens with the existing tests?

rafaelmudafort:
@michaelasprague on mac with gfortran, every test case passed the regression test

PR #48, opened 2017-09-14
Add variable in conf.py to switch Doxygen on or off.

jrood-nrel:


michaelasprague:
I asked @jrood-nrel to make this change.  The Doxygen builds have been making RTD builds take too long and often time out.  Developers can enable Doxygen and build locally.

PR #58, opened 2017-10-16
Update the linux-intel baseline solutions and regression test documentation

rafaelmudafort:
The baseline solutions for linux-intel (peregrine) have until now been generated either on a login node or a compute node chosen by the queueing system. Because of hardware inconsistencies, the regression test was sporadically failing on linux.

This pull request ties the linux-intel baseline solutions to a specific hardware, compiler, and math library

- hardware: [Haswell](https://hpc.nrel.gov/users/systems/peregrine)
- compiler: Intel Fortran 17.0.2
- math library: Intel MKL 17.0.2

The regression test documentation is also updated to link to the [r-test repository documentation](https://github.com/openFAST/r-test/tree/pullrequest/rafaelmudafort#baselines) which specifies these details above.

### DIFF
--- a/docs/source/testing/regression_test.rst
+++ b/docs/source/testing/regression_test.rst
@@ -5,13 +5,17 @@ Regression test
 
 The regression test executes a series of test cases which intend to fully describe
 OpenFAST and its module's capabilities. Each locally computed result is compared
-to a static set of baseline results. To account for machine and compiler
+to a static set of baseline results. To account for system, hardware, and compiler
 differences, the regression test attempts to match the current machine and
 compiler type to the appropriate solution set from these combinations
 
 - macOS with GNU compiler (default)
 - Red Hat Enterprise Linux with Intel compiler
 - Windows with Intel compiler
+
+The compiler versions, specific math libraries, and hardware used to generate these baseline
+solutions are documented in the 
+`r-test repository documentation <https://github.com/openFAST/r-test#baselines>`__.
 
 The regression test system can be executed with CMake and CTest or manually with
 an included Python driver. Both systems provide similar functionality with respect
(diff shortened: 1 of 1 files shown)


rafaelmudafort:
@michaelasprague does this configuration for linux-intel hardware, compiler, math library look ok to you?

rafaelmudafort:
Note: the last two commits in this pull request will not be included in the merge. I have corrected this on my fork's dev branch, but the pull request form has not updated

rafaelmudafort:
Update on my previous comment: the pull request has updated on GitHub so the two commits shown here are correct.

jrood-nrel:
I don't understand the first statement here. Don't you only generate the baselines once? Also, even though you shouldn't generate baseline results on a login node, how can you arrive on a login node from the queue? Also, I have never seen the machine have any effect on calculations, but the compiler and blas/math library definitely do, so I don't think you need to be tied to a specific compute node type, but it's a fine thing to do.

rafmudaf:
The baseline solutions are regenerated as required after merging a new pull request or upon a tagged release. That being said, most pull requests will probably not require updated baselines. In any case, I want to nail down and document every parameter which could possibly affect the solutions so that any future questions regarding these baselines is readily available.

In the most recent baseline update prior to this PR, I ran the cases on a login node not through the queueing system (I know, not kosher). In previous updates, I was not specifying the node type when using the queue so it was just choosing the first available.

And its more likely that variations in the math library was causing the differences in solution that I blamed on hardware. For example,
```
[rmudafor@login4 build]$ diff /nopt/nrel/apps/epel/6.6/usr/lib64/libblas.so.3.2.1 /usr/lib64/libblas.so.3.2.1
Binary files /nopt/nrel/apps/epel/6.6/usr/lib64/libblas.so.3.2.1 and /usr/lib64/libblas.so.3.2.1 differ
```
so not being explicit may have been an issue. Nevertheless, machine type is a parameter that can be kept constant.

jrood-nrel:
Ok, I think I get it. As long as you are being consistent with the environment and node type now. I recommend a minimal login environment, `#!/bin/bash -l` and `module purge` in any scripts to enforce consistency. Though the libraries in your example could be at different optimizations, or any number of things causing differing results, they may also merely have different rpaths in them which would still show them as different so I wouldn't expect diff to show two libraries as equivalent very easily.

rafmudaf:
Thanks for the feedback. I'm moving further questions about environment config to [slack](https://a2ehfm-ecp.slack.com/messages/C3PLP0LM6) since it is outside the scope of this pr.

PR #53, opened 2017-10-09
C++ API developer documentation

gantech:
This pull request adds developer documentation through a link to xml output from `doxygen` using `breathe`. I have also reorganized `doxygen` to use different input files when running through cmake and on rtd along the lines of github.com/NaluCFD/NaluWindUtils. You can see a preview of the added documentation here: http://ganesh-openfast.readthedocs.io/en/latest/source/dev/cppapi/index.html

PR #65, opened 2017-11-06
add pdf documentation target and doxygen updates

rafmudaf:
This pull request includes the following documentation updates
- fix broken links in readthedocs to doxygen
- fix broken link to openfast logo references from doxygen
- add a doxygen layout file
- update the default doxygen menu to more appropriate options
- remove graph generation - speeds up the build to avoid timeout on readthedocs
- add instruction for `sphinx-html` and `sphinx-pdf` build targets
- document output locations for local documentation builds

Also included is an added `make` target to build the documentation in pdf format with latex.
**NOTE:** The pdf target builds with errors due to incorrect inline documentation. Fixing that exceeds the scope of this pull request and should be tackled in another pull request.


rafmudaf:
@sayerhs The Breathe functionality seems very useful, but taking on the task of adding the source code documentation into sphinx in the near future is unlikely. Since the doxylink usage is minimal, missing doxylinks in the pdf are acceptable.

However, thanks for pointing out the possibility of Breathe. @gantech is using it in pull request #53 (C++ API documentation), and its clearly worth considering again in the future.

Issue #52, opened 2017-09-30
Setup continuous integration


petebachant:
To automatically check the code (and PRs) will compile and pass tests, these services could be used:

- [ ] Travis CI for Linux
- [ ] AppVeyor for Windows

rafaelmudafort:
@petebachant do you know of any resources for compiling fortran on Travis CI?

petebachant:
If you're running the tests on an Ubuntu machine (default I believe), installing gfortran should be as easy as adding this to `.travis.yml`:

```yaml
before_install:
    - sudo apt-get update -qq
    - sudo apt-get install -qq gfortran
```

rafmudaf:
At the moment, running the regression test suite in a continuous integration environment is impractical due to the time required to run all tests. Some options for reducing the length of the regression test:

- Shorten the simulation times
- Start from a restart file
- Run a subset of the tests

rafmudaf:
TravisCI now supports Windows in beta, so I've set up a preliminary Windows build. There are also hacks for getting an intel compiler on the TravisCI container. There is an existing bug which prevents the Windows container from properly launching when a `secret` is included in the TravisCI configuration and a `secret` is required to pass the Intel Fortran license serial number.

The long regression test issue still persists.

rafmudaf:
Conditions to test
- [ ] Mac - gfortran - single precision
- [ ] Mac - gfortran - double precision
- [ ] Linux - gfortran - single precision
- [ ] Linux - gfortran - double precision
- [ ] Linux - ifort - single precision
- [ ] Linux - ifort - double precision
- [ ] Windows - ifort - single precision
- [ ] Windows - ifort - double precision

Echolon:
Hi @rafmudaf 

I would recommend to put the task element all in the first comments so they get recognized by Github

PR #67, opened 2017-11-09
Add TravisCI configuration file

rafmudaf:
This pull request adds an initial, primitive TravisCI configuration file in response to issue #52 . Currently, this config file simply builds openfast on a macOS instance with the latest gfortran available through homebrew.

This pull request is intended to establish the basis of a continuous integration pipeline which will be expanded in future pull requests. **TravisCI is not to be used for verification, validation, or code quality checks until further testing is integrated.**


rafmudaf:
@petebachant I went with a mac instance for this basic setup. Is there anything you would add to an initial configuration? Of course, actually running the regression tests should be integrated soon, but a couple of them currently take about an hour so TravisCI will time out.

petebachant:
Looks like a good start just to make sure any pull requests compile. Are users more likely to use Mac OS versus Linux?

rafmudaf:
@petebachant Most users (not developers) of OpenFAST are on Windows or Linux and the developers at NREL use a mix of mac and Windows. However, we are currently supporting all three systems in the regression test, so ideally all three will be included in continuous integration.

PR #74, opened 2017-11-27
Respect the option for the user to turn off Tower Aerodynamics in Aerodyn15 input file

gantech:


gantech:
Ok.. I have verified that this option to turn off tower aerodynamics works through simulations of the elliptic wing case. Could we please merge this.

Issue #73, opened 2017-11-22
Warning in header file


jrood-nrel:
Is it possible to fix this warning directly in OpenFAST?

```
include/FAST_Library.h:37:12: warning: ‘AbortErrLev’ defined but not used [-Wunused-variable]
```

rafmudaf:
hmm ... but it is used:
https://github.com/OpenFAST/openfast/blob/master/modules-local/fast-library/src/FAST_Library.f90#L411
@gantech ?

jrood-nrel:
Then it should probably just be declared in fortran and not the header?

gantech:
Ah.. yes.. I have to fix this issue. The warning keeps showing up on every compile. The problem is I do use it in C++ as well here https://github.com/OpenFAST/openfast/blob/master/glue-codes/fast-cpp/src/OpenFAST.cpp#L322. So I need to remove including this file from OpenFAST.H and just include it directly in OpenFAST.cpp.

gantech:
Fixed through https://github.com/OpenFAST/openfast/pull/76

Issue #59, opened 2017-10-18
Need a Way to Compile the FAST Dynamic Library for Calling OpenFAST within Simulink


jjonkman:
The Simulink interface to FAST consists of Simulink calling the FAST S-Function, which calls the FAST dynamic library.  While the OpenFAST repository includes the glue code and MATLAB script for compiling the FAST S-Function, the OpenFAST repository doesn't include a CMake script or VS-Build solution for compiling the FAST dynamic library.

Related to this missing build, is a misplacement of the FAST_Library.f90 source file.  FAST_Library.f90 is to the FAST dynamic library for the Simulink interface as FAST_Prog.f90 is to compiling the OpenFAST executable.  FAST_Library.f90 is currently included in the modules-local/fast-library/src directory; however, it seems like FAST_Library.f90 should be moved to the glue-codes/Simulink directory.

sayerhs:
Wouldn't turning on shared library `ON` in CMake allow using `openfastlib.so` with Simulink?

ghaymanNREL:
@sayerhs, yes you are correct (I didn't know this until reading your comment), for cmake builds, but there is currently no equivalent build for visual studio.  (The vs-build of the FASTlib, does not include the FAST_Library.f90 file, and the vs-build output is only a static library)

bjonkman:
A couple of differences with the `openfastlib.so` as it seems to be implemented in CMake:
- It should use the Matlab libmex library so that all messages that normally get written to the screen appear in the Matlab command window. On Windows with Intel Fortran, we replace SysIVF.f90 with SysMatlab.f90 (which calls Matlab routines for printing) and link with libmex.lib from the Matlab installation. As far as I know, no one has modified the other Sys*.f90 files for other platforms/compilers.
- It should use the `COMPILE_SIMULINK` preprocessor directive.

Instructions for how this is supposed to work are on pages 54-55 of the [FAST8 ReadMe](https://wind.nrel.gov/nwtc/docs/README_FAST8.pdf)

bjonkman:
One other thing to note here is that FAST_Library.h is used in FAST_SFunc.c (which gets compiled in MATLAB as a mex file). It looks like the interfaces defined in the header file have changed, but the code that calls those routines in FAST_SFunc.c has not.

gantech:
Thanks @bjonkman. 

[...]

PR #76, opened 2017-11-27
Move initialization of static variable from header file to a .cpp file

gantech:
@jrood-nrel  Fixes issue https://github.com/OpenFAST/openfast/issues/73

@jjonkman @ghaymanNREL : Do we know of any one who uses FAST_Library.h? I moved the definition of "AbortErrLev" into the OpenFAST class and initialized in the C++ file. 

Other minor edits to move some header includes into the .cpp file.

### DIFF
--- a/glue-codes/fast-cpp/src/OpenFAST.H
+++ b/glue-codes/fast-cpp/src/OpenFAST.H
@@ -2,9 +2,6 @@
 #define OpenFAST_h
 #include "FAST_Library.h"
 #include "sys/stat.h"
-#include "math.h"
-#include <iostream>
-#include <fstream>
 #include <string>
 #include <cstring>
 #include <stdexcept>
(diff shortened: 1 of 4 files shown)


jjonkman:
@gantech -- I'm not familiar with the FAST_Library.h file.  However, the the FAST_Library.f90 file is used by the Simulink-FAST interface, but this interface has not yet been tested in OpenFAST (see issue #59).

gantech:
@jjonkman : Thanks. This commit shouldn't affect any simulink/fortran users at all. I'm concerned if this commit will affect any one who depends on the the variable "AbortErrLev" being defined in the FAST_Library.h file. I don't see a reason for any one to use the FAST_Library.h file at all. My understanding was that this file was created primarily for interfacing to C driver programs. However, the new interface supports only C++ and not C drivers. Not sure if this is a concern.


If any one has any other suggestions to get around issue https://github.com/OpenFAST/openfast/issues/73 without modifying FAST_Library.h, please let me know.

Issue #83, opened 2018-01-18
regression test always requires matplotlib


rafmudaf:
The regression test framework always checks for matplotlib, but should only do so when the cmake flag `CTEST_PLOT_ERRORS` is `ON`.

rafmudaf:
fixed in pull request #91

PR #91, opened 2018-02-20
Move matplotlib import to the public function that uses it

rafmudaf:
Importing matplotlib for the entire module causes an error if matplotlib is not available. The plotting functionality is not currently supported by NREL so it should remain optional. Moving the import statement to a function moves the failure to that function call rather than the module import.

With this pull request, if matplotlib is not available and plotting is off, the result summaries in html format are still exported correctly. If matplotlib is not available and plotting is on, the html files are written but the plotting routine fails.

When running the regression test through CTest, this python `ModuleNotFoundError` fails silently. Users can see the error with the verbose flag in the CTest call: `ctest -V`

This fixes issue #83 .

### DIFF
--- a/reg_tests/executeOpenfastRegressionCase.py
+++ b/reg_tests/executeOpenfastRegressionCase.py
@@ -175,7 +175,11 @@ if not pass_fail.passRegressionTest(normalizedNorm, tolerance):
         failMaxNorm = [maxNorm[i] for i,channel in enumerate(testInfo["attribute_names"]) if normalizedNorm[i] > tolerance]
         initializePlotDirectory(localOutFile, failChannels, failRelNorm, failMaxNorm)
         for channel in failChannels:
-            plotOpenfastError(localOutFile, baselineOutFile, channel)
+            try:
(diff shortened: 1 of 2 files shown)


sayerhs:
@rafmudaf Why not use check and gracefully disable plotting? 

```
try:
    import matplotlib
    matplotlib.use("Agg")
    import matplotlib.pyplot as plt
    from matplotlib.ticker import FormatStrFormatter
    have_plt = True
except ImportError:
   have_plt = False

# and in function 

if not have_plt:
    return
````

**Comment by rafmudaf** [comment]
@sayerhs If a user doesn't realize matplotlib is not installed and attempts to plot, this gives no indication of the failure. My update prevents a crash when plotting is off and matplotlib is not available, but maintains the Python exception reporting.

**Comment by sayerhs** [comment]
Well, just print a message instead of a return 

```
if not have_plt:
    print("Matplotlib is not installed on this system; plotting functions will be disabled")
```

rafmudaf:
Is this functionally different than letting Python handle the exception?

Issue #1, opened 2017-01-06
AeroDyn14 Driver


sayerhs:
AeroDyn14 driver contains `USE AeroDyn_Types` instead of `USE AeroDyn14_Types`. Looks like the code needs to be updated to reflect the changes since introduction of AeroDyn v15.

rafmudaf:
For future reference, I've updated cmake to include aerodyn14_driver in the build and I've updated AD_Driver to use the aerodyn14 types and function calls. However, it looks like updates need to be made to the registry
```
Scanning dependencies of target aerodyn14_driver
[ 95%] Building Fortran object modules-local/aerodyn14/CMakeFiles/aerodyn14_driver.dir/src/AD_Driver.f90.o
/Users/rmudafor/Development/openfast/modules-local/aerodyn14/src/AD_Driver.f90:68:23:

    InitInData%ADOptions%ADFile = 'MyInputFileName.inp'
                       1
Error: 'adoptions' at (1) is not a member of the 'ad14_initinputtype' structure
/Users/rmudafor/Development/openfast/modules-local/aerodyn14/src/AD_Driver.f90:77:67:

    CALL AD14_Init( InitInData, u(1), p,  x, xd, z, OtherState, y, m, TimeInterval, InitOutData, ErrStat, ErrMsg )
                                                                   1
Error: Symbol 'm' at (1) has no IMPLICIT type
/Users/rmudafor/Development/openfast/modules-local/aerodyn14/src/AD_Driver.f90:106:85:

       CALL AD14_CalcOutput( Time, u(1), p, x, xd, z, OtherState, y, ErrStat, ErrMsg )
                                                                                     1
Error: Type mismatch in argument 'm' at (1); passed INTEGER(4) to TYPE(ad14_miscvartype)
/Users/rmudafor/Development/openfast/modules-local/aerodyn14/src/AD_Driver.f90:115:84:

       call AD14_UpdateStates( Time, u(1), p, x, xd, z, OtherState, ErrStat, ErrMsg )
                                                                                    1
Error: Type mismatch in argument 'n' at (1); passed TYPE(ad14_inputtype) to INTEGER(4)
/Users/rmudafor/Development/openfast/modules-local/aerodyn14/src/AD_Driver.f90:127:69:

    CALL AD14_End( u(1), p, x, xd, z, OtherState, y, ErrStat, ErrMsg )
                                                                     1
[...]

PR #92, opened 2018-02-20
Remove legacy AeroDyn14 driver

rafmudaf:
This pull request removes the AeroDyn14 driver code as it is not fully implemented and references an outdated version of AeroDyn 14.

This pull request fixes issue #1 .


jjonkman:
I agree; delete it.

Issue #88, opened 2018-02-06
Add minimum compiler versions in cmake config


rafmudaf:
There are some known compiling errors that were documented in the Compiling Instructions for FAST8:

1. The version of gfortran must be at least 4.6.0. NWTC Library uses some quad-precision variables, which are not supported in earlier versions of gfortran.
Possible workarounds: You might be able to set QuKi to some other supported real kind that is not 4 or 8 (not sure REAL(2) is supported, but you could try); otherwise, you'd have to set QuKi to 4 or 8 and remove the QuKi MODULE PROCEDURE lines from all the INTERFACE blocks at the top of the NWTC Library source files.
2. Intel Fortran must be version 11 or later. NWTC Library uses the gamma() function, which is not provided in earlier versions.
Possible workarounds: You can write your own gamma() function in the SysI*.f90 files, or remove the calls to gamma() (e.g., set nwtc_gamma = 1) and avoid using IceDyn and the wave spreading features of HydroDyn.

It is recommended to put a minimum version requirement for both gfortran and ifort compilers in cmake.

@bjonkman Thanks for pointing this out.

rafmudaf:
Fixed in pull request #96

Issue #87, opened 2018-01-30
OpenFast Compilation Error


Jochen-IST:
Dear developers, 

I am new with the OpenFast project and tried to compile OpenFast on a 64-bit LINUX machine (CentOS 6.9) and a GNU 4.4.7 C and Fortran compiler.
I could create the Makefiles (cmake) for OpenFAST. Here are the messages of "cmake": 

-- The CXX compiler identification is GNU 4.4.7
-- The C compiler identification is GNU 4.4.7
-- The Fortran compiler identification is GNU 4.4.7
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working Fortran compiler: /usr/bin/gfortran
-- Check for working Fortran compiler: /usr/bin/gfortran  -- works
-- Detecting Fortran compiler ABI info
-- Detecting Fortran compiler ABI info - done
-- Checking whether /usr/bin/gfortran supports Fortran 90
-- Checking whether /usr/bin/gfortran supports Fortran 90 -- yes
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Looking for pthread_create
-- Looking for pthread_create - not found
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE
-- Looking for Fortran sgemm
-- Looking for Fortran sgemm - found
-- A library with BLAS API found.
-- A library with BLAS API found.
-- Looking for Fortran cheev
-- Looking for Fortran cheev - found
-- A library with LAPACK API found.
-- Setting system file as: src/SysGnuLinux.f90
-- Configuring done
-- Generating done
-- Build files have been written to: /fs/jochen/OpenFAST/build

But when running "make", I get the following error messages for the compilation of SysGnuLinux.f90:

build-g44:/fs/jochen/OpenFAST/build>make
[  4%] Built target fast_registry
[  5%] Building Fortran object
[...]

PR #96, opened 2018-02-27
Add minimum compiler versions to the cmake configuration

rafmudaf:
This pull request adds minimum version requirements for the GNU gfortran and Intel ifort compilers. 

As @bjonkman points out [here](https://github.com/OpenFAST/openfast/issues/87#issuecomment-363187357):
```
There are some known compiling errors that were documented in the Compiling Instructions for FAST8:

1. The version of gfortran must be at least 4.6.0. NWTC Library uses some quad-precision 
variables, which are not supported in earlier versions of gfortran.

Possible workarounds: You might be able to set QuKi to some other supported real kind that is not 
4 or 8 (not sure REAL(2) is supported, but you could try); otherwise, you'd have to set QuKi to 4 or 
8 and remove the QuKi MODULE PROCEDURE lines from all the INTERFACE blocks at the top of the 
NWTC Library source files.

2. Intel Fortran must be version 11 or later. NWTC Library uses the gamma() function, which is not 
provided in earlier versions.

Possible workarounds: You can write your own gamma() function in the SysI*.f90 files, or remove 
the calls to gamma() (e.g., set nwtc_gamma = 1) and avoid using IceDyn and the wave spreading 
features of HydroDyn.

It is recommended to put a minimum version requirement for both gfortran and ifort compilers in 
cmake.
```

This pull request is a fix for issue #88 and regression tests are not required as this does not affect the calculations.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,6 +41,13 @@ option(FPE_TRAP_ENABLED "Enable FPE trap in compiler options" off)
 option(ORCA_DLL_LOAD "Enable OrcaFlex Library Load" off)
 option(BUILD_FAST_CPP_API "Enable building FAST - C++ API" off)
 
+# Verify proper compiler versions are available
+if(CMAKE_Fortran_COMPILER_ID STREQUAL "GNU")
+  if(CMAKE_Fortran_COMPILER_VERSION VERSION_LESS "4.6.0")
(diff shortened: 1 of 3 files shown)


rafmudaf:
@sayerhs thanks for pointing that out, done.

PR #98, opened 2018-03-12
TurbSim bug fixes

bjonkman:
This fixes two minor problems in TurbSim
- The default power law exponent (PLExp) was set to 0 if you used any of the EWM models.
- If ScaleIEC=2 was used with the IEC spectra and the coherence for v- or w- components was *not* the default model (i.e., not "NONE"), you wouldn't get the same TI at all points in the v- and w- components. It would effectively use ScaleIEC=1 instead.


rafmudaf:
@ghaymanNREL could you take a second look at this pull request when time allows?

michaelasprague:
@jjonkman -- I think @rafmudaf is waiting for your approval to merge.

PR #102, opened 2018-03-15
Output format for CR

NobuhiroKUSUNO:
Just tiny fix of write format in WrOver.
CR (ACHAR(13)) can work fine.
This fix may be applicable for the other Sys*.f90.
But I hove no such work environments.

### DIFF
--- a/modules-local/nwtc-library/src/SysGnuLinux.f90
+++ b/modules-local/nwtc-library/src/SysGnuLinux.f90
@@ -468,7 +468,7 @@ CONTAINS
 
    ELSE
       ! bjj: note that this will almost certainly write more than MaxWrScrLen characters on a line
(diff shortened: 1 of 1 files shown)


rafmudaf:
@ghaymanNREL @NobuhiroKUSUNO is the carriage return actually 2 characters long?
```
mbp@~/Desktop: more test.f90 
program test_achar
  character c
  c = achar(13)
  print *, len(c)
  print *, c
end program test_achar
mbp@~/Desktop: gfortran test.f90 ; ./a.out
           1
 

```

rafmudaf:
That being said, I'm not sure how much this padding really matters. If there's not downside, lets merge this.

bjonkman:
I think the issue here is that it's printing two strings, `CR` and `Str`, which needs format (2A).  One string 2 characters long would be `A2`. I think this change looks correct.

Issue #105, opened 2018-03-27
Incorrect help prompt


petebachant:
Build OpenFAST on Ubuntu, and tried running

    openfast -h

The help flag doesn't seem to work, and the executable is still looking for an input file. Also, it seems the executable names are `FAST.exe` and `FAST_x64.exe`, not  `openfast`.

Full output:

```
pete@pete-VirtualBox:~/projects/openfast/build$ openfast -h

  Syntax is:

     FAST [-h] <InputFile>

  where:

     -h generates this help message.
     <InputFile> is the name of the required primary input file.

  Note: values enclosed in square brackets [] are optional. Do not enter the brackets.


 **************************************************************************************************
 OpenFAST

 Copyright (C)  National Renewable Energy Laboratory

 This program is licensed under Apache License Version 2.0 and comes with ABSOLUTELY NO WARRANTY.
 See the "LICENSE" file distributed with this software for details.
 **************************************************************************************************

 OpenFAST-v0.1.0-948-gfcb6fab7-dirty
 Compile Info:
  - Architecture: 64 bit
  - Precision: double
 Execution Info:
  - Date: 03/27/2018
  - Time: 13:50:04-0400


  Syntax is:

     OpenFAST [-h] <InputFile>

  where:

     -h generates this help message.
     <InputFile> is the name of the required primary input file.

  Note: values enclosed in square brackets [] are optional. Do not enter the brackets.


  Syntax is:

     FAST_x64.exe [-h] <InputFile>

  where:

     -h generates this help message.
     <InputFile> is the name of the required primary input file.

  Note: values enclosed in square brackets [] are optional. Do not enter the brackets.


 FAST_InitializeAll:The required input file was not specified on the command line.

 FAST encountered an error during module initialization.
  Simulation error level: FATAL ERROR

  Aborting OpenFAST.
```

rafmudaf:
Pull request #116 updates the executable names and fixes the bug where the help prompt is shown twice with the `-h` flag.

rafmudaf:
fixed by #421

PR #107, opened 2018-03-27
Feature/help

NobuhiroKUSUNO:
#105 @petebachant
Add message for restart option and modify the program name.


NobuhiroKUSUNO:
Thanks for comments, Checked #116 , @rafmudaf

PR #111, opened 2018-05-01
Update to latest HDF5

tonyinme:
The wget command does not work for version 1.8.19.
The version on the web is 1.8.20.
The script has been updated to do this.

### DIFF
--- a/share/fast-install.sh
+++ b/share/fast-install.sh
@@ -69,11 +69,11 @@ compileHDF5() {
     echo -n "   Getting source"
     [ -d modules-ext/hdf5 ] || mkdir modules-ext/hdf5
     cd modules-ext/hdf5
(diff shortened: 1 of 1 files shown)


tonyinme:
Hey Ganesh,

Yes, I agree.
I have always had the same issue with hdf5, there is no clean way of
getting the source code (like git clone).
If that link you are suggesting works, then go ahead and change it to that.
The scripts in share are very good, and they should be the go-to for
compilation.

sayerhs:
@gantech It would be good if you set the same version for OpenFAST as Trilinos/Nalu so that there are no conflicts when we attempt to link to Nalu.

rafmudaf:
@gantech is this pull request good to go or do you need to address @sayerhs suggestion above?

gantech:
As soon as @tonyinme accepts this pull request https://github.com/tonyinme/openfast/pull/1, this will be ready to go.

tonyinme:
Looks good!

PR #112, opened 2018-05-04
Add CMake rules for detecting Intel MKL

sayerhs:
Add `FindMKL.cmake` to detect and configure Intel MKL settings for BLAS/LAPACK libraries. This eliminates the need for specifying BLAS_LIBRARIES manually when compiling with Intel compilers + Intel MKL libraries.


sayerhs:
@rafmudaf and @gantech Would be great if you can test it.

PR #113, opened 2018-05-06
CMake: Change logic for default install location

sayerhs:
Change default installation logic in top-level CMakeLists.txt so that the installation directory not dependent on the value of `CMAKE_Fortran_MODULE_DIRECTORY`. The default location will be configured correctly even if the user does not name the build directory `build`, but
instead chooses some other name for the build directory.


sayerhs:
@rafmudaf You already did have the `CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT`, I just changed the hard-coded `build/ftnmods` logic.

Issue #80, opened 2018-01-04
Change BeamDyn to Define the Motion Output Mesh Based on Quadrature


jjonkman:
Currently, the motion output mesh from BeamDyn is based on the finite-element nodes.  However, with low-order elements commonly used (Order_Elem <= 7), this results in only a handful of nodes where motions are passed from BeamDyn to the aerodynamic solver (e.g. AeroDyn or CFD) resulting in a coarse distribution of motions used in the aerodynamics calculation.  It would be better to define the motion output mesh from BeamDyn based on the spatial-integration method, so that if trapezoidal quadrature is used (Quadrature = 2), the motion distribution passed to the aerodynamics solver is better resolved.

PR #120, opened 2018-05-21
Changing BeamDyn blade motion mesh output to quadrature points

ashesh2512:
Changed the output of blade motion mesh from BeamDyn to be defined on quadrature points - 

-- This works for both Gauss and Trapezoidal quadrature. 
-- The Calc_WriteOutput routine has been modified such that it still outputs the data to log files based on FE nodes.
-- Only the Fast with BeamDyn regression test (5MW_Land_BD_DLL_WTurb) fails with negligible differences. Solutions will be updated. 

This pull request fixes issue #80 

**NOTE: This pull request is not ready for merge**

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,11 +34,6 @@ if (NOT CMAKE_BUILD_TYPE)
     "Choose the build type: Debug Release" FORCE)
 endif (NOT CMAKE_BUILD_TYPE)
 
-# debug flags
-if(CMAKE_BUILD_TYPE MATCHES Debug)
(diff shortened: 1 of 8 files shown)


rafmudaf:
This pull request will cause merge conflicts with concurrent work at Envision. @ghaymanNREL is the process in place for syncing the `openfast/openfast` repository with Envision's fork? If so, this is a great opportunity to test it and document the process for future reference.

ashesh2512:
For the time being it was determined that the current pull request will create significant code divergence with Envision's upcoming pull request. Hence this request is being closed.

PR #121, opened 2018-05-21
Improve BD_InitShpDerJaco unit test

rafmudaf:
- Use a 3 node element rather than 2 node
- Deallocate variables after use

Currently, this unit test has memory management issues which cause test failures intermittently. This pull request fixes the memory problem and, as of openfast/dev, executes successfully every time.


ashesh2512:
The unit test has also been modified to test functionalities added to the function after commit #114

Issue #16, opened 2017-04-20
Issue with VTK output file naming convention


mchurchf:
When you turn on the VTK output, the file naming convention is something like:
Model.AD_Blade1.t0.vtp
Model.AD_Blade1.t1.vtp
...
Model.AD_Blade1.tN.vtp

In order for Paraview to read in and animate the entire time series, the "t" should not be there.  Instead, the filename should look like:
Model.AD_Blade1.0.vtp
Model.AD_Blade1.1.vtp
...
Model.AD_Blade1.N.vtp

sayerhs:
@mchurchf Please see if #17 does what you are looking for. If you and @michaelasprague agree, I'll merge and update `master` and `dev`

jjonkman:
I'm not an expert in ParaView, but I've not seen any problems having the "t" when trying to animate as the *.t#.vtk files get sorted properly.  If this really is a problem, we'll not only have to change the visualization output from FAST, but we've used the same numbering convention for the ambient wind input to FAST.Farm (from ABLSolver) and the disturbed wind visualization output from FAST.Farm, which should be changed as well.

gruchalla:
There are many ways to use ParaView. I think both xxx.t0.vtp and xxx.0.vtp will work in many instances. Although, xxx.0.vtp is a bit more standard and likely to be recognized by a wider range of use cases. 

Also, it is generally a good idea to pad your time steps with 0 to facilitate sorting, e.g., xxx.0000.vtp ... xxx.9999.vtp  vs  xxx.0.vtp ... xxx.9999.vtp

ghaymanNREL:
I'm inclined to only make a change if someone can present a concrete use-case where the current naming convention is causing an issue.

mchurchf:
Yes, my Paraview 5.3.0-RC2 on my Mac won’t open it with the t in the name.

mchurchf:
Yes, I'm running Paraview 5.3.0-RC2 on the Mac, and it won't open the files with the "t" in the name.

michaelasprague:
@ghaymanNREL -- can we make this change along with @gruchalla 's suggestion of padding with zeros, "e.g., xxx.0000.vtp ... xxx.9999.vtp vs xxx.0.vtp ... xxx.9999.vtp"

rafmudaf:
@mschmidt271 and @rafmudaf are currently working on this.

rafmudaf:
Fixed by pull request #126

PR #125, opened 2018-05-24
RE: Issue with VTK output file naming convention #16

mschmidt271:
This update addresses Issue #16: Issue with VTK output file naming convention.

The "t" prefix to the timestamps has been eliminated in MeshWrVTK(), MeshWrVTKfields(), and MeshWrVTK_Ln2Surface(), all of which are in ModMesh.f90.

Additionally, the subroutines in FAST_Subs.f90 that call the above subroutines [WrVTK_AllMeshes(), WrVTK_BasicMeshes(), and WrVTK_Surfaces()] pass the maximum number of VTK write-out steps to the above subroutines, so that the timestamps may be padded with the appropriate number of zeros, as discussed in the Issue thread. This approach certainly means that the "width" of the number is repeatedly calculated (at every write-out step), but it seems relatively cheap in the grand scheme of things, and avoids adding it onto one of the persistent derived types, like the parameters. However, this alternate approach may be preferable.

Also, as a side note, my editor stripped a lot of whitespace out of those files. So, I just discovered that adding "?w=1" to the URL when you are looking at the diff on GitHub lets you ignore whitespace and makes the diff much more readable.


bjonkman:
I have been working on a significant amount of code for another pull request--including some changes in these files--and would really appreciate it if you could revert the whitespace changes on lines that have otherwise not been changed. It makes unnecessary merge conflicts.

PR #143, opened 2018-06-20
Build up BeamDyn Unit Testing and modify Subroutine Arguments

mschmidt271:
This pull request updates and adds unit tests for subroutines in BeamDyn and DeamDyn_Subs.

As well, the arguments to some subroutines have been altered to avoid passing in full derived types (e.g., parameters, miscvar) in favor of only the relevant variables within them.

This pull request is not completely finished, but just wanted to submit it at this point to get some eyes on it and make sure this is moving in the right direction before going much further.


andrew-platt:
Just thought I'd throw my 2 cents in here.  Removing types and passing variables makes the subroutines a little less readable in my opinion.  For example, removing the paramters type from the `BD_DisplacementQP` routine seems a little extreme given there are now 6 additional variables passed in.

When I started the rewrite of BeamDyn 2+ years ago at Envision Energy, there were very few derived types in BeamDyn.  That made it nearly impossible to read, particularly a couple of subroutine layers in. So, I strongly caution against removing the passing of the derived types.  I know it complicates the testing routines a little, but it makes it much more readable in my opinion.

mschmidt271:
@andrew-platt : Thank you for the input. Readability (in conjunction with testability) is definitely something that @michaelasprague, @rafmudaf, and I have discussed in regards to these subroutines. While breaking out the variables from the derived types may make the subroutine signatures longer in some cases, we feel that there is value in being able to see exactly what comes into and out of these subroutines, rather than opaquely passing in the derived types when only small portions are being used. This pattern adopts best practices for testable code by establishing high cohesion, low coupling, and separation of concerns. The design intent generally follows this hierarchy:

* Low level subroutines do not read from or write to derived types and have associated unit tests
* Mid level subroutines have limited access to derived types and have associated integration tests (the definition of "limited" is left to the developer)
* High level subroutines have full access to derived types and are tested in the regression tests or with associated integration tests

To be sure, there is a balance to be struck, as we don't want to create monstrous subroutine signatures. As well, we only plan to conduct these changes on the low-level subroutines that qualify for unit testing, and do not plan on making these same changes to subroutines that will be passing these derived types down through multiple layers.

rafmudaf:
I'm closing this pull request for now as work has stalled. This work is saved at https://github.com/rafmudaf/openfast/tree/feature/beamdyn_api_improvement in addition to the original branch at https://github.com/mschmidt271/openfast/tree/f/BD_mod.

PR #145, opened 2018-06-25
Strong coupling feature along with major C++ API upgrades

gantech:
This pull request implements a strong coupling feature for OpenFAST that can be used with external driver programs through the C++ API. The C++ API now also allows for use of BeamDyn for actuator line simulations. The C++ API has also been upgraded to make the code cleaner along with error checking for typical user errors. The C++ API gets a significant documentation upgrade that also addresses several comments from https://github.com/OpenFAST/openfast/pull/53. The updated documentation can be viewed live here http://ganesh-openfast.readthedocs.io/en/latest/source/dev/cppapi/index.html.


docs/conf.py
```diff
@@ -27,12 +27,16 @@
 sourcedir = sys.argv[-2]
 builddir = sys.argv[-1]
 
+# Only build API docs if the user specifically requests it. On RTD we build it
+# all the time
+use_breathe = tags.has("use_breathe") or readTheDocs
+
 # Use this to turn Doxygen on or off
-useDoxygen=False
+useDoxygen=True
```
sayerhs: This is set to `False` in the `dev` branch, have you confirmed that this doesn't lead to timeouts/errors with the Fortran sources?

glue-codes/openfast-cpp/src/FAST_Prog.cpp
```diff
@@ -11,23 +11,61 @@ inline bool checkFileExists(const std::string& name) {
 void readTurbineData(int iTurb, fast::fastInputs & fi, YAML::Node turbNode) {
 
   //Read turbine data for a given turbine using the YAML node
-  fi.globTurbineData[iTurb].TurbID = turbNode["turb_id"].as<int>();
-  fi.globTurbineData[iTurb].FASTInputFileName = turbNode["FAST_input_filename"].as<std::string>() ;
-  fi.globTurbineData[iTurb].FASTRestartFileName = turbNode["restart_filename"].as<std::string>() ;
+  if (turbNode["turb_id"]) {
```
sayerhs: Recommend building some higher level YAML utilities and using it here.

glue-codes/openfast-cpp/src/FAST_Prog.cpp
```diff
@@ -11,23 +11,61 @@ inline bool checkFileExists(const std::string& name) {
...
+  } else {
+      fi.globTurbineData[iTurb].TurbineBasePos = std::vector<float>(3,0.0);
   }
   if (turbNode["turbine_hub_pos"].IsSequence() ) {
-    fi.globTurbineData[iTurb].TurbineHubPos = turbNode["turbine_hub_pos"].as<std::vector<double> >() ;
+      fi.globTurbineData[iTurb].TurbineHubPos = turbNode["turbine_hub_pos"].as<std::vector<double> >() ;
+  } else {
+      fi.globTurbineData[iTurb].TurbineHubPos =  std::vector<double>(3,0.0);
+  }
+  if (turbNode["num_force_pts_blade"]) {
```
sayerhs: It might be better to throw error on this (i.e., make these options mandatory) instead of defaulting to values that are not correct.

glue-codes/openfast-cpp/src/FAST_Prog.cpp
```diff
@@ -103,43 +184,70 @@ int main() {
...
+  bool setExpLawWind; // Set wind speed at Aerodyn nodes based on an exponential profile. Useful for testing the C++ API before running actuator line simulations.
 
   std::string cDriverInputFile="cDriver.i";
   fast::OpenFAST FAST;
   fast::fastInputs fi ;
   try {
-    readInputFile(fi, cDriverInputFile, &tEnd);
+      readInputFile(fi, cDriverInputFile, &tEnd, &couplingMode, &setExpLawWind);
   }
   catch( const std::runtime_error & ex) {
```
sayerhs: Why a catch here? You are printing the error, but removing the output of stacktrace that would have otherwise been printed by letting the default handler handle this exception.

glue-codes/openfast-cpp/src/OpenFAST.H
```diff
@@ -22,53 +22,138 @@
...
   TOWER = 2,
   ActuatorNodeType_END
 };
 
+/** An id to indicate the start type of a simulation.
+ * init - Start the simulation from scratch
+ * trueRestart - Restart from a checkpoint file. Code expects checkpoint files for all parts of the simulation including the controller.
+ * restartDriverInitFAST - Start all turbines from scratch and use the velocity data in 'velData.h5' file to run upto desired restart time, then continue the simulation like ''trueRestart'.
+ */
 enum simStartType {
```
sayerhs: Might want to follow the convention of all caps for enum that you used for `ActuatorNodeType`

glue-codes/openfast-cpp/src/OpenFAST.H
```diff
@@ -22,53 +22,138 @@
...
+    double vel_vel_resid;
+    double x_force_resid;
+    double xdot_force_resid;
+    double orient_force_resid;
+    double vel_force_resid;
+    double force_resid;
+};
+
+//! An enum to keep track of information stored at different time steps
+enum timeStep {
```
sayerhs: Descriptive names? `STATE_NM1`, `STATE_NP1` and so on?

glue-codes/openfast-cpp/src/OpenFAST.H
```diff
@@ -182,75 +329,134 @@ class OpenFAST {
...
-  int get_procNo(int iTurbGlob) { return turbineMapGlobToProc[iTurbGlob] ; } // Get processor number of a turbine with global id 'iTurbGlob'
+  //! Get the global MPI rank of the processor containing turbine number 'iTurbGlob'
+  int get_procNo(int iTurbGlob) { return turbineMapGlobToProc[iTurbGlob] ; }
+  //! Get the local turbine number of the turbine number 'iTurbGlob'
   int get_localTurbNo(int iTurbGlob) { return reverseTurbineMapProcToGlob[iTurbGlob]; }
+  //! Get the total number of turbines in the simulation
   int get_nTurbinesGlob() { return nTurbinesGlob; } 
 
+  //! Get the nacelle drag coefficient of turbine number 'iTurbGlob'
   float get_nacelleArea(int iTurbGlob) { return get_nacelleAreaLoc(get_localTurbNo(iTurbGlob)); }
```
sayerhs: Is this safe to call on every MPI rank?

glue-codes/openfast-cpp/src/OpenFAST.cpp
```diff
@@ -7,304 +8,722 @@ int fast::OpenFAST::AbortErrLev = ErrID_Fatal; // abort error level; compare wit
...
+                velForceNodeData[iTurb][fast::nm1].xdot_vel[i*3+j] = velForceNodeData[iTurb][fast::np1].xdot_vel[i*3+j];
+                velForceNodeData[iTurb][fast::nm2].xdot_vel[i*3+j] = velForceNodeData[iTurb][fast::np1].xdot_vel[i*3+j];
+                velForceNodeData[iTurb][fast::n].vel_vel[i*3+j] = velForceNodeData[iTurb][fast::np1].vel_vel[i*3+j];
+                velForceNodeData[iTurb][fast::nm1].vel_vel[i*3+j] = velForceNodeData[iTurb][fast::np1].vel_vel[i*3+j];
+                velForceNodeData[iTurb][fast::nm2].vel_vel[i*3+j] = velForceNodeData[iTurb][fast::np1].vel_vel[i*3+j];
+            }
+        }
+        for (int i=0; i<nfpts; i++) {
+            for (int j=0 ; j < 3; j++) {
+                velForceNodeData[iTurb][fast::n].x_force[i*3+j] = velForceNodeData[iTurb][fast::np1].x_force[i*3+j];
```
sayerhs: Recommend creating some `set_state`, `swap_states` functions.

glue-codes/openfast-cpp/src/OpenFAST.cpp
```diff
@@ -338,584 +757,622 @@ void fast::OpenFAST::calc_nacelle_force(
...
+    // Allocate memory for ExtInfw Input types in FAST
+    i_f_FAST.resize(nTurbinesProc) ;
+    o_t_FAST.resize(nTurbinesProc) ;
+    
+    sc_i_f_FAST.resize(nTurbinesProc) ;
+    sc_o_t_FAST.resize(nTurbinesProc) ;
+    
+}
+
+void fast::OpenFAST::allocateMemory2(int iTurbLoc) {
```
sayerhs: What's the difference between `allocateMemory` and `allocateMemory2`, give it a descriptive name may be?

docs/conf.py
gantech: Yes. I was able to generate the documentation here http://ganesh-openfast.readthedocs.io/en/latest/source/dev/cppapi/index.html. The Doxyfile.rtd that is run on readthedocs only runs Doxygen on C++ files for now. My thinking was that when breathe works for Fortran and/or a different documentation strategy is adopted for generating automatic documentation for Fortran sources, the Doxyfile.rtd can be modified appropriately.

glue-codes/openfast-cpp/src/OpenFAST.cpp
gantech: Thanks. This comment was super useful. Reduced the code quite a bit! Implemented in 9f34f9eb119d38dc0c78bfe58fdd23d084d02430.

glue-codes/openfast-cpp/src/OpenFAST.cpp
gantech: Thanks. Changed names to `allocateMemory_preInit()` and `alllocateMemory_postInit()`. Implemented in 9f34f9eb119d38dc0c78bfe58fdd23d084d02430.

glue-codes/openfast-cpp/src/OpenFAST.H
gantech: Ah... no. I realize this is true of many functions. I've added comments in the documentation above each function to indicate whether a particular function MUST be called from every MPI rank, is safe to be called from every MPI rank, or should be called only from the processor containing the turbine. Implemented in commit 9f34f9eb119d38dc0c78bfe58fdd23d084d02430.

glue-codes/openfast-cpp/src/OpenFAST.H
gantech: Agreed. Implemented in 9f34f9eb119d38dc0c78bfe58fdd23d084d02430.

glue-codes/openfast-cpp/src/OpenFAST.H
gantech: Agreed. Implemented in 9f34f9eb119d38dc0c78bfe58fdd23d084d02430.

glue-codes/openfast-cpp/src/FAST_Prog.cpp
gantech: Agreed. Implemented in 9f34f9eb119d38dc0c78bfe58fdd23d084d02430.

glue-codes/openfast-cpp/src/FAST_Prog.cpp
gantech: Agreed. I rectified a few of these automatic assumptions in 9f34f9eb119d38dc0c78bfe58fdd23d084d02430. However, a user shouldn't have to specify the turbine location for running a standard OpenFAST simulation through the C++ API with `InflowType` as `1`  in the OpenFAST `.fst` input file (specified using the InflowWind module). For actuator line simulations `InflowType` must be `2` and would require a turbine location to be specified. Unfortunately, I wouldn't know if a user specified  or `2` until the OpenFAST is initialized.

glue-codes/openfast-cpp/src/FAST_Prog.cpp
gantech: Agreed. Implemented in 9f34f9eb119d38dc0c78bfe58fdd23d084d02430. Reduced the code quite a bit.

glue-codes/openfast-cpp/src/FAST_Prog.cpp
```diff
@@ -8,25 +8,44 @@ inline bool checkFileExists(const std::string& name) {
...
-      fi.globTurbineData[iTurb].FASTInputFileName = "";
-  }
-  if (turbNode["restart_filename"]) {
-      fi.globTurbineData[iTurb].FASTRestartFileName = turbNode["restart_filename"].as<std::string>() ;
-  } else {
-      fi.globTurbineData[iTurb].FASTRestartFileName = "";
-  }
-  if ( (fi.globTurbineData[iTurb].FASTRestartFileName == "") && (fi.globTurbineData[iTurb].FASTInputFileName == "") )
+  get_if_present(turbNode, "turb_id", fi.globTurbineData[iTurb].TurbID, iTurb);
+  std::string emptyString = "";
```
sayerhs: Since you have a default `T()`, you don't need to pass `emptyString` in each call.

glue-codes/openfast-cpp/src/FAST_Prog.cpp
```diff
@@ -38,31 +57,12 @@ void readTurbineData(int iTurb, fast::fastInputs & fi, YAML::Node turbNode) {
...
-      fi.globTurbineData[iTurb].nacelle_area = 0.0;
-  }
-  if (turbNode["air_density"]) {
-      fi.globTurbineData[iTurb].air_density = turbNode["air_density"].as<float>();
-  } else {
-      fi.globTurbineData[iTurb].air_density = 0.0;
-  }
+  get_if_present(turbNode, "num_force_pts_blade", fi.globTurbineData[iTurb].numForcePtsBlade, 0);
+  get_if_present(turbNode, "num_force_pts_tower", fi.globTurbineData[iTurb].numForcePtsTwr, 0);
+  float fZero = 0.0;
```
sayerhs: No need for default, since that is the default parameter in the function definition. 

I think these parameters should be `get_required`?

gantech: I think these parameters should be `get_required` when reading in an input file through a CFD solver for actuator line simulations. 

`FAST_Prog.cpp` is only provided as an example and not intended to be a production level glue code using the C++ API. The executable generated using `FAST_Prog.cpp` takes no arguments and expects an input file specifically named `cDriver.i`. If we decide to make `FAST_Prog.cpp` a more production level glue-code, I think it'll require a few more changes.

glue-codes/openfast-cpp/src/FAST_Prog.cpp
gantech: Ah.. yes. I suppose I have to create another `get_if_present` with just 3 arguments.

sayerhs: You misunderstood me. The definition of `get_if_present` is

```
void get_if_present(const YAML::Node & node, const std::string& key, T& result, const T& default_if_not_present = T())
```

Which means if you call it with 3 arguments, it will default to empty string when `T = std::string`. You *don't* have to create another function with 3 arguments.

gantech: Ah.. got it. Thanks. Fixed in  https://github.com/OpenFAST/openfast/pull/145/commits/4f16f1b50b147880f05a7fda054d82ff5ec06801

rafmudaf:
@gantech how is this pull request going? Do you need anything from me to help it along?

gantech:
I should be merging this by tomorrow. I've run all the regression tests multiple times. I need to merge this along with a corresponding pull request in Nalu.

rafmudaf:
Ah ok cool. Let me know how I can support, if you need.

rafmudaf:
@gantech Is this pull request still valid? If you give me contributor access to your fork of OpenFAST, I'll resolve the merge conflicts.

Issue #2, opened 2017-01-06
DWM Driver


sayerhs:
DWM driver contains `USE ifport` which is specific to Intel Fortran Compiler. This code fails to compile with `gfortran`.

rafmudaf:
I've made some fixes to the DWM driver (use ntwc lib for input handling, comment the intel specific library imports because they don't seem to be used).

This can be seen at https://github.com/rafmudaf/openfast/tree/bugfix/dwm_driver

@ghaymanNREL @jjonkman @bjonkman does anyone have an input file for dwm driver?

rafmudaf:
Looks like it should be called:
`
    CHARACTER:: PriFile =  'DWM-driver\wind_farm.txt'
`

jjonkman:
@rafmudaf -- You can find an example wind_farm.txt file in the test directory of the DWM archive on NWTC Information Portal: https://nwtc.nrel.gov/DWM.

That said, I would not spend time working to resolve issues with DWM because--as mentioned in issue #93--we plan to replace DWM with FAST.Farm once it is publicly released.

rafmudaf:
Thanks @jjonkman. Since DWM will be deprecated by FAST.farm, I'll hold off on any further improvements to the DWM driver and leave this issue open for future reference until FAST.farm is released.

rafmudaf:
With @mylonasc's help, the dwm driver compiling bugs have been resolved for linux, mac, and windows. There are some hard coded paths so runtime errors are expected if these aren't updated in the driver source code.

I'm closing this issue since it is related specifically to compiling, and I will not open another issue for runtime problems since DWM will be deprecated per #93. If other issues arise regarding this, please comment on issue #93.

PR #146, opened 2018-06-25
[WIP] Fixed dwm-driver to compile on linux, as a temporary half-way solution until FAST.Farm is available.

mylonasc:
**DO NOT MERGE THIS YET!**
Dear OpenFAST team,

This  is a dirty fix for the DWM driver. This should compile ONLY on Linux (tested only on CentOS 7.4). You will notice there are debug messages (I am not experienced with fortran and I didn't have the time to learn a debugger).  I have left the debug messages so a potential other developer will not have to put them there again. 

Due to the fact that FAST.Farm is still not publicly released, I needed to fix this. This is shared so it may be useful for others. The results were briefly validated with what DWM gives on Windows.

The DFLIB was changed to IFQWIN where needed.
The hardcoded paths were changed to linux hardcoded paths so this runs on linux. 
I didn't have the patience to find out how to get platform independent paths in fortran.

This is not ready for merging - the paths should be platform independent.

We are sharing in hope that this is going to facilitate other researchers.

On behalf of the ETH-WINDMIL team,
Charilaos Mylonas


rafmudaf:
@mylonasc see https://github.com/OpenFAST/openfast/issues/2 and https://github.com/rafmudaf/openfast/tree/bugfix/dwm_driver.

If that does not solve your current problem, I'm happy to help find a full solution. As stated in issue #2, we don't intend to fix DWM driver since fast farm will be released soon(ish) so I'd like to close this pull request and move any discussion to issue #2.

PS the `Sys*.f90` configuration files set a `PathSep` variable that should be available in modules using NWTC Lib.

PR #147, opened 2018-06-27
Add a cmake flag for c-based shared libraries

rafmudaf:
Without this flag, OpenFAST does not link properly to `libmapcpp` when using gfortran on peregrine.


sayerhs:
Does it work without this flag in Intel compiler?

rafmudaf:
Yes

PR #150, opened 2018-07-12
Support a CMake-generated Visual Studio solution

rafmudaf:
This pull request updates the CMake configuration to better support the CMake-generated Visual Studio solution. Specifically included are:
- An update to the FindMKL function to support Windows Visual Fortran installations
- An explicit declaration for the output location of the registry executable
- A bug fix regarding a fortran preprocessor flag
- Minor improvements for cross platform compatibility
- Documentation reflecting these updates

The Visual Studio solution does not build the first time without error. As a workaround, the user must issue the build command in Visual Studio multiple times with the error count dropping each time. Eventually (after two or three builds), all libraries and executables will build without error.

### DIFF
--- a/cmake/FindMKL.cmake
+++ b/cmake/FindMKL.cmake
@@ -14,22 +14,22 @@
 # limitations under the License.
 #
 
-find_path(MKL_INCLUDE_DIRS
-  mkl.h
-  HINTS $ENV{MKLROOT}
-  PATH_SUFFIXES include)
+# find_path(MKL_INCLUDE_DIRS
+#   mkl.h
+#   HINTS $ENV{MKLROOT}
+#   PATH_SUFFIXES include)
 
 find_library(MKL_IFACE_LIB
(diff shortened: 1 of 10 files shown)


bjonkman:
@rafmudaf: are there plans to make this work with the 32-bit version of the Intel Visual Studio compiler on Windows? When I've tried to compile this on Win32, it fails to link because of the 64-bit versions of MKL libraries it is trying to use.

rafmudaf:
@bjonkman Good catch and thank you for pointing that out. I've opened issue [151](https://github.com/OpenFAST/openfast/issues/151) regarding this bug. Please add any further notes on that issue.

rafmudaf:
Also, I am curious if you have any insight into the problem described at the bottom of this section in the docs: https://openfast.readthedocs.io/en/dev/source/install/install_cmake_windows.html#cmake-build-instructions

Issue #130, opened 2018-05-29
Improved developer options/interface for BeamDyn


ashesh2512:
This issue is aimed at adding input file options that might help the user better diagnose standalone runs with  BeamDyn.

1) A finite difference routine is required for the developed to help compare the analytical BeamDyn formulation against a numerical counterpart. Relevant developer flags should be added to the input file.
2) A few variables are currently hard coded. Specifically, the maximum load retries and maximum static iterations. These should ideally be input file options.

PR #155, opened 2018-08-05
BeamDyn finite differencing capability and variational consistent output in summary file

ashesh2512:
This pull request corresponds to #130 and adds/improves the following capabilities:

1) Adding finite differencing capabilities to BeamDyn for developer debugging purposes. A central finite difference routine has been added for comparison of the analytical tangent stiffness matrix in static and dynamic cases. Input file user options have been added to turn finite differencing on/off and set other options such as perturbation size or maximum allowable relative difference. 

2) Computation of summary file options such as beam length, geometry of the beam, and center of mass were previously carried out using the cubic spline as the basis of beam geometry. While the cubic spline is used to compute the position of the FE nodes, the previous summary file was a misrepresntation of computations performed by BeamDyn as the system of equations is constructed using spectral finite elements by extension the beam length and center of mass. All these computations have been altered to be variational consistent. Consequently, the variable p%QuadPt has been removed because it did not make sense to compute position of quadrature points using the spline basis or use these locations to compute the above mentioned quantities. 

3) The Newton convergence criteria in both the static and dynamic cases has been updated to compare energy norms (F*u) rather than the sqrt(F*u) and sqrt(F*a) in the static and dynamic cases respectively. This causes 2 BeamDyn tests to fail because the sort in the convergence criteria means the Newton drop tolerance is squared. Theses changes are consistent with the widely used energy norm in FE literature for convergence of Newton iterations. Please note, regardless of changes made above, the code converges to the same solution. The test fails because currently we compare solution at all load steps and not just the converged solution.

4) Minor changes include - 
a) Adding an input file option for maximum load retries. Previously this was hard coded to 20 in the code. 

Once the above changes have been approved, I will be updating the documentation to reflect the updated convergence criteria and input file options. I will further submit a separate pull request for the failing tests. Only after the above will this pull request be merged.

The above changes also hope to start a discussion on the following topics - 

1) Adding an input file option in OpenFAST currently means changing all input files. Users not involved with a specific pull request may be unaware that their input files need to be updated alongside the code. I would strongly suggest moving to an avenue wherein not all input file options are required to run the code.

2) Currently we test for every single entry in the .out file as part of our regression tests. As engineers and applied mathematecian I believe it is important that we introduce a human element into deciding what entries should be tested for when implementing a regression test. e.g. In my opinion a regression test should not fail if the converged solution is unchanged.


ashesh2512:
The documentation for the updated convergence criteria can be viewed at - 
http://openfast-ashesh-fork.readthedocs.io/en/latest/source/user/beamdyn/theory.html#convergence-criterion-and-generalized-time-integrator

The documentation for the updated list of input file options can be viewed at - 
http://openfast-ashesh-fork.readthedocs.io/en/latest/source/user/beamdyn/input_files.html#simulation-controls

rafmudaf:
After running the regression test
```
The following tests FAILED:
	 25 - 5MW_Land_BD_DLL_WTurb (Failed)
	 31 - bd_curved_beam (Failed)
	 34 - bd_static_twisted_with_k1 (Failed)
```

### 5MW_Land_BD_DLL_WTurb
The differences here are negligible.
b3rootmzr

### bd_curved_beam
The solution converges to the baseline as described in point 3 in the pull request description.
screen shot 2018-08-27 at 2 31 00 pm

### bd_static_twisted_with_k1
The solution converges to the baseline as described in point 3 in the pull request description.
screen shot 2018-08-27 at 2 32 04 pm

bjonkman:
After quickly glancing through the code, I have a few comments:

1. Can we avoid writing to `*` and instead call `WrScr()`? Using `print *`' or `write(*)` make a lot of assumptions about the underlying (non-standard) write-to-console process. For example, it won't write to the Matlab console in the Simulink interface.

2. There are potential memory leaks in BD_DynamicSolutionGA2 with the new allocatable arrays there. Seems like the code would also be faster if those variables were part of the misc variable type. 

3. There are problems with the error handling in this PR. `BD_FD_Stat` doesn't check that the allocate statement was valid before using the variables. Same problem with `BD_DynamicSolutionGA2`.

ashesh2512:
Following the comments by @bjonkman and some discussion with @rafmudaf I have for the time being removed the write statements from this commit. The output was excessive and other information was easily lost when reading the terminal/log file for any dynamics case. 

I have further made the finite differencing arrays a part of the misc 'm' data structure to allow for more efficient usage of memory.

bjonkman:
@ashesh2512 I noticed that you removed the call to `SimStatus_FirstTime` in the BeamDyn driver, but you still call `RunTimes`. Variables `SimStrtTime` and `UsrTime2` get initialized in `SimStatus_FirstTime` and are used later in `RunTimes`, but now I don't see where they are initialized.

ashesh2512:
@bjonkman Thanks for catching that! Removing these calls wasn't intentional and probably happened when resolving merge conflicts while bringing my pull request up to date with the latest dev branch.

PR #157, opened 2018-08-15
Updates to AeroDyn (DBEMT), BeamDyn (linearization, quasi steady start up)

rafmudaf:
This pull request completes Envision Energy's pull request to merge a variety of updates. The initial pull request can be seen at [rafmudaf:6](https://github.com/rafmudaf/openfast/pull/6).

## Unit Test
All unit tests pass.

## Regression Test
Due to the quasi steady start up option in BeamDyn, the BeamDyn OpenFAST regression test case `5MW_Land_BD_DLL_WTurb` fails. However, the new solution shows improved startup transients and a consistent solution in time. For example, see the plot for `B1RootFzr` below. The baseline solutions for this test case have been updated.
b1rootfzr

Two linearization test cases have been added and the regression test framework is updated to compare the linearization matrices: `Ideal_Beam_Fixed_Free_Linear` and `Ideal_Beam_Free_Free_Linear`.

All other regression test cases pass through visual inspection of the results.


jjonkman:
@Raf -- Will you also add another r-test case for linearization of BeamDyn coupled to OpenFAST (i.e. one of the rotor speeds of our Campbell diagram test)?

rafmudaf:
@jjonkman is one rotor speed of that case supposed to take ~1 hour to run?

Also fyi "@raf" is the wrong account, I'm "@rafmudaf" :)

jjonkman:
It may take a while to run a time-domain simulation until all transients die out.  This was needed for the purposes of the verification of the eigensolutoin (against the equivalent ElastoDyn solution), but it is not needed within the r-test.  It is probably sufficient for the r-test to simulate for only one azimuth rotation (linearizing every 10 degrees or so of rotation).

Oops, I'll use @rafmudaf from now on :-).

rafmudaf:
@jjonkman I've added the OpenFAST/BeamDyn linearization case to the regression test system called `5MW_Land_BD_Linear`.

PR #159, opened 2018-08-21
BD bug fix: damping coefficients read in wrong order

bjonkman:
The damping coefficients were read in the old BeamDyn Z-X-Y coordinate system instead of the IEC (X-Y-Z) system it now uses.

All of the OpenFAST regression tests with BeamDyn use the value 1.0e-3 for all 6 damping coefficients, so this will not be noticed in any current tests.


jjonkman:
We found this bug when testing the new BeamDyn linearization functionality within OpenFAST.  I approve this pull request.

PR #160, opened 2018-08-22
Compile problem on gcc for single precision

NobuhiroKUSUNO:
Compilation can not be executed on the gcc environments

I have just tried gcc -6.3.0, -8.2.0. The other environments are not available for me.

The error messages are the below:

> ${HOME}/OpenFAST/modules-local/servodyn/src/ServoDyn_Driver.f90:118:100:
> 
>        call SrvD_JacobianPInput( Time, u(1), p, x, xd, z, OtherState, y, misc, ErrStat, ErrMsg, dYdu)
>                                                                                                     1
> Error: Type mismatch in argument ‘dydu’ at (1); passed REAL(4) to REAL(8)
> make[2]: *** [modules-local/servodyn/CMakeFiles/servodyn_driver.dir/src/ServoDyn_Driver.f90.o] Error 1
> 

This pull requset is just for the safe compilation for single precision version of openfast.

I have not examined the reg-tests yes, but this modifications works fine for me.


jjonkman:
Dear @NobuhiroKUSUNO,

I'm not really familiar with compiling with gcc (hopefully someone else can comment on that), but we can't approve this pull request based on my brief skim of your changes.  The choice of what KIND each REAL variable is declared as in OpenFAST (ReKi, R8Ki, SiKi, DbKi, or QuKi) was determined carefully based on consideration of the numerics each variable undergoes to ensure that the solution is accurate whether you compile in single or double precision.

Again, I'll let someone else comment on your issue associated with compiling OpenFAST unaltered with gcc.

Best regards,

NobuhiroKUSUNO:
Thank you for comments, @jjonkman.

I was afraid of the destroy of your careful choice of the precision. 
I will think out another way to compile the single-precision OpenFAST.

bjonkman:
@NobuhiroKUSUNO: I always compile OpenFAST in single precision (using Visual Studio/Intel Fortran), and do not get the kind of errors you are reporting. The error message you have says that the ServoDyn driver has an error. I think you could fix the problem with one line of code, which I have modified in https://github.com/OpenFAST/openfast/pull/161.

NobuhiroKUSUNO:
@bjonkman: I could compile OpenFAST  in single precision by GCC-8.2.0. Thanks!

PR #163, opened 2018-09-05
BeamDyn: Updated integration scheme for nodal load outputs

bjonkman:
@andrew-platt updated the integration scheme he used for the nodal load outputs in BeamDyn. The updated integration scheme better matches the analytical result:
regtest01_internalforcecalcimprove

### DIFF
--- a/modules-local/beamdyn/src/BeamDyn.f90
+++ b/modules-local/beamdyn/src/BeamDyn.f90
@@ -1469,7 +1469,7 @@ subroutine Init_MiscVars( p, u, y, m, ErrStat, ErrMsg )
       CALL AllocAry(m%elm,          p%dof_node,p%nodes_per_elem,p%dof_node,p%nodes_per_elem,    'elm',         ErrStat2,ErrMsg2); CALL SetErrStat( ErrStat2, ErrMsg2, ErrStat, ErrMsg, RoutineName )
 
          ! Point loads applied to FE nodes from driver code.
-      CALL AllocAry(m%PointLoadLcl, p%dof_node,p%nodes_per_elem,                       'PointLoadLcl',         ErrStat2,ErrMsg2); CALL SetErrStat( ErrStat2, ErrMsg2, ErrStat, ErrMsg, RoutineName )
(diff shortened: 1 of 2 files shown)


ashesh2512:
There are 3 minor merge conflicts resulting from the recently merged pull request #155. I have gone through the conflicts and once this pull request is approved, I should easily be able to resolve those conflicts before merge.

ashesh2512:
@bjonkman the code overall looks good to me. I have a few comments - 

1) Following pull request #155 the error calls at lines 3657 and 3663 can very well use RETURN instead of CYCLE because the calling function (BD_Static) now checks for errors and aborts immediately upon calling BD_StaticSolution. Would you agree ?

2) In the subroutine BD_diffmtc, the data structure p is used only to access nodes_per_elem. If so, I suggest passing in nodes_per_elem instead of p. This would make that routine modular and hence unit-test-able. 

3) Do you know if any of the regression test results are affected as a result of the update integration scheme ?

4) Do you (or @andrew-platt) and have any documentation regarding the  computation of BldInternalForceFE ? It would really helpful for those trying to dive into the code.

bjonkman:
@ashesh2512 

I don't have time to address about all of your comments today, but I did want to mention something regarding the error handling in your first comment: Your static algorithm is a little different than ours, but I do not think that #155 should have added a return after the call to `BD_StaticSolution`. Ignoring the error is intentional there (i.e., the lines 3512-3516 that you added to BeamDyn.f90 were intentionally left out) otherwise the `else` branch in the statement that follows is irrelevant. You could perhaps change the `if (piter .le. p%niter) then` to `if (ErrStat2 < AbortErrLev)` instead, but the error checks after BD_StatisSolution shouldn't occur.

Also, the lines you added at the end of that routine (lines 3562-3566 in BeamDyn.f90) are redundant given the two lines that follow them.

I think there are some other potential error handling issues in your BD_StaticSolution subroutine in #155 (again, related to returning on error with the loop counter less than the max specified in that loop).

ashesh2512:
@bjonkman 

The `else` condition still works because I removed line `3746` in `BD_StaticSolution` in #163 .  I believe that line wasn't significant anymore considering the fact when `(piter .le. p%niter)` we reduce the load and move on till the maximum number of load steps is reached. So we never really exit the code for a static simulation based on whether the maximum Newton steps have been reached.  

I agree with the redundancy of lines `3562-3565`. I will submit a pull request for that right away. Thank you for pointing it out.

ashesh2512:
This pull request has been modified in #265 due codes diverging significantly.

PR #164, opened 2018-09-06
Solve Option 2 Performance enhancements

bjonkman:
Parts of the SolveOption2 algorithm have been moved to the the AdvanceStates routine so that BeamDyn, InflowWind, AeroDyn, and ServoDyn use calculated inputs instead of extrapolations for inputs at t+dt. 

This change allows BeamDyn to use calculated rotations and AeroDyn/DBEMT to use calculated velocities, which otherwise would require a correction step. In DBEMT this was a problem with high turbulence and low wind speeds.


andrew-platt:
After chattting @ashesh2512, we should rerun a test with `BeamDyn` coupled to `OpenFAST` (with all forces and degrees of freedom off) and check if the residual is zero once this is merged into `dev`,

bjonkman:
As expected, due to the slight changes in inputs, the regression tests fail. However, the results do not look much different than the baseline. These differences should be similar to the what is seen if you add a correction step to the models (particularly the onshore models).

I would appreciate if this could be merged in the near future... it's been waiting more than a year.

rafmudaf:
We could add a `Theory` section here: https://raf-openfast.readthedocs.io/en/docs-fast8readme/source/user/openfast/index.html

rafmudaf:
@bjonkman are the updated AeroDyn input files already available?

bjonkman:
This PR doesn't change any input files. The updated AeroDyn input files for https://github.com/OpenFAST/openfast/pull/213 are here: https://github.com/bjonkman/r-test/tree/pullrequest/bjonkman-airfoil

bjonkman:
Thanks, @rafmudaf and @andrew-platt!

PR #168, opened 2018-09-08
Explicit type cast for compile error on gfortran

NobuhiroKUSUNO:
This pull request solves the compilation errors "passed REAL(4) to REAL(8)" on gfortran,
which is introduced by pull request #155.

I hope this dose not violate the numeric accuracy policy of OpenFAST. ^_^
Best regards,


bjonkman:
I think if we changed lines 224 and 225 in `Registry_BeamDyn.txt` to be type `R8Ki` instead of `ReKi`, the numerics would be better in BeamDyn.

@rafmudaf, could your nightly build process include building in both single and double precision so it catches these kinds of errors?

rafmudaf:
@bjonkman great idea. I've opened up an issue regarding this at https://github.com/OpenFAST/openfast/issues/169

rafmudaf:
This and other bugs are fixed by #184 as per @bjonkman's recommendation.

Issue #151, opened 2018-07-23
FindMKL does not search for 32 bit libraries


rafmudaf:
Currently, `FindMKL` is looking only for 64 bit MKL libraries, but this should instead be directed by the architecture type chosen in CMake configuration (whether default 64 bit or explicitly setting 32 bit).

A current workaround is to explicitly pass the BLASLIB variable to cmake:
```
cmake .. -DBLAS_LIBRARIES=$BLASLIB -DLAPACK_LIBRARIES=$BLASLIB
```

where $BLASLIB contains something like 
```
-Wl, C:\Program Files (x86)\IntelSWTools\compilers_and_libraries\windows\mkl\lib\intel64 -Wl, C:\Program Files (x86)\IntelSWTools\compilers_and_libraries\windows\compiler\lib\intel64 -LC:\Program Files (x86)\IntelSWTools\compilers_and_libraries\windows\mkl\lib\intel64 -LC:\Program Files (x86)\IntelSWTools\compilers_and_libraries\windows\compiler\lib\intel64 -lmkl_intel_lp64 -lmkl_intel_thread -lmkl_core
```
An update to the documentation detailing how to explicitly set the math libraries should be included in the fix to this issue.

rafmudaf:
Fixed by #174

PR #174, opened 2018-09-12
Include 32 bit libraries in the MKL search

rafmudaf:
Fixes issue #151

### DIFF
--- a/cmake/FindMKL.cmake
+++ b/cmake/FindMKL.cmake
@@ -19,17 +19,25 @@
 #   HINTS $ENV{MKLROOT}
 #   PATH_SUFFIXES include)
 
+# infer the architecture build type
+# https://cmake.org/cmake/help/v3.0/variable/CMAKE_SIZEOF_VOID_P.html
+if("${CMAKE_SIZEOF_VOID_P}" STREQUAL "4")
+  set(ARCH "32")
+elseif("${CMAKE_SIZEOF_VOID_P}" STREQUAL "8")
+  set(ARCH "64")
+endif()
+
 find_library(MKL_IFACE_LIB
-  NAMES mkl_intel_lp64 libmkl_intel_lp64.a mkl_intel_lp64_dll.lib
-  PATHS $ENV{MKLROOT}/lib $ENV{MKLROOT}/lib/intel64 $ENV{INTEL}/mkl/lib/intel64 $ENV{INTEL}/mkl/lib/intel64_win)
+  NAMES mkl_intel_lp${ARCH} libmkl_intel_lp${ARCH}.a mkl_intel_lp${ARCH}_dll.lib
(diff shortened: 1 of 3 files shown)


rafmudaf:
32 bit mkl libraries are now found for 32 bit installations. However, if both 64 bit and 32 bit libraries are installed, cmake may choose the 64 bit versions of `mkl_core` and `mkl_sequential` depending on the user's path configuration. In this case, set `MKL_SEQ_LIB` and `MKL_CORE_LIB` manually.

Additionally, documentation is included describing the process for setting the above and any other cmake flag explicitly through the command line.

saeidfn:
Hi all,
I reached to the step "2.2.4.1. Running CMake" and I've been trying to run CMake with the syntax "cmake .. -DBUILD_SHARED_LIBS=ON", but the response I got is as follows:

E:\OpenFAST\build>cmake .. -DBUILD_SHARED_LIBS=ON
-- Selecting Windows SDK version 10.0.18362.0 to target Windows 10.0.18363.
**-- Could NOT find MKL (missing: MKL_LIBRARIES MKL_IFACE_LIB MKL_SEQ_LIB MKL_CORE_LIB)**
-- A library with LAPACK API found.
-- Setting system file as: src/SysIVF.f90
-- Configuring done
-- Generating done
-- Build files have been written to: E:/OpenFAST/build

I already installed "Math Kernel Library" on my system, but I still don't know how to solve the error in line 2 (bold text).
Would you mind helping me with how to tackle this issue??

rafmudaf:
Are you in the correct terminal? Depending how you've configured your Intel MKL installation, you may need to use the Intel Visual Studio Command Prompt in order for MKL to be found. You can also configure Powershell or whatever you're using to load the batch scripts in the Intel installation location.

Another thing you might try is to pass the location of the MKL libraries directly to CMake as noted in the docs: https://openfast.readthedocs.io/en/master/source/install/index.html#id3.

saeidfn:
Dear Rafael,
what do you mean by the correct terminal??
By your explanations, I could find "Intel Visual Studio Command Prompt" but I don't know what would be the next step to find MKL.

rafmudaf:
Try running CMake from that terminal since it should have the correct paths loaded to find the Intel libraries.

By the way, Anaconda Prompt, Command Prompt, the Intel Visual Studio Command Prompt, Powershell ... these are all different terminal applications. They all exist because they have different settings and configurations for their environment. The Anaconda Prompt has Anaconda preloaded. The Intel Visual Studio Command Prompt has Intel and Visual Studio executables and libraries in its path.

saeidfn:
Actually I didn't understand your explanation. Initially, I ran CMake with the Command Prompt terminal at the E:OpenFAST/build directory with the syntax "cmake .. -DBUILD_SHARED_LIBS=ON" and I've got the mentioned error. By your explanation, I couldn't still understand how to find MKL Libraries.
In all of the steps, I tried to use Command Prompt and I haven't used any other options.

Issue #169, opened 2018-09-10
Nightly testing does not test single precision


rafmudaf:
The existing nightly testing capability does not test single precision so any errors generated there (for example, in setting precision) are not caught.

Tests:

- [ ] Single precision compile

- [ ] Single precision regression test (can the double precisions baselines be used with a higher tolerance or does this require a set of single precision baselines?)

rafmudaf:
Here's @bjonkman's https://github.com/OpenFAST/openfast/pull/184#issuecomment-427067955 on precision using `PI` and `PI_D` in the unit tests:

One comment on the precision conversions in the BD unit tests: You should use real(Pi_D, BDKi) instead of instead of real(Pi, BDKi) because in single precision the 4-byte Pi converted to 8-bytes will not be the same as pi computed in 8-byte precision in BeamDyn. Pi_D is the double-precision version of Pi.

When you do run single-precision tests, you will notice a problem (especially with the trig functions using these as inputs to compare with double-precision BeamDyn).

ebranlard:
When the time comes to work on that, I bumped into two issues, likely related to AeroDyn, while using a Win32, single precision, and debug build. 

These issues where quite hard to debug and I still haven't found the sources of them. Putting print statements in `BEMTU_Wind` seems to delay the error, but it still occurs somewhere else in the code. 

I'm enclosing these two cases for references: 
-  Case Wind0: access violation occurs when inflow wind is not used or if the wind is very small. 
-  Case TwrPotent: acceess violation occurs when TwrPot = 0
Theses cases do not fail with double precision or 64 bit. 

[AccesssViolations.zip](https://github.com/OpenFAST/openfast/files/3129529/AccesssViolations.zip)

PR #184, opened 2018-10-03
Fix single precision build bugs

rafmudaf:
This pull request improves on #168 regarding single precision build bugs introduced in #155. There are also improvements to the BeamDyn unit tests allowing them to build properly. This pull request **does not** address the question of tolerance in single precision for any of the tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,10 +58,10 @@ matrix:
 
 script:
     # beamdyn unit tests
-    - ctest -R beamdyn_utest
+    - if [[ "$DOUBLE_PRECISION" == "ON" ]]; then ctest -R beamdyn_utest; fi
 
     # beamdyn regression tests
-    - ctest -R bd_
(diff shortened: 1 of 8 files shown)


bjonkman:
One comment on the precision conversions in the BD unit tests: You should use `real(Pi_D, BDKi)` instead of instead of `real(Pi, BDKi)` because in single precision the 4-byte `Pi` converted to 8-bytes will not be the same as pi computed in 8-byte precision in BeamDyn. `Pi_D` is the double-precision version of `Pi`.

When you do run single-precision tests, you will notice a problem (especially with the trig functions using these as inputs to compare with double-precision BeamDyn).

rafmudaf:
Good point, thank you for the insight. The tests included in OpenFAST don't fully support single precision at the moment. I'm moving your comment to issue #169 since that one deals with single precision testing.

PR #186, opened 2018-10-04
Modified limits in DBEMT for numerical robustness 

bjonkman:
The time-varying `tau1` in DBEMT has tighter limits to ensure that it doesn't exceed 100 s. The calculation also limits of value of rotor-averaged axial induction to be a maximum of 0.5 instead of 0.7 and the induced axial relative air speed to be a minimum of 0.1 m/s instead of 0.0001 m/s. 

This helps in cases where extremely large values of `tau1` at initialization take a long time to settle out. It would probably be helpful to add a low-pass filter on `tau1`, but this request ignores that.


rafmudaf:
The current test suite won't test this change as none use DBEMT. Everything looks logical, so I'm merging this as is.

PR #190, opened 2018-10-16
Minor fixes

friedrichatgc:
Some minor fixes (mainly compiler warnings). Please see the commit messages for details.

### DIFF
--- a/modules-local/aerodyn/src/UnsteadyAero.f90
+++ b/modules-local/aerodyn/src/UnsteadyAero.f90
@@ -826,7 +826,7 @@ subroutine UA_Init( InitInp, u, p, xd, OtherState, y,  m, Interval, &
    type(UA_ParameterType),       intent(  out)  :: p           ! Parameters
    type(UA_DiscreteStateType),   intent(  out)  :: xd          ! Initial discrete states
    type(UA_OtherStateType),      intent(  out)  :: OtherState  ! Initial other states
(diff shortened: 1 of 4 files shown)


rafmudaf:
Ok, so those arguments should not be removed since there's a design decision dictating the subroutine signature. However, changing an `intent(out)` argument to `intent(inout)` to silence a compiler warning seems unsafe in the registry and inconsistent with the rest of the OpenFAST api in UnsteadyAero so lets leave that out.

I do agree with the change in line 2010 and 2011 in [NWTC_IO.f90](https://github.com/OpenFAST/openfast/pull/190/files#diff-6dd4b6447221394b6ff84ad08cc4c87cL2010).

Can a Window's user (@ghaymanNREL @bjonkman) provide a review on the registry updates?
Those changes are [here](https://github.com/OpenFAST/openfast/pull/190/files#diff-70c5ff96b02bce4a8ee2b5b0d8a01c5bR8) and [here](https://github.com/OpenFAST/openfast/pull/190/files#diff-0f66494c99829636c58188be7f94e923R8).

rafmudaf:
Also just FYI there's a nice [FAST Programmer's Handbook](https://nwtc.nrel.gov/system/files/ProgrammingHandbook_Mod20130717.pdf) available which has a template module in Appendix E. That module and the rest of the guide are good references for the modularization framework and how data should be passed.

PR #191, opened 2018-10-16
Fixed undefined behaviour in VariousWaves_Init.

friedrichatgc:
InitOut%WaveDirArr(InitOut%NStepWave2) is not set in case of
multi-directional waves. This leads to undefined behaviour (read of
uninitialized value).

Set this value to 0 as workaround (this generates defined behaviour).
The correct fix is unclear: maybe InitOut%WaveDirArr(0) should be set
to 0 and the index K starting from 1 then???

### DIFF
--- a/modules-local/hydrodyn/src/Waves.f90
+++ b/modules-local/hydrodyn/src/Waves.f90
@@ -1642,6 +1642,8 @@ SUBROUTINE VariousWaves_Init ( InitInp, InitOut, ErrStat, ErrMsg )
             ENDDO
          ENDDO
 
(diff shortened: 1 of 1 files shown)


ebranlard:
I confirmed that is still the case with version 2.1. 

The issue is that the last value of `WaveDirArr `is never set in the loop over `WvSpreadFreqPerDir`. 

The value should not be set to zero but set to a random value of `WvTheta`.

I'll send another pull request to fix this.

ebranlard:
I would cancel this pull request and merge #237 instead.

rafmudaf:
@friedrichatgc thanks for your suggestion. Pull request #237 fixed this issue.

Issue #71, opened 2017-11-22
The Renolds number not calculated correctly in BEMTUncoupled.f90


gerwangcy:
Dear developers,

I find an error regarding the Renolds number calculation in the file:
modules-local/aerodyn/src/BEMTUncoupled.f90,  line 96, which is: 
`Re = airDens * W * chord / mu`
It seems to be problematic, because the variable 'mu' here is actually the kinematic viscosity, rather than the dynamic viscosity. This line is within the subroutine:
`subroutine BEMTU_Wind( axInduction, tanInduction, Vx, Vy,  chord, airDens, mu, W, Re )`
which is called by many other functions, an example is in the file:
modules-local/aerodyn/src/BEMT.f90,  line 977, which is:
`call BEMTU_Wind( 0.0_ReKi, 0.0_ReKi, u1%Vx(i,j), u1%Vy(i,j), p%chord(i,j), p%airDens, p%kinVisc, Vrel, Re )`
The seventh argument is kinematic viscosity instead of dynamic viscosity! It is true for other calls of the function. The easiest way to correct may be just to replace the line 96 of BEMTUncoupled.f90 with:
`Re =  W * chord / mu`

It is also the case for fastv8 repository. I hope this could help. Thanks.

Best,
Chengyu Wang

rafmudaf:
Fixed by #195

PR #195, opened 2018-10-18
AD bug fix: calculation of Re

bjonkman:
BEMTU_Wind was calculating Re based on dynamic viscosity instead of kinematic viscosity as reported here: https://github.com/OpenFAST/openfast/issues/71

I fixed the calculation and cleaned up calling routines that don't need the `airDens` variable anymore.

This should not currently change any results except for the Reynolds number output channels (local to blade or tower nodes) from AeroDyn.


rafmudaf:
All regression tests pass so I'll go ahead and merge. Thanks @bjonkman !

Issue #196, opened 2018-10-18
Openfast crashes on linux with turbulent wind and tower aerodynamics turned off


evan-gaertner:
Openfast crashes with turbulent wind (inflowwind WindType = 3) on linux with tower aerodynamics turned off (TwrPotent=0, TwrShadow=False, TwrAero=False).  The error does not occur when tower aerodynamics are on.  It first came up on Peregrine and I was able to replicate it on my laptop in Ubuntu for openfast v1.0.0, v2.0.0, and the current dev branch.  It is not a problem on Windows, compiled with cygwin.

>  ./openfast /mnt/c/linux/WT_Codes/models/openfast2/r-test/glue-codes/openfast/5MW_Land_DLL_WTurb/5MW_Land_DLL_WTurb.fst
> 
>  **************************************************************************************************
>  OpenFAST
> 
>  Copyright (C)  National Renewable Energy Laboratory
>  Copyright (C)  Envision Energy USA LTD
> 
>  This program is licensed under Apache License Version 2.0 and comes with ABSOLUTELY NO WARRANTY.
>  See the "LICENSE" file distributed with this software for details.
>  **************************************************************************************************
> 
>  OpenFAST-v2.0.0
>  Compile Info:
>   - Architecture: 64 bit
>   - Precision: single
>   - Date: Oct 18 2018
>   - Time: 13:06:03
>  Execution Info:
>   - Date: 10/18/2018
>   - Time: 13:59:24-0600
> 
>  OpenFAST input file heading:
>      FAST Certification Test #18: NREL 5.0 MW Baseline Wind Turbine (Onshore)
> 
>  Running ElastoDyn.
>  Running AeroDyn.
>  Running AirfoilInfo.
>  Running BEM.
>  Running UnsteadyAero.
>  Warning: Turning off Unsteady Aerodynamics because C_nalpha is 0.  BladeNode = 1, Blade = 1
>  Warning: Turning off Unsteady Aerodynamics because C_nalpha is 0.  BladeNode = 2, Blade = 1
>  Warning: Turning off Unsteady Aerodynamics because C_nalpha is 0.  BladeNode = 3, Blade = 1
>  Warning: Turning off Unsteady Aerodynamics because C_nalpha is 0.  BladeNode = 4, Blade = 1
>  Warning: Turning off Unsteady Aerodynamics because C_nalpha is 0.  BladeNode = 1, Blade = 2
[...]

PR #201, opened 2018-11-09
Bug fix in allocating tower elevation array

rafmudaf:
This pull request moves an array allocation into an if-statement which determines whether the array should be allocated.

The problem arises when `NumTwrNds` is not 0 in the input but this block at [AeroDyn.f90L928](https://github.com/OpenFAST/openfast/blob/dev/modules-local/aerodyn/src/AeroDyn.f90#L928) executes:
```
   if (p%TwrPotent == TwrPotent_none .and. .not. p%TwrShadow .and. .not. p%TwrAero) then
      p%NumTwrNds     = 0
```
This causes `InitOut%TwrElev` to be allocated to an array of length 0 while `InputFileData%TwrElev` is nonzero so this line at [AeroDyn.f90L239](https://github.com/OpenFAST/openfast/blob/dev/modules-local/aerodyn/src/AeroDyn.f90#L239) is a seg fault:
```
InitOut%TwrElev(:) = InputFileData%TwrElev(:)
```

All regression tests pass since none test this particular case.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ env:
 # mac configuration
 before_install:
     # first uninstall a conflicting package
-    - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew cask uninstall oclint; fi
+    # - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew cask uninstall oclint; fi
 
     # install required packages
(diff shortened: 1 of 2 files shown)


rafmudaf:
This also fixes a small bug in the TravisCI configuration and fixes issue #196.

PR #203, opened 2018-11-14
Continuous integration improvement

rafmudaf:
This pull request expands the TravisCI configuration to build with Intel Fortran for linux. This is helpful in catching any ifort-specific compile-time and run-time problems as in the case of pull request #205.

In configuring this TravisCI configuration, I found a bug where a variable was unallocated in BeamDyn, so that too has been fixed here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ before_install:
 
 install:
     - mkdir build && cd build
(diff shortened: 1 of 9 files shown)


rafmudaf:
@ashesh2512 could you please review the change to BeamDyn here?

PR #207, opened 2018-11-27
HydroDyn: check for valid file unit before attempting to close

rafmudaf:
Turning off `Hdsum` in HydroDyn leads to a seg fault since HydroDyn tries to close a file which is not opened to begin with. This pull request checks for a valid file unit before attempting to close the HydroDyn summary file.

This was reported at https://wind.nrel.gov/forum/wind/viewtopic.php?f=4&t=1111&start=15.


rafmudaf:
@jjonkman good call, I've made that change. I'll merge after the automated tests complete.

PR #204, opened 2018-11-15
OpenFAST-Simulink interface

bjonkman:
I'm making a pull request on this code so I can delete my branch that reinstates the FAST8 Simulink interface capability to OpenFAST that I helped several people at NREL with. 

It could be cleaned up a bit for consistency between the vs-build and cmake build by someone who wants to spend time on that. (e.g., `create_FAST_SFunc.m` uses the naming convention for the vs-build output for win32 and win64, but uses names from cmake build [and manual file moves?] for mac i64) And as far as I know, there is no documentation in OpenFAST on the Simulink interface, yet (see https://github.com/OpenFAST/openfast/issues/177).

This should resolve https://github.com/OpenFAST/openfast/issues/59 and https://github.com/OpenFAST/openfast/issues/156.

PR #213, opened 2018-12-06
Added option for 2D airfoil interpolation

bjonkman:
## Summary
This pull request adds 2D interpolation capability to AeroDyn airfoils, based on code @ghaymanNREL sent me earlier this year. 

Previous versions interpolated airfoils within a table based on angle of attack. This version also interpolates between airfoil tables based on Reynolds number or a user-control parameter.

**The AeroDyn input file has changed.** New input files for the regression test cases can be found here: https://github.com/bjonkman/r-test/tree/pullrequest/bjonkman-airfoil. None of the regression tests currently use 2D airfoil interpolation, but the SWRT airfoil files do contain data for multiple Reynolds numbers, so they could be modified to use the additional tables.

Regression tests pass if built with `-DSPLINE_INTERP`, which overrides the default airfoil interpolation method, which has changed. 

### AeroDyn changes:
- There is a new line for the `AFTabMod` input parameter in the AeroDyn input file. 
- AeroDyn has a new input for a User Control property, which is passed to AirfoilInfo so it can be used for 2D interpolation.  The input is currently set to 0, and would need to be set in source code for a specific project.
- Types in the AirfoilInfo (AFI) module were changed, requiring changes to many other parts of AeroDyn, its driver, and the UA driver.
- The default interpolation on angle-of-attack in the airfoil tables has changed from cubic to linear. 
  **This will change regression-test results** if the interpolation order is not specified in the airfoil files.   (This includes all of the regression tests that use AeroDyn.)
- The single type defined in fmin_fcn.f90 was moved to mod_root1dim.f90
- Routine `UncoupledErrFn()` was replaced with `BEMTU_InductionWithResidual(`), which already existed and performed the same basic tasks.
- Routines `GetSteadyOutputs(`) in UA and `ComputeSteadyAirfoilCoefs(`) in BEMT were removed and replaced with `AFI_ComputeAirfoilCoefs()`.
- `AFI_GetAirfoilParams(`) was replaced with `AFI_ComputeUACoefs()`.
- The UA computation for `C_nalpha_circ` was removed from `AFI_ComputeUACoefs()` and placed in UA's `ComputeKelvinChain()` instead.
- The UA module now validates the UA parameters from AFI and checks if UA should be shut off based on the AFI parameters (instead of having AeroDyn do this)
- The AFI module validates its own data (with the exception of parameters in UA tables)

### OpenFAST Registry changes:
- Fixed an issue so that ExtrapInterp routines aren't generated for types that don't exist (-noextrap flag isn't strictly necessary anymore, 
  but using it will result in smaller files if types exist but ExtrapInterp isn't necessary for the module)
- added ExtrapInterp for two specific types in airfoilinfo (this module name and its types are hardcoded)

### Visual Studio build changes:
- added checks for unused variables in non-*_Types.f90 files in debug mode.

### General changes:
- removed some unused variables
- updated error handling
- fixed spelling errors

### The following documentation has been updated:
- text for the new AeroDyn input and interpolation capability
- example AeroDyn input files
- [bug fix] example BeamDyn input files that were not consistent with previous changes to BeamDyn

### The following source file are not used but are still in the repo, in case they are useful elsewhere:
- NetLib/Dierckx_FitPack/dierckx_fitpack.f
- NetLib/Dierckx_FitPack/NWTC_FitPack.f90

### The following source file are unnecessary and were deleted from the repo:
- modules-local/aerodyn/src/fmin_fcn.f90
- modules-local/aerodyn/src/ToDoList.txt

## One more thing...
The BEMT solution computes the Reynolds number using axial and tangential induction factors set to 0. Then, it gets cl and cd from the airfoil tables (potentially based on Reynolds number), which allow it to compute the axial and tangential induction factors. The BEMT solve would have to be modified to solve these items simultaneously. Apparently there was discussion at NREL that using induction factors or zero was an okay approximation for Reynolds number (for now).


andrew-platt:
@rafmudaf: I think there is a build issue here similar to the PR I submitted yesterday.  `CMAKE_Fortran_COMPILER not set`

abhineet-gupta:
Are there any plans to add simulink build capability to this version?

bjonkman:
Hopefully both https://github.com/OpenFAST/openfast/pull/204 and https://github.com/OpenFAST/openfast/pull/213 will be merged into the OpenFAST dev branch in the near future, which will give you both 2D airfoil interpolation and Simulink build capability. You could alternatively merge the branch from https://github.com/OpenFAST/openfast/pull/204 into your own copy of this one to have both capabilities while waiting for the NREL merge.

andrew-platt:
I made a minor update in the comment for `InterpOrd` in the airfoil tables on `r-test`. Do we need to point to the new git hash for that?

rafmudaf:
Yep, you'll need to commit the change to the r-test commit within OpenFAST.

From `openfast/` and with the r-test submodule on the correct commit, do this:
```
git add reg_test/r-test
git commit -m "Your message"
```

andrew-platt:
I checked the results using the `SPLINE_INTERP` compiler flag (this currently sets the default interpolation to cubic for testing -- this was the previous default).  This gives a very close match to the baseline results (3 offshore tests fail due to a very small timeshift).

I also checked the results with the new default interpolation.  The results mostly show a timeshift relative to the baseline for most cases with very minor changes in loads.  I would expect this to happen.

So, I propose we rebaseline the results, and remove the `SPLINE_INTERP` preprocessor directive and merge this PR.

ebranlard:
that sounds good to me

andrew-platt:
Some polars will be more sensitive to changing from cubic to linear interpolation than others.  For the `AWT27_35` polar, this will result in slightly different values for `Cl`, particularly around the sharp transitions.  Cubic interpolation has a tendency to produce overshoot.

AirFoil_Polar_Cl_AWT27_35
Figure: Absolute value of lift curve from AWT27_35 polar.  Blue indicates positive lift, red indicates negative lift.  Linear interpolation near 0 degrees angle of attack will yield different Cl values than cubic interpolation.

PR #218, opened 2018-12-13
Work in progress to make beamdyn more cache friendly 

ashesh2512:
BD_GA2 has been identified as the most computationally intensive subroutine. This pull request is a step towards making BeamDyn more cache friendly for reduced time to solution.

### DIFF
--- a/modules-local/beamdyn/src/BeamDyn.f90
+++ b/modules-local/beamdyn/src/BeamDyn.f90
@@ -2927,14 +2927,14 @@ SUBROUTINE BD_AssembleStiffK(nelem,p,ElemK,GlobalK)
 
    INTEGER(IntKi)                            :: i
    INTEGER(IntKi)                            :: j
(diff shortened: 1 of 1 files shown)


ashesh2512:
Please do not merge this branch yet. As discussed with @sayerhs, to get full benefit of the SIMD vectorizations, we will have to provide alignment hints to the compiler which would require us to have `*_Types.f90` be not overwritten.

andrew-platt:
I'm rather curious on some of the indices you are changing.  We did a fair amount of rearranging these indices to improve the speed a few years ago.  The index order in fortran is the faster changing index is first (opposite to the C / C++ convention).  So you will want to be careful with changing any of them as they may not do what you are expecting.

ashesh2512:
Work on this pull request has not taken off, and given the recent changes in the code elsewhere, these changes may not be the most efficient any longer. I am hence closing the pull request for now.

PR #220, opened 2018-12-19
[BugFix] BeamDyn damping coefficients

andrew-platt:
The damping coefficients were in the in incorrect reference frame.  This PR corrects this.

@ashesh2512, can you verify this produces the same results as the linearization results from yesterday?

My apologies for not fully testing this (it has been verified in the Envision-Energy code base and cherry-picked for this PR).

### DIFF
--- a/modules-local/beamdyn/src/BeamDyn.f90
+++ b/modules-local/beamdyn/src/BeamDyn.f90
@@ -2328,7 +2328,10 @@ SUBROUTINE BD_StifAtDeformedQP( nelem, p, m )
 
    INTEGER(IntKi)                :: idx_qp         !< index counter for quadrature point
    INTEGER(IntKi)                :: temp_id2       !< Index to last node of previous element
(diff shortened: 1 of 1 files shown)


andrew-platt:
It looks like the build system has some issues with finding `ifort` and `gcc`.

ashesh2512:
It does look like that, however when when the pre-build steps went through, the code did compile without any issues.

rafmudaf:
Its possible that the travis ci container changed. I restarted the failing tests to verify.

rafmudaf:
It seems that Intel is on my trail. I've requested more than a few free trials of the intel compiler so I can develop the TravisCI workflow, but the latest one is now expired even though its only a few days old :( I'll work this out and update the TravisCI configuration asap.

As for the GCC problem, these containers aren't super reliable and sometimes error for weird reasons. Rerunning that job fixed it.

PS Sorry if you got a bunch of emails from TravisCI, @andrew-platt, I triggered a few jobs while finding the problem.

PR #222, opened 2018-12-21
BeamDyn driver: option to set initial root orientation separately from the global rotation

andrew-platt:
BeamDyn: Add initial azimuth to driver

Added a way to set the initial rotation separately from the global orientation. This is a flag added to the input file for the driver (`GlbRotBladeT0`).

When true, the orientation as the blade root orientation and calculation reference frame orientation.
If the flag is false, it is used only as the blade root orientation (identity matrix for the calculation reference frame orientation).

beamdyn azimuth 90
Figure: the initial orientation of the NREL 5 MW blade is [[1 0 0] [0 0 1] [0 -1 0]] corresponding to an azimuth of 90 degrees.  The reference is the white line.  A tip load in the +z direction (global) of 3E3 is applied, with no gravity. The blade root is displaced by 1 meter from [0,0,0] then rotated.

Notes:

- this is a dynamic simulation, so the blade is deflected further than equilibrium).
- the output shown is from the envision code which includes vtk outputs (checked against tip displacements in OpenFAST code).

Test cases and updated driver input files for this are in a separate pull request on the r-test repository.


andrew-platt:
Gravity loading only comparison for 0 and 90 azimuth.

tdxr

tdyr

PR #97, opened 2018-03-06
Update of Unsteady Aero model UAMod=2, corrections and added parameters

alvarogonzalezsalcedo:
Some changes related to the correct implementation of the code DYSTOOL in the option UAMod=2 of FAST have been performed. Most of them here and there are corrections, but there are also changes related to the application of parameters similar to f in the classical BL theory to Cd and Cm apart from Cl. The value of those parameters is modified depending on the flow separation progress. A standard selection of values for each range is hard-coded, for a performance similar to the original code.

PR #226, opened 2019-01-01
Unsteady Aero changes

bjonkman:
# Description
This pull request is based on changes to unsteady aero code from 
1. pull request https://github.com/OpenFAST/openfast/pull/97, which introduces changes to the Gonzalez model (UAMod=2) 
2. changes emailed to me from @ghaymanNREL a year(?) ago, based on meetings and conversations between NREL and Envision, and 
3. a modification to `Cn_FS` from Envision (also previously discussed with NREL)

This code is uses the 2D airfoil interpolation in the yet-unmerged pull request https://github.com/OpenFAST/openfast/pull/213.

# Results 
For comparisons, here are some plots generated using the UnsteadyAero Driver code with the DU21_17.dat airfoil file from the NREL 5MW model:
Old code with UAMod=2:
ua2-dev
New code with UAMod=2:
pr2-dev

Old code with UAMod=3:
ua3-dev
New code with UAMod=3:
pr3-dev


rafmudaf:
@jjonkman @ghaymanNREL I've added you as reviewers here but feel free to pass that on to someone else

andrew-platt:
I vote to approve this and update the regression tests accordingly.

bjonkman:
@andrew-platt I think the SWRT model always has weird results. That is probably partly due to the fact that it's supposed to be furling, but that capability doesn't exist in OpenFAST. I think there is an instability in the SWRT model, and almost any change to code can give it very different results.  

The AOC model also has some weird aero because AD15 thinks the tower diameter is 6 cm (to get equivalent tower shadow as what was used in AD14).

Issue #216, opened 2018-12-11
Rebuild DISCON_*.dll using cmake


liuyingyi-ukyushu:
Dear Rafeal,

I followed your instructions via our email communications below to rebuild the DISCON.dll (For the convenience of other users, I repost the information here):
-----------------------------------------------------------------------------------------------------------------
"Hello Yingyi,
 
As Jason mentioned, the compiled libraries are included in the latest OpenFAST release for both 32 bit and 64 bit architectures. If you are on Windows, using cmake is a little bit more complicated since you have to do the environment configuration that Visual Studio does for you. The simplest way to do this is to find the Visual Studio command prompt installed with intel tools; mine is called “Compiler 17.0 Update 7 for Intel 64 Visual Studio environment”. Then, navigate to one of the DISCON directories your listed below and do these steps:
 
mkdir build
cd build
cmake .. -G   # this will list the available visual studio generators on your system
cmake .. -G”Visual Studio 15 2017 Win64”   # this is what I use
start DISCON.sln   # this opens the discon solution file and you can compile using visual studio from there
 
The process is similar for a linux machine except leave off the -G”Visual Studio” and use “make” to compile instead”.
 
Please let me know if you run into any problems.
 
Thanks
Rafael M Mudafort"
-------------------------------------------------------------------------------------------------------------------
Based on your inputs, I installed the latest cmake for windows (cmake-3.13.1-win64-x64.msi) from here: https://cmake.org/download/. I choose the newly installed cmake to be included in the system path.

On my desktop, the only installed fortran compiler is "Intel Parallel Studio XE 2018", therefore, I navigate to its "Command Prompt with Intel Compiler 18.0 Update 2" at the location of "C:\ProgramData\Microsoft\Windows\Start Menu\Programs\Intel Parallel Studio XE 2018\Compiler and Performance Libraries", I choose the Intel 64 Visual Studio 2015 environment. A command window is opened which is named "Intel Compiler 2018  Update 2 Intel(R) 64 Visual Studio 2015", which means that I'm using an Intel Compiler 2018 (Compiler 18.0) under the environment of Visual Studio 2015 (VS2015). Sorry for the prolixity.

In the command window, I navigate to the path where "CMakeLists.txt" and "DISCON_OC3Hywind.F90" locate, followed by typing "mkdir build" and "cd build". When I type "cmake .. -G”Visual Studio 15 2017 Win64”, an error occurs reporting that: 

-----------------------------------------------------------------
[...]

PR #228, opened 2019-01-03
Update the CMake configuration for the included DISCON projects

rafmudaf:
Updates are included in both the r-test submodule and the “share” directory. This improves the cmake configurations for linux, mac, and windows, and fixes issue #216.

### DIFF
--- a/share/discon/CMakeLists.txt
+++ b/share/discon/CMakeLists.txt
@@ -14,6 +14,49 @@
 # limitations under the License.
 #
 
+# Helper macros for setting appropriate compiler flags
+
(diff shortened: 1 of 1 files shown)


rafmudaf:
The last commit here fixes @bjonkman's comment in #216 about the discon libraries not being statically linked. As @bjonkman points out, the discon libraries distributed in the v2.0.0 release do not execute on a windows machine which does not have intel installed.

Thank you very much for pointing out the size difference @liuyingyi-ukyushu and for the suggested fix @bjonkman.

Issue #172, opened 2018-09-11
Issues with VTK naming


bjonkman:
I have noticed a few problems with the VTK file naming merged in https://github.com/OpenFAST/openfast/pull/126:
1. The width of the field used to add zero-padding is based on `DT_out` instead of `vtk_fps`. If you take the 5MW_Land_DLL_WTurb regression test, ask to output vtk files at 15 fps, and set DT_out to 1 or 10, you will change the width of the field and thus the number of vtk files that are generated. In the case DT_out = 10, I get files numbered 0-9 that get overwritten every 10 steps. Using something like `tWidth = log10( (p_FAST%n_TMax_m1+1) / p_FAST%n_VTKTime ) ) + 1` with appropriate type casting would work better.

2. `get_vtkroot_path()` and `get_vtkdir_path()` assume that path names must include the `PathSep` variable to parse where the new `vtk` directory should be created. This is not a valid assumption on Windows. It would be better to call `GetPath()` to determine the directory name instead of the simple logic implemented in those routines.  For example, if I run the command `openfast.exe 5MW_Land_DLL_WTurb/5MW_Land_DLL_WTurb.fst`, it creates a `.\vtk` directory, then tries to write to `.\5MW_Land_DLL_WTurb\vtk`. If that directory doesn't exist, no vtk files get written.

rafmudaf:
Hi Bonnie,

To your first point, good catch. I've updated that locally and will issue a pull request soon.

On your second point, the vtk directory path is based on the output file root, `p_FAST%OutFileRoot`, so in the use case you mentioned the vtk directory should be located in the same directory as `5MW_Land_DLL_WTurb.out`. On Windows, is the `.out` file created in the directory containing the `.fst` file?

Thanks for the feedback and help!

Rafael

bjonkman:
The `.out` file and the `.fst` file are in the same location on Windows, just like other operating systems.
I am just saying the way the `last_separator_index` is currently calculated is not sufficient for Windows systems. 

Windows doesn't care if you use `\` or `/` in a path name, so `p_FAST%OutFileRoot` may contain _either or both_ of those characters to separate directory names. However, the current OpenFAST logic assumes it will only contain `\` when it looks for the last path separator.

Also, if you _do_ pass a path name with a '/' character on Windows, the `mkdir` command will need the path name in quotes. Otherwise, it will give you an error about syntax (though you might not see it because there is no error checking in the OpenFAST MKDIR routine).

[...]

PR #257, opened 2019-03-07
Zero padding of VTK names, and log10 of zero (see #250)

### LINKED ISSUE #250: Initilization error and VTK outputs prevents a clean abort
**Bug description**

If an error occurs during the fast library initialization, the routine `ExitThisProgram` is called, which then attempts to write the mesh for debugging if `WrVTK>0`. The issue is that  `y_FAST%NOutSteps` might be zero, and then the program will crash on the following statement `Twidth = int(log10(real(y_FAST%NOutSteps))) + 1` . 

**To Reproduce**

Try to trigger an error at initialization (e.g. by reproducing #249), and set `WrVTK` to on. 

**Expected behavior**

The program should abort nicely. 

**Suggested solution**

The calls to `log10` in `FAST_Subs.f90` should be protected as follows: 
```fortran
   if (y_FAST%NOutSteps==0) then
      CALL SetErrStat( ErrID_Warn,'NOutStep is zero, setting Twidth to 1',ErrStat2,ErrMsg2,RoutineName)
      Twidth = 1
   else
       Twidth = int(log10(real(y_FAST%NOutSteps))) + 1
   endif
```
I'll send a pull request later that fixes this if people agree.

ebranlard:

Fixes #250

`openfast-library`, file `FAST_Subs`

### DIFF
--- a/modules-local/openfast-library/src/FAST_Subs.f90
+++ b/modules-local/openfast-library/src/FAST_Subs.f90
@@ -5006,12 +5006,12 @@ SUBROUTINE WrVTK_AllMeshes(p_FAST, y_FAST, MeshMapData, ED, BD, AD14, AD, IfW, O
    CHARACTER(ErrMsgLen)                    :: ErrMSg2
    CHARACTER(*), PARAMETER                 :: RoutineName = 'WrVTK_AllMeshes'
 
-   ! calculate the number of digits in 'y_FAST%NOutSteps' (Maximum number of output steps to be written)
-   ! this will be used to pad the write-out step in the VTK filename with zeros in calls to MeshWrVTK()
-   if (y_FAST%NOutSteps==0) then
-      Twidth = 1
+   ! Calculate the number of digits for the maximum number of output steps to be written.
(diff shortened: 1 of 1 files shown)


ebranlard:
This revised pull request should now address the two issues mentioned in #250. The second one concerns the variable `twidth` which was wrongly set, as mentioned by @bjonkman in #172. 

I believe it's ready to merge, it implements the solution discussed in #250.

AnitaLeite:
Hello! I'm trying the visualization tool for the first time on OpenFast v3.1 and the following error shows up: 'FAST_InitializeAll:HydroDyn_Init: WaveElevSeries arrays for first and second order wave elevations are of different sizes.'
I'm running for irregular waves with AeroDyn and Inflow disabled and using Volturn-US semi-sub model. 

I was checking this conversation for a solution and this was flagged in 2019. Is there a solution for this please? I thought that if the problem was on FAST_subs.f90 that the binaries for my version were already updated?

Thank you!

Anita

jjonkman:
Dear @AnitaLeite,

It looks like this error was reported recently--see: https://github.com/OpenFAST/openfast/issues/1484.  I'm not aware if this has been debugged yet.

Best regards,

PR #263, opened 2019-03-20
Sync the git versions for windows and linux builds

rafmudaf:
**Complete this sentence**

Makes the version numbers consistent between cmake and visual studio builds. I found this improvement in @bjonkman's fork at this commit https://github.com/bjonkman/openfast/commit/797c692e4b169843e5da1bffed11e895adfae6a8. This is currently a bug since the `dirty` version will not currently be detected.
@bjonkman do approve until your `vs-simulink` branch is merged?

None

Version number for all modules

None

None


bjonkman:
For clarification, the version number displayed in the VS code does already display the `dirty` version (see line 5). 

This improvement just makes sure that the exact git command that was used to display the version in the code is later checked for an error. This error handling makes sure that it would still build in VS if someone downloaded the code as a ZIP file and not as a git clone. Previously it was just checking for an error on `git describe`, but I noticed that sometimes `get describe` returned an error code when `get describe --abbrev^=8 --always --tags --dirty` did not.

rafmudaf:
ah, now I see it. Thanks for clarifying that!

PR #266, opened 2019-03-27
Avoid recompiling everything when git status changes

rafmudaf:
**Complete this sentence**

This pull request changes the versioning system as it was initially implemented in CMake. This does **not** affect the build for Windows with Visual Studio.

The existing implementation inserts the git hash (AKA the version) into the source code through a compiler flag and preprocessor directive. Therefore, anytime the git status changes the compile command changes so everything has to be recompiled. This pull request changes the method for monitoring the git status and passes the version through an included header file similar to the method in the vs-build Visual Studio solution.

Also, I've removed the intel branch of the TravisCI tests until we get a proper license.

None

The build system through CMake

This is important when dealing with multiple branches as the git status changes frequently. Also, it will enable improved performance in the automated testing.

Only the existing TravisCI tests are necessary.


ebranlard:
I like the idea that the version is set via a `configure_file` command of `cmake`!  Just to be sure, my understanding is that the workflow is something like this: 
- everything is compiled with the preprocessor definition `-DGIT_INCLUDE_FILE=src\git.h`. So when the file `NWTC_IO` is compiled, the preprocessor includes the file `src\git.h` and then the variable `QueryGitVersion` ends up being set.
- nwtclibs depends on `check_git_repository`
- `check_git_repository` looks if the repository hash has changed or if it's dirty, and if so a new `git.h` file is written based on the template `git.h.in`. 

If that's correct, a change of `git.h` should trigger a recompilation of `NWTC_IO`. Is that something handled by cmake? I'm not sure I see how this dependency would be triggered, but I might have missed something in the cmake files.

rafmudaf:
Thanks for asking these questions @ebranlard. After doing more testing, I'm closing this pull request without merging while I work on an improved solution.

Issue #267, opened 2019-04-01
Suggestion for changing build dependencies


bjonkman:
**Is your feature request related to a problem? Please describe.**
Every time the git hash changes, the NWTC Library is (or at least should be) rebuilt. Since almost all code depends on this library, this results in a lot of unnecessary rebuilding/linking.

**Describe the solution you'd like**
I suggest moving the the `QueryGitVersion` subroutine in the NWTC Library to a separate stand-alone module that only the glue codes and drivers depend on.

ebranlard:
Yes! That's the way forward in my opinion. I've been discussing this with @rafmudaf  and I believe he is working on a solution.

ebranlard:
I believe `VersionInfo.f90` is not recompiled in Visual studio if the git version changes. I don't know how to force a dependency between this file and the git header file. 
An alternative, is to force a timestamp change of VersionInfo.f90. https://stackoverflow.com/a/35498399/627911

@bjonkman do you confirm this is indeed an issue?

bjonkman:
@ebranlard I've played around with a few things in the OpenFAST Visual Studio files, but it looks like Visual Studio does not notice that changing the header file should rebuild the VersionInfo.f90 file. When I generate the dependency list for that file, it does show there is a dependency on the header file (though not on the file specified in the intermediate directory), so I think that would be something to ask Intel.

You can change the timestamp using a pre-Build step like that example you've linked to. HOWEVER, it still won't build that file on that particular build step because it seems to look at the timestamp _before_ pressing build. It will build it the _next_ time you build. If you delete the versioninfo.obj and versioninfo.mod files instead of changing the VersionInfo.f90 time stamp, it would force you to build the second time. **I'd probably change the timestamp or delete the intermediate files on the post-build step instead**. Then you would force it to recompile VersionInfo.f90 every time without having to hit build twice.

bjonkman:
Note: After #428, the VersionInfo module is now used in NWTC_IO.f90, which reverts the dependency changes discussed here.

PR #269, opened 2019-04-08
Add a “version” module to remove git dependency

rafmudaf:
**Complete this sentence**

This pull request changes the versioning system as it was initially implemented in CMake. This does not affect the build for Windows with Visual Studio.

The existing implementation inserts the git hash (AKA the version) into the source code through a compiler flag and preprocessor directive in NWTC_IO. Therefore, anytime the git status changes the compile command also changes so everything has to be recompiled.

This pull request improves the build system
1. Moving the git status checking to a standalone module
2. Moving the compiler flag from the main CMakeLists to the new module CMakeLists

These changes mean that the compile command for the other modules does not change with the git commit and only the lightweight Version module needs to be recompiled. The driver programs which use this module are relinked during compile time.

https://github.com/OpenFAST/openfast/issues/267

The build system through CMake

This is important when dealing with multiple branches as the git status changes frequently. Also, it will enable improved performance in the automated testing.

Only the existing TravisCI tests are necessary.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,12 +22,6 @@ include(${CMAKE_SOURCE_DIR}/cmake/OpenfastFortranOptions.cmake)
 
 set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
 
-# Get the git info into the executable
-list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/")
-include(GetGitRevisionDescription)
-git_describe(GIT_DESCRIBE)
-add_definitions(-DGIT_VERSION_INFO="${GIT_DESCRIBE}")
-
 # CMake Configuration variables
 if (NOT CMAKE_BUILD_TYPE)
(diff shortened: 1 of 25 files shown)


rafmudaf:
@bjonkman thank you for creating issue #267. I think this pull request addresses it as you've described. Could you please take a look?

bjonkman:
I just briefly glanced through the changed files. Are you sure the Windows Visual Studio solution files will actually successfully build with these changes? I would be very surprised if they work without adding the new `VersionInfo.f90` file to their list of source files.

bjonkman:
@rafmudaf I just made a pull request to your branch to address the changes that need to happen in the Visual Studio builds.

rafmudaf:
Thank you very much @bjonkman. After that bug fix, @ebranlard do you have any comments?

rafmudaf:
By the way, the failed TravisCI check is due to the expired intel license. In another pull request, I'll remove that test from the Travis setup until we get a better solution.

ebranlard:
Everything looks good to me!

Issue #283, opened 2019-04-29
Option SumPrint not respected for primary input file


ebranlard:
**Bug description**

The option `SumPrint` from the primary input file is not respected and the summary file is written irrespectively of this input. 

I'll submit a pull request to address this.

rafmudaf:
Fixed by #284

PR #284, opened 2019-04-29
Only printing summary file when requested

ebranlard:

This small fix should correct #283


rafmudaf:
Good catch @ebranlard. I've compiled this code and verified it works as advertised.

PR #330, opened 2019-09-16
Feature/interactive regression test plots

RHammond2:

 This pull requests modifies the regression testing suite to replace the visualizations with interactive visualizations, better accentuate which cases have failed in the export summary, and to clean up a few spots in the code to make it more pythonic. This work was done under the guidance of @rafmudaf.

This only impacts the regression testing suite.


rafmudaf:
@RHammond2 The Bokeh JavaScript files don't load for me. This is what I see in the JS console:

```
[Error] Failed to load resource: the server responded with a status of 403 () (bokeh.min.js.map, line 0)
[Error] Failed to load resource: the server responded with a status of 403 () (bokeh-widgets.min.js.map, line 0)
```

Any idea why?

RHammond2:
@rafmudaf I can't seem to find anything related to this issue. Are you blocking access within your browser; though, I'm not convinced that would give that error.

rafmudaf:
I thought we discovered that the error above was only a problem when generating plots for cases that hadn't actually failed, but I ran this again after compiling in single precision so that most test cases fail and still same issue. So this is still pending.

On another note, since each plot is opened by clicking on a link (the id selector `#`), it can take a lot of clicks on the browser back button to get back to the main page. Adding a `back to home` button may be worthwhile.

RHammond2:
So this issue isn't actually blocking any of the functionality. There was a bit of confusion as to what was expected for output because we were looking at a case where none of the tests failed causing there to be no plots that are output. Even when plots are output the "error" comes up.

Where would the `back to home` button lead to for this if we are only looking at a single case? Would the idea be to access any of the cases as opposed to going through each directory separately?

Issue #303, opened 2019-07-10
Is the openfast-registry dependency necessary?


rafmudaf:
**Is your feature request related to a problem? Please describe.**
The OpenFAST Registry is currently used to generate fortran code for the module types files. It is the first executable compiled any time OpenFAST is compiled as every module needs to have its types module generated before it can compile. The dependency on source code that isn't included in the repository has been causing headaches in CMake and Visual Studio.

*Is there still a good reason to generate the module types files on every compile instead of including them in git directly?*

I'm *not* proposing to get rid of the registry, but rather remove it from the OpenFAST build and remove the autogeneration of types files for every build. Whenever a change is made to the module registries, the `.txt` files would still be modified and run through the registry executable but the output of that would be committed to the repo.

**Describe the solution you'd like**
- Remove the registry as a dependency of OpenFAST
- Commit the module types files
- Include documentation for when to use the registry, how to compile it, and how to use it; specifically document the process for updating the module types
  - update the Module_Registry.txt
  - run registry
  - commit x,y,z

**Describe alternatives you've considered**
- Adding type-file targets which the modules depend on - down side is this almost doubles the number of targets so theres a lot of noise in `make help` or the Visual Studio project tree
- Adding a single registry target which contains all types - down side is that then every module contains all of the types in OpenFAST. This increases build time for small targets like `beamdyn_driver`.

**Additional context**
This would also enable optimizing the module types files for performance without having to modify the registry source code.

**Seeking input from**
@bjonkman 
@michaelasprague 
@sayerhs

rafmudaf:
also @andrew-platt

andrew-platt:
In almost every code development effort I am involved in within OpenFAST, modifications of registry files are required and frequent.  So for my workflow, it is much better to keep the registry as a dependency rather than impose another process that I would need to keep track of and manually trigger.

In my experience, generating the types files is not a large bottleneck, but compilation step is very time consuming.

ebranlard:
I think it makes sense to have them auto-generated, but I personally prefer having those backed up in git.  Unless they really are build dependent (OS, architecture, compiler), which I don't think the types should be.

andrew-platt:
I don't think there is anything build dependent in the types files.

[...]

PR #337, opened 2019-09-30
Remove the default dependency on the OpenFAST Registry

rafmudaf:
**Complete this sentence**

This pull request removes the default dependency of the OpenFAST Registry on any module with registry-generated types files. Instead, the OpenFAST Registry is enabled via a CMake flag, `GENERATE_TYPES`. When this flag is enabled, the `openfast-registry` target is added as a dependency to any other relevant targets and an additional command is added to the build step to execute the registry. Though the OpenFAST registry will be executed for any modules that use it, only the types-files that have changed are recompiled and show different in version control.

The types files for each module as they are in the `openfast/dev` branch are also added into version control in this pull request.

https://github.com/OpenFAST/openfast/issues/303

CMake configuration and documentation

None

Not applicable to tests. A live version of the documentation is available at https://raf-openfast.readthedocs.io/en/feature-registry_rework/source/dev/types_files.html


rafmudaf:
As an added bonus, this fixes a problem in building with Visual Studio where the autogenerated files aren't recognized by the build system in the same step as when they are generated.

rafmudaf:
@ebranlard @andrew-platt I fixed a problem you may have noticed where the C++ API wasn't building correctly. And I turned on the `GENERATE_TYPES` flag in one of the Travis CI cases.

modules/aerodyn14/src/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
+#
```
andrew-platt: I don't think we need this file.

rafmudaf: you're right, it's gone!

bjonkman:
Could you also modify the vs-build project files to point to the new locations for the _Types.f90 files? They are still pointing to the `.\build\types-files\` directory.

rafmudaf:
@bjonkman is that the only change that needs to happen for the vs-buid project? I see these lines in the projects for the drivers:
```
Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "FAST_Registry", "..\Registry\FAST_Registry.vcxproj", "{DA16A3A6-3297-4628-9E46-C6FA0E3C4D16}"
EndProject
```
Do those also need to be removed?

bjonkman:
@rafmudaf, that line in the drivers allows the driver codes to build types files if the registry files change, so I wouldn't remove it from the solution files. 

This is what I think needs to be changed:
- I don't think I've ever used the `vs-build\FAST-Types` project file. Unless someone else has a need for it, I would remove everything in that directory.
- The `RunRegistry.bat` file should have the `Output_Loc` environment variable changed for each module that generates files. You might be able to just put `SET Output_Loc=%CURR_LOC%` after each module sets `CURR_LOC`.
- I would remove all of these sections that I noticed in the .vfproj files: `AdditionalIncludeDirectories="..\..\build\types-files;..\..\OpenFAST\modules\nwtc-library\src"`. The source directories are included by default, so there's no need to specify them. (And if people don't name their directory `OpenFAST`, it probably won't work anyway.)
- You'll have to replace all of the file names that contain `\build\types-files\` with the new appropriate locations.

rafmudaf:
@andrew-platt I've updated the documentation for this to add some guidance on working with modified or new registry input files: https://raf-openfast.readthedocs.io/en/feature-registry_rework/source/dev/types_files.html.

What do you think?

andrew-platt:
@rafmudaf The documentation is pretty clear in my opinion.  That should help any developer wanting to modify or update modules.  I'm guessing we will have some questions that come up when this PR is merged.

rafmudaf:
@bjonkman I've made the changes to the vs-build project, but unfortunately this did *not* solve the problem where Visual Studio doesnt recognize the new types files. I guess the way to do that is to have a switch or otherwise turn off running running the registry in Visual Studio.

Also, this was a good find: [asdf](https://github.com/OpenFAST/openfast/pull/337/commits/0394e2cd8fcd8aa4d1c39a7dbaa07eb7ac4da87c?file-filters%5B%5D=.bat&file-filters%5B%5D=.vcxproj&file-filters%5B%5D=.vfproj&hide-deleted-files=true#diff-eb04d9e1395a40d07e0199a3920e229aL111)

rafmudaf:
@bjonkman could you please double check what I've done?

rafmudaf:
The CMake-generated Visual Studio solution will get around the problem where VS doesnt recognize the new types files because they wont be new. I mentioned this above because I was hoping that the standalone project would also be fixed since the files are actually there.

bjonkman:
@rafmudaf I fixed a few things in the vs-build and made a pull request to your branch. 

Unless this was caused by the IceFloe issue I just fixed, I'm not sure what you mean by 
>  this did not solve the problem where Visual Studio doesnt recognize the new types files

As far as I can tell, this works as I expect it to. This change means that on a new checkout of the code, you should not have to hit build, close the solution when it finishes, then reopen the solution file and build again. I didn't check, but if you still have that guidance in the documentation, you can remove it.

However, Visual Studio does still use the time-file-was-last-modified at the time when you hit "build" to determine if a *_Types.f90 file should be compiled, so if it calls the Registry to regenerate a file, it will need to be built a second time to take those changes into account (it doesn't seem to realize that the timestamp on the Types files change during the build process). That said, if you "build" instead of "rebuild" or "clean" a solution, the Registry will almost never run. "rebuild" and "clean" will delete all of the registry-generated files.

Interesting find on "asfasdf". 👍

rafmudaf:
Thanks Bonnie. I think what I was seeing was indeed due to the bug you fixed for icefloe. I've also removed the bit about closing and reopening the Solution in the docs.

PR #356, opened 2019-11-07
Using Findloc in SubDyn

ebranlard:

An important part of the discretization routine `SD_Discrt` is used to reindex the user inputs from "IDs" to indices. This involved several lookup of indexes in arrays. This was previously done by copying pasting several time the same code ("while loops"). 

I've added a function `findloci`, which is a simplified version of the fortran intrisinc `findloc` (which is not yet available on all compilers), and used this instead of the copy-pasted while loops. 

I've also isolated the code that performs the reindexing since it doesn't really belong to the discretization routine. 

I'm committing this now to limit the differencess with my other branch `f/flexsub2` 

Follow up of  #351 

SubDyn module


andrew-platt:
There is a subroutine interface `LocateStp` that does essentially the same thing as the `FindLoci`.

However, given that `FindLoci` is a function, I say let's keep it rather than try converting to the `LocateStp` routine.  I think this makes the code more readable.

ebranlard:
Thanks for the feedback, I'll try to see if I can use `LocateStp` in the future

ebranlard:
I'd like to use `findloci` to fix #378. I believe this PR it can be safely merged.

Issue #319, opened 2019-08-02
Feature Request - More than 9 output nodes for AeroDyn and BeamDyn


JelmerPolman:
**Is your feature request related to a problem? Please describe.**

Modelling a large blade with BeamDyn leads to large computational times. Especially when the cross-sectional matrices are fully populated, a small time step has to be chosen for the coupling with AeroDyn, and the order of elements is relative high. In my case, a Time/CPU-ratio in the order of 10^(-3) is no exception, leading to CPU Times in the order of days. 

For large blades, 9 output nodes is not enough to fully capture the detailed deformation or load distribution of the blade. It is a shame that after all that computational work, only the results at 9 nodes can be written as an output. Hence, only a small part of the computational work is than available for the user. If one needs more nodes, you will have to redo the whole computation only then with different output nodes.

**Describe the solution you'd like.**

Introduce an option where users can define an unlimited number of output nodes.

**Describe alternatives you've considered.**

Related to this feature request. The elements in BeamDyn are spanned by the basis functions. The solution BeamDyn finds is thus continuous. It would be great if the user can also define output nodes in BeamDyn that are not equal to the FE nodes or the quadrature nodes. Deformations at nodes other than the FE or quadrature nodes should be easily computed since this is just an expansion of the basis functions, weighted with the deformations at the FE nodes. This expansion as a post processing step has become more difficult for the user to do since the deformations at the FE nodes are no longer an output as from OpenFAST v2.0.0. 

By doing so, the user can specify a sensor location equal to an actual location for a real blade test, or a location equal to other codes for a code-to-code comparison, and is unlimited by a post processing step that requires interpolation.

bjonkman:
This issue was addressed in #373 with the new nodal output features in AeroDyn, ElastoDyn, and BeamDyn.

I suggest closing this issue and creating a new issue for the related request listed in the `Describe alternatives you've considered`, if that feature is still desired.

Issue #304, opened 2019-07-12
AeroDyn standalone driver output options


whophil:
Hi all, two questions related to the AeroDyn standalone driver output.

1. Is it possible in the OpenFAST version of the AeroDyn driver to have more than 9 blade station outputs?
2. Is it possible to concatenate the output in "combined-case analysis" into a single file instead of bursting it into one file per case?

These may sound familiar to @andrew-platt and @bjonkman.

If these features are currently not available, I'm happy to contribute but I thought I'd check first.

andrew-platt:
Hi Phil,

At present neither of these features that we are used to are in the OpenFAST version of the AeroDyn driver.

1.  @bjonkman  and I have discussed bringing the nodal outputs into OpenFAST for BeamDyn. We may be able to do that with AeroDyn as well.
2. At present the AD15 driver only handles single cases at a time.  Perhaps this is code Envision would be willing to share with the community?

Do you have a timeline for when you would like to see this happen?

whophil:
Thanks @andrew-platt. I don't have any particular timeline in mind.

Feature 1 (nodal outputs) would higher priority for me. Feature 2 would be nice to have but isn't critical for my use case.

Somewhat related - is the interface for `libaerodyn` exposed or documented anywhere?

whophil:
Hey @andrew-platt and @bjonkman , any word on if nodal outputs will be coming to NREL's AD driver?

bjonkman:
Hey @whophil , yes, I think nodal outputs will be coming .... sometime. I'm planning to give NREL some code related to our joint linearization project in a month or so, and that will probably be part of the code dump.

whophil:
Great news @bjonkman and @andrew-platt ! Let me know if there's anything I can do to help.

bjonkman:
Item 1 in this issue was addressed and merged into dev with #373. 
Item 2 is part of a not-yet-ready-to-merge branch here: https://github.com/bjonkman/openfast/tree/f/LinearAero. I expect to make a pull request with that code base sometime this fall.

Issue #99, opened 2018-03-12
Replace TRANSFER Function in FAST Registry-Generated Pack/Unpack Routines


jjonkman:
The TRANSFER function puts data on the stack, so pack and unpack routines generated by the FAST Registry can cause stack overflow errors when run on large data sets (e.g., large wind files or large offshore models).

bjonkman:
This was completed in #373, which is now merged into 'dev', so I think this issue can be closed.

PR #350, opened 2019-10-24
Add offshore linearization capability

ghaymanNREL:
Complete this sentence
**THIS PULL REQUEST IS READY TO MERGE**

**Feature or improvement description**
This pull request adds new linearization functionality to OpenFAST for floating offshore wind turbines.  Specifically, linearization capability has been added for HydroDyn (and associated sub-modules) and MAP++.  A new state-space wave excitation module has been developed.  Additionally the glue-code has been modified to support these additional linearization features.  The HydroDyn input file now includes options (ExctnMod) related to the new state-space wave excitation module.

**Related issue, if one exists**
None

**Impacted areas of the software**
OpenFAST glue-code
ElastoDyn
HydroDyn
WAMIT
SS_Excitation
SS_Radiation
MAP++

**Additional supporting information**
https://www.nrel.gov/docs/fy19osti/71865.pdf
https://www.sintef.no/globalassets/project/eera-deepwind-2019/presentations/h_jonkman_nrel.pdf

**Automated test results**

PR #373, opened 2019-12-18
New Linearization and Mode-Shape Capabilities (and more)

bjonkman:
**Complete this sentence**

**Summary**
This code is the result of the Envision-NREL 2019 collaboration on finding a trim solution for linearization and visualization of mode shapes.  

_As warned, this is a BIG pull request. In an ideal world, this would be broken up into several different pull requests. However, that was not possible given the current states of our respective code bases and time considerations._

I have not provided much in terms of updated documentation in the pull request because the original documentation is not in a place I can easily modify it. I can help update the documentation after the original documentation has been merged in.

- **Trim Solution** 
  + I added new capability to find a periodic steady-state trim solution where linearization will be performed, based on this plan from NREL: https://github.com/ebranlard/temp-lin. OpenFAST will calculate the times to perform linearization instead of requiring the user to specify them.
- **Mode-Shape Visualization**
  + I added capability to visualize mode shapes using VTK output files, based on an algorithm from @jjonkman 
  + Generating these figures is a multi-step process that requires restarting OpenFAST after performing MBC3 analysis. I added the `-VTKLIN` command-line switch to indicate that OpenFAST is being restarted with the visualization option.
  + I created a Matlab script to describe the process here: https://github.com/bjonkman/r-test/blob/pullrequest/bjonkman-linear/glue-codes/openfast/5MW_Land_ModeShapes/AnalyzeModeShapes.m
  + You can also view some documentation about viewing mode shapes here: https://github.com/OpenFAST/r-test/blob/pullrequest/bjonkman-linear/glue-codes/openfast/5MW_Land_ModeShapes/vtk-visualization.md
- **Nodal Outputs on Blades**
  + We added (optional) additional sections at the end of ElastoDyn, BeamDyn, and AeroDyn input files to request channel outputs at every node on a blade. 
  + The structural codes use naming conventions `B#N###channel`, and AeroDyn's nodal output naming convention is `AB#N###channel`. This allows the user to distinguish between aero and structural channels that may have the same name. 
  + The available channel names are found in https://github.com/bjonkman/openfast/blob/f/Linear/docs/OtherSupporting/OutListParameters.xlsx
- **Longer channel names**
  + I set the max channel length to allow up to 20 characters in a channel name in OpenFAST. This number can be easily changed by modifying the value of  `ChanLen` in `NWTC_Base.f90` and rebuilding. 
  + The compressed binary file format will now contain the length of the channel names.
  + FAST.Farm should be checked for compatibility because I removed the now-unnecessary `ChanLenFF` variable and corresponding type

- **Registry Improvements**
  + The Registry now allows for interpolation/extrapolation of periodic signals. Angles that have 2pi periodicity are treated as a special case to avoid issues where interpolating between 0 and 360 degrees would give an answer 180 degrees off.
  + The pack and unpack routines have been updated to use less stack space. This was necessary since the mode-shape visualization uses these routines.
  
- **Performance improvements**
  - `WriteOutput` variables in AeroDyn and BeamDyn are not computed if they will not be used at that step. 

- **Bug fixes**
  - The AeroDyn driver `OutFileRoot` is now specified relative to the location of the driver input file instead of the current working directory.
  - Fixed a problem where AeroDyn could potentially think it was using a different DT than what it reported to the glue code.

- https://github.com/OpenFAST/openfast/issues/319
- https://github.com/OpenFAST/openfast/issues/304
- https://github.com/OpenFAST/openfast/issues/99

- ElastoDyn
  - New nodal outputs on blades
  - Linearization
- AeroDyn
  - New nodal outputs on blades
  - Linearization
  - New method of calculating the `D` matrix in linearization (will not produce error about being non-differentiable between AeroDyn solution regions). **Warning: There may be an issue with FrozenWake with this modification. It is not apparent in the cases I ran, but the logic seems a little off to me. This has been fixed in a future pull request for linearized aerodynamics from Envision.**
  - Performance improvements: do not calculate `WriteOutputs` unless they are required
- BeamDyn
  - New nodal outputs
  - Linearization
  - Performance improvements: do not calculate `WriteOutputs` unless they are required
- ServoDyn
  - Linearization
  - Bladed DLL interface code: partial implementation of API from Bladed 4.9
  - Can be built with preprocessor definition `LOAD_DLL_TWICE_FOR_LOGGING_CHANNELS` to output logging channels from the DLL (including legacy version) in the OpenFAST output file. This, however, requires the DLL to be unloaded and loaded again.
  - Removed `ElecPwr_prev` and `GenTrq_prev` from ServoDyn's inputs. These are now stored in misc vars to send back to the Bladed DLL (and simplifies the glue code)
  - The DLL controller is now called in UpdateStates (and when it is first called in CalcOutput, if it hasn't already been initialized)
  - Updates to how the HSS brake is applied from the Bladed-stlye DLL (uses linear ramp)
- Glue-code
  - Linearization
  - Changes to input file
  - Changes to linearization output file format
- Registry
  - Pack/Unpack routines
  - Extrap/Interp routines
- FAST Library (Simulink interface)
  + I increased the number of outputs that are allowed through the interface from 1000 to 4000 (the new nodal outputs gives the ability to generate a lot of data!)
- NWTC Library
  + Modifications to the binary file format
  + Increased tolerance on "close" node in mesh mapping to 7.5mm
  + Routines to read real variables now check that they are finite values that are not NaN.
  + Removed OutParmFFType type (please verify that this is not a problem with FAST.Farm)
- Documentation
  + Updated example input files
  + Fixed some text about where the static/dynamic switch is located in BeamDyn
    
- I added an example of the trim solution and mode-shape visualization in https://github.com/bjonkman/r-test/blob/pullrequest/bjonkman-linear/glue-codes/openfast/5MW_Land_ModeShapes/. This is not very robust, but is a good starting point for NREL (or someone else) to automate the mode-shape visualization with different options.
 - The matlab-toolbox has been updated in https://github.com/OpenFAST/matlab-toolbox/pull/11 to work with this code base

- Input files for the regression tests have been updated here: https://github.com/bjonkman/r-test/tree/pullrequest/bjonkman-linear
- I would not expect results to change except for 
  + minor differences in linearization output in models that have enabled AeroDyn
  + cases that may have different Bladed DLL controller options enabled (e.g. HSS brake), though I don't think any of the NREL regression tests test that feature.


bjonkman:
For fun, here are the first two mode-shape plots that I generated with the scripts in the `5MW_Land_ModeShapes` regression test directory:

Mode 1 (tower side-side)
5MW_Land_ModeShapes Mode1 LinTime1

Mode 2 (tower fore-aft)
5MW_Land_ModeShapes Mode2 LinTime1

ebranlard:
I was able to test this branch and successfully generate the `avi` files for the test case `5MW_Land_ModeShapes/` 

When appropirate, we need to remember to document the API changes (i.e. https://openfast.readthedocs.io/en/master/source/user/api_change.html )

AeroDyn input file line 37 (after ---- Airfoil Information ----) 
```
          1   AFTabMod           - Interpolation method for multiple airfoil tables {1=1D interpolation on AoA (first table only); 2=2D interpolation on AoA and Re; 3=2D interpolation on AoA and UserProp} (-)
```
Main fst file, line 44 (after `Linearize`)  
```
False          CalcSteady      - Calculate a steady-state periodic operating point before linearization? [unused if Linearize=False] (flag)
          3   TrimCase        - Controller parameter to be trimmed {1:yaw; 2:torque; 3:pitch} [used only if CalcSteady=True] (-)
     0.0001   TrimTol         - Tolerance for the rotational speed convergence [used only if CalcSteady=True] (-)
      0.001   TrimGain        - Proportional gain for the rotational speed error (>0) [used only if CalcSteady=True] (rad/(rad/s) for yaw or pitch; Nm/(rad/s) for torque)
          0   Twr_Kdmp        - Damping factor for the tower [used only if CalcSteady=True] (N/(m/s))
          0   Bld_Kdmp        - Damping factor for the blades [used only if CalcSteady=True] (N/(m/s))
```

bjonkman:
Okay, I updated the `api_change.rst` file. I've been trying to keep the sample input files up to date (though most of them aren't in the docs, yet), but I guess I missed when you guys added that API summary section to the docs. I also updated some previous API changes that weren't documented there, yet, either.

jonmtz:
Hi,

I am trying to compile OpenFAST from branch f/Linear in order to try the linearization tool. Nevertheless, y found some problems when compiling it: 

(I am using Cygwin 64 bit + CMake since even if I do have Visual Studio, I do not have license for Intel Fortran Compiler and therefore I think that I am not able to compile that way).

1.	I got an error saying that the file src/BeamDyn_BldNdOuts_IO was not found in modules/beamdyn/src/CMakeLists.txt

I was able to solve it by changing the name of the call to: src/BeamDyn_BldNdOuts_IO.f90 (I think this is a bug)

> set(BD_SOURCES
  src/BeamDyn.f90
  src/BeamDyn_IO.f90
  src/BeamDyn_BldNdOuts_IO.f90
  src/BeamDyn_Subs.f90
  src/BeamDyn_Types.f90
  )

2.	After doing that, I have tried to compile again, and now I get the following message:

> [ 60%] Building Fortran object modules/hydrodyn/CMakeFiles/hydrodyn_driver.dir/src/HydroDyn_DriverCode.f90.o
/cygdrive/d/Users/jmrico/Documents/9131_SEAPOWER/openfastLinearBJ/modules/hydrodyn/src/HydroDyn_DriverCode.f90:543:105:
call RunTimes( StrtTime, REAL(UsrTime1,ReKi), SimStrtTime, REAL(UsrTime2,ReKi), time, UsrTimeDiff )
1
Error: Type mismatch in argument ‘unsum’ at (1); passed REAL(8) to INTEGER(4)
make[2]: *** [modules/hydrodyn/CMakeFiles/hydrodyn_driver.dir/build.make:63: modules/hydrodyn/CMakeFiles/hydrodyn_driver.dir/src/HydroDyn_DriverCode.f90.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:862: modules/hydrodyn/CMakeFiles/hydrodyn_driver.dir/all] Error 2
make: *** [Makefile:130: all] Error 2


I am not sure what this ‘unsum’ makes reference to, but I was not able to find it.

I would really appreciate if someone could help me with this problem.

Thanks in advance,

Jon

bjonkman:
@jonmtz, Thanks for reporting those issues. I've pushed some changes that should hopefully fix both problems. Let me know if you find anything else.

jonmtz:
Dear @bjonkman , thank you for the quick answer. With the new version I compiled without problems, so seems to be OK.

If I find any problems when performing simulations I will let you know.

Best regards,

Jon

modules/hydrodyn/src/HydroDyn_Input.f90
```diff
@@ -2398,7 +2408,23 @@ SUBROUTINE HydroDynInput_ProcessInitData( InitInp, ErrStat, ErrMsg )
       END IF
    END IF
 
-
+      ! Linearization Checks
+   ! LIN-TODO:
```
ebranlard, line 2412: Are these checks not needed anymore?

ebranlard: ok, it seems to be done below

modules/nwtc-library/src/NWTC_Num.f90
```diff
@@ -55,6 +55,13 @@ MODULE NWTC_Num
    REAL(ReKi)                                :: TwoByPi                       !< 2/Pi
    REAL(ReKi)                                :: TwoPi                         !< 2*Pi
 
+   REAL(SiKi)                                :: Pi_R4                         !< Ratio of a circle's circumference to its diameter in 4-byte precision
```
ebranlard, line 58: Justa  suggestion, I believe pi can be defined as a parameter like this:
```
real(SiKi), parameter :: Pi_R4 = acos(-1.0_SiKi) 
```

bjonkman: Yes, except that unfortunately doesn't work for all compilers.

modules/hydrodyn/src/HydroDyn_Input.f90
bjonkman: These are part of the changes I started with from #350.

andrew-platt:
Items to check on this PR:

- [ ] remove `DBGOUTS` from AD15 -- these are superseded by the new outputs
- [ ] add documentation on the node based outputs (in user guides for each module)
- [ ] check BD changes against the bugfix branch (in progress)
- [ ] add documentation on how to use the linearization and mode shape plotting

andrew-platt:
Notes on regression tests for PR 373 (after updates are merged from my branch to @bjonkman branch):

- 5MW_OC3Mnpl_DLL_WTurb_WavesIrr: usual numerical noise
- 5MW_OC3Trpd_DLL_WSt_WavesReg: usual numerical noise
- 5MW_OC4Jckt_DLL_WTurb_WavesIrr_MGrowth: usual numerical noise
- 5MW_Land_BD_DLL_WTurb: the timestep changed so comparisons to older baselines don't make sense.  I verified it is still correct.
- WP_Stationary_Linear: new test case. No prior baselines
- Ideal_Beam_Fixed_Free_Linear: not sure
- Ideal_Beam_Free_Free_Linear: not sure

- 5MW_OC4Semi_Linear: **THIS DOES NOT RUN!**  ` FAST_InitializeAll:HydroDyn_Init:WAMIT_Init:SS_Exc_Init:HydroDyn Wave direction does not match
 the wave excitation wave direction`

ebranlard:
Just a small note. I think if we want to do linearization at WS=0, InflowWind report an error, so the user has to set CompInflow=0. For VTK surface outputs, we still need AeroDyn, so in that case we need to trust AeroDyn without InflowWind. It seems to work fine.

bjonkman:
@andrew-platt, a few comments on your regression test notes:

- WP_Stationary_Linear isn't exactly new. :) It's the old "Test14" that was there already in FAST 6. Originally we ran the MBC3 analysis on it and compared the frequencies it reported.

- I found that one of the many `WaveDir` variables wasn't getting set properly in HydroDyn for the `5MW_OC4Semi_Linear` case. I'm not sure my fix is the best way to do that, so someone more familiar with all the HydroDyn data flow should maybe review that. Hopefully that fixes the problem you were seeing (the earlier version ran fine on my VS build).

- I think it would be helpful to add some Matlab scripts for the linearization cases in r-test so that when the output files don't match _exactly_, we can still manually check what frequencies and damping are actually predicted from those matrices.

ebranlard:
I think I've encountered two small corner cases, more or less related.  If somehow I run openfast from another directory: something like:

    openfast .exe ../path/to/Main.fst

Then I can run again for VTKLin

     openfast .exe -VTKLin ../path/to/Main.viz

As long as the viz file inputs have `CheckpointRoot` and `MatlabFileName` as follows:

          ../path/to/Main.ModeShapeVTK 

I'm wondering if these paths should be relative to the VIZ file instead? 
Because if I go to the folder, then

    cd ../path/to/ 
    openfast.exe -VTKLin Main.viz

It would fail with checkpoint file not found. (`FAST_RestoreForVTKModeShape_Tary:FAST_RestoreFromCheckpoint_Tary:FAST_RestoreFromCheckpoint_T:Open
BInpFile`)
So I'd have to replace the variables `CheckpointRoot` and `MatlabFileName`. 

But, if I do replace them, the checkpoint file is opened, but somehow, OpenFAST cannot find the `.outb` file. 
```
FATAL ERROR when writing binary output file: WrBinFAST:OpenBOutFile:Cannot open file
"../path/to/Main.outb"
```
Likely because the checkpoint file "remembers" that it was created while called from another directory. 

It might make sense to make the Viz inputs relative to the viz file like other input files? As for the checkpoint, that might be related to the `Rootname` of the simulation needing reinitialization. 
There are some design choices there I guess, I just wanted to point that out.

bjonkman:
@ebranlard , Yes, this is because we used the checkpoint capability. You are basically restarting the simulation with all the data it had previously stored--including path names, location of DLLs, etc. (and EVERY module has the potential to store some of these variables, so I don't think it's feasible to overwrite the RootName in just one place). I thought I had written some of these limitations of the checkpoint-restart capability in the FAST8 ReadMe, but since the NWTC website is semi-permanently down, I don't have access to that right now. But, basically the main limitation is that you should be restarting it from the same directory as the code was originally run. I think MAP++ might even need to re-read its input file (or maybe I stored that info another way, I don't remember at the moment).

Anyway, I just modified the code so that those two files can have paths specified to the location of the input file where they are specified. Just keep in mind those other checkpoint-restart limitations if you do run it somewhere else.

ebranlard:
Great thanks!

ebranlard:
Another detail, I think the `vtk` folder is not created when running with `-VTKLin`, which would make openfast crash.

bjonkman:
@ebranlard , yes, the `vtk` folder is created on the initialization step. If someone deletes it between initially running the code and restarting with `-vtklin`, then OpenFAST would likely just not write the vtk files. It does check for an error when writing files (though the errors are not reported back to OpenFAST), so I'd be surprised if it actually "crashed". Did you try this?

I think this is something that should be noted as a limitation of checkpoint-restart capability (not just vtklin). There are MANY things that could go wrong between writing a checkpoint file and restarting if the user changes/deletes files/folders or even decides to use a different version of OpenFAST.

ebranlard:
Thank you, I was running from the matlab command prompt, and don't think I saw an error message. But I'll double check. 

A final question. I have the feeling the code uses NLinTimes to generate the vtk files. Is it possible to perform visualization, for a case at 0 rpm? I typically use NLinTimes=1 in this case and only one VTK is generated then with `-VTKLin`. Thank you!!

bjonkman:
I'm not sure there will be an error message because the way the existing `WriteVTK` routine currently works, it ignores the errors and just continues the simulation.

At 0 RPM, the trim-solution plan said to overwrite NLinTimes (or NAzimStep) to be 1, and "the number of rotations is then understood as the number of time steps." You should be able to perform visualization with this setting, but you will have to set the [`VTKLinTim` input setting](https://github.com/OpenFAST/r-test/blob/pullrequest/bjonkman-linear/glue-codes/openfast/5MW_Land_ModeShapes/ElastoDyn-Modes.viz) to be 2 (separate animations for each LinTimes).

Does that answer your question?

ebranlard:
Thanks, yes! Having `VTKLinTim=2` works with 0 rpm. I'll try to add some checks for that in the matlab scripts. 

And I confirm that openfast exits "normally" even when the vtk folder is not here. Maybe we could add a small check when the  `VTKLin` flag is there? But I understand it's just one of the cornercases related to the checkpoint.

bjonkman:
@ebranlard I created [an issue ](https://github.com/OpenFAST/openfast/issues/477) to have someone fix that issue in the future. It feels like a lot of code would need to be changed, and I don't have the time to do that right now. It's mostly mindless work, but there are a lot of lines with calls to `MeshWrVTK` that need to be checked. I'd prefer to address that issue outside of this particular PR.

ebranlard:
that's totally fine by me!

nickjohnson13:
Hi Bonnie, I'm trying to view the documentation for the linearization mode shape visualization but when I click https://github.com/bjonkman/r-test/blob/pullrequest/bjonkman-linear/glue-codes/openfast/5MW_Land_ModeShapes/vtk-visualization.md I get a 404 page not found. Also, can you let me know which version of the matlab toolbox works with AnalyzeModeShapes.m?

Thank you!!!

bjonkman:
@nickjohnson13 , that r-test branch got merged in. Here's an updated link: https://github.com/OpenFAST/r-test/blob/pullrequest/bjonkman-linear/glue-codes/openfast/5MW_Land_ModeShapes/vtk-visualization.md (and I fixed it in the comments above).

The example should work with the current matlab-toolbox master branch, but it's been quite a while since I've run these. :)

bjonkman:
Thanks, guys!

PR #390, opened 2020-02-12
Fix bug with setting initial properties in SubDyn

bjonkman:
**Complete this sentence**

Fixed a bug that was introduced in #356. When `NDiv = 1`, an array holding the property data was set to an unallocated array, which caused serious issues (simulation never completes or numerical errors) when running the `5MW_OC3Trpd_DLL_WSt_WavesReg` regression test. 

SubDyn

I made this fix based on a quick scan of the logic in `SD_Discrt` and comparing the code with the previous version of the subroutine. @ebranlard should verify that this fix does what is intended.


ebranlard:
Yes, this line should have been deleted! Thanks a lot!

PR #395, opened 2020-02-20
Added flap control to DLL

andrew-platt:
This is ready to merge.

Modifications were made to the airfoil polar interpolation feature to use signals from the DLL controller.  The flap deflection angles are read from the DLL swap array and passed to AeroDyn through the `UserProp`.  These are then used in the airfoil interpolation to simulate the effects of an active flap on the airfoil section.

ServoDyn: avrswap array.

No test cases use this feature, so no tests are expected to fail.


andrew-platt:
@bjonkman, would you mind taking a quick look at this to see if there might be some unintended consequences?

modules/servodyn/src/ServoDyn.f90
```diff
@@ -939,12 +950,15 @@ SUBROUTINE SrvD_CalcOutput( t, u, p, x, xd, z, OtherState, y, m, ErrStat, ErrMsg
             CALL SetErrStat( ErrStat2, ErrMsg2, ErrStat, ErrMsg, RoutineName )
       END IF
       
+      ! Set ServoDyn output for flap deflection angle
+      y%BlFlapCom(1:p%NumBl) = m%dll_data%BlFlapCom
```
bjonkman: This will give you a seg fault if the turbine doesn't have 3 blades. Use `y%BlFlapCom(1:p%NumBl) = m%dll_data%BlFlapCom(1:p%NumBl)` instead

modules/openfast-library/src/FAST_Solver.f90
```diff
@@ -513,7 +517,12 @@ SUBROUTINE AD_InputSolve_NoIfW( p_FAST, u_AD, y_ED, BD, MeshMapData, ErrStat, Er
       
             
    END IF
-   
+        ! Set Conrol parameter (i.e. flaps)
+      DO k_bl=1,p_AD%numBlades
+          DO k_bn=1,p_AD%NumBlNds
+              u_AD%UserProp(k_bn , k_bl) = y_SrvD%BlFlapCom(k_bl) !bem: This takes in flap deflection for each blade (only one flap deflection angle per blade) from ServoDyn (which comes from Bladed style DLL controller)
```
bjonkman: I'd personally prefer not to use the `p_AD` data type in the glue code unless absolutely necessary. In this case, you could loop from `k=1 to size(u_AD%UserProp,2)` and set `u_AD%UserProp(:,k) = y_SrvD%BlFlapCom(k)`

Then you don't need to pass the AD parameter type to this routine at all.

bjonkman: Also, you should put an `if (p_FAST%CompServo == Module_SrvD` around this so you don't get seg faults if you use AeroDyn but not ServoDyn.

modules/servodyn/src/ServoDyn_Registry.txt
```diff
@@ -133,6 +133,7 @@ typedef	^	BladedDLLType	IntKi	GenState	-	-	-	"Generator state from Bladed DLL"	N
 #typedef	^	BladedDLLType	ReKi	ElecPwr	-	-	-	"Electrical power sent to Bladed DLL"	W
 typedef	^	BladedDLLType	ReKi	BlPitchCom	3	-	-	"Commanded blade pitch angles"	radians
 typedef	^	BladedDLLType	ReKi	PrevBlPitch	3	-	-	"Previously commanded blade pitch angles"	radians
+typedef	^	BladedDLLType	ReKi	BlFlapCom	3	-	-	"Commanded lap deflection angles"	degrees
```
bjonkman: I **highly recommend** using radians instead of degrees. Everything else in the Bladed interface and everything that is passed between modules in the FAST code uses radians.

modules/servodyn/src/ServoDyn.f90
bjonkman: Also, if ServoDyn is not using the DLL for control, you will be using uninitialized data here. Either put an `if` statement around setting `y%BlFlapCom`, or initialize m%dll_data%BlFlapCom to 0.

andrew-platt: Good point.  I'm adding your changes.

modules/servodyn/src/ServoDyn_Registry.txt
andrew-platt: I set this to radians and changed the names to more generalized names for those channels.

modules/openfast-library/src/FAST_Solver.f90
andrew-platt: Good catch!  I've added this as well.

bjonkman:
Looks good, @andrew-platt!

Issue #385, opened 2020-01-26
Bug report: MoorDyn crashes when no output channels are assigned


andyen:
**Bug description**
MoorDyn and consequently OpenFAST crashes if no output channels are configured in the `Moordyn.dat`. The corresponding output (reported to STDERR) looks like this:

```
At line 1210 of file /home/andy/openfast-2.1.0/modules-ext/moordyn/src/MoorDyn_IO.f90
Fortran runtime error: Expected P edit descriptor in format
(F10.4,0(A1,e10.4))
^

Error termination. Backtrace:
#0 0x7f8cfc5a1cb1 in ???
#1 0x7f8cfc5a2809 in ???
#2 0x7f8cfc5a328f in ???
#3 0x7f8cfc7b96c9 in ???
#4 0x7f8cfc7b990e in ???
#5 0x7f8cfc7c9e07 in ???
#6 0x563a39eb7f07 in ???
#7 0x563a39eb0190 in ???
#8 0x563a39184658 in ???
#9 0x563a3918d386 in ???
#10 0x563a3902e2c4 in ???
#11 0x563a3902ea61 in ???
#12 0x563a38fdbf46 in ???
#13 0x563a38f25c6e in ???
#14 0x7f8cfc1d71e2 in ???
#15 0x563a38f25cad in ???
#16 0xffffffffffffffff in ???
```

**To Reproduce**
The bug can be reproduced by simply removing all output channels from the `Moordyn.dat` file. Starting line 34, the file will look like:

```
------------------------ OUTPUTS --------------------------------------------
END
------------------------- need this line --------------------------------------
```

**OpenFAST Version**
OpenFAST Version 2.1.0

```
**************************************************************************************************
OpenFAST

Copyright (C) National Renewable Energy Laboratory
Copyright (C) Envision Energy USA LTD

This program is licensed under Apache License Version 2.0 and comes with ABSOLUTELY NO WARRANTY.
See the "LICENSE" file distributed with this software for details.
**************************************************************************************************

OpenFAST-GIT-NOTFOUND
Compile Info:
- Architecture: 64 bit
- Precision: double
- Date: Nov 26 2019
- Time: 20:22:05
Execution Info:
- Date: 01/26/2020
- Time: 19:46:22+0100
```

**System Information (please complete the following information):**
 - OS: Ubuntu 19.10

Fortran Compiler:
```
andy@any:~/openfast-2.1.0$ gfortran -v
Using built-in specs.
[...]

PR #404, opened 2020-02-28
[BugFix] MoorDyn crash if no outputs requested

andrew-platt:
This is ready to merge

If no output channels were requested from MoorDyn, it would cause OpenFAST to crash (see #385).

This PR adds a check on number of outputs requested and bypasses the entire subroutine.

There should be no changes in the regression tests.


andrew-platt:
@mattEhall please review

andrew-platt:
The issue was with an ill formed format statement when `p%NumOuts==0` (https://github.com/OpenFAST/openfast/blob/2372174c7871435609861546aeebe6615299a988/modules/moordyn/src/MoorDyn_IO.f90#L1213)

PR #408, opened 2020-03-05
Improve command line argument handling

rafmudaf:
**Complete this sentence**

This pull request rewrites the subroutine for handling the input arguments. It explicitly checks for arguments in a particular position and assigns them to the appropriate variable within the routine. Further, these changes allow the help-prompt to exit OpenFAST normally resulting in an exit code of 0.

Another change is that the errors in argument parsing are handled when they are checked rather than when OpenFAST is attempting to read the input file.

Finally, the unit test CMake configuration is changed to better support unit testing in multiple modules.

This solves https://github.com/OpenFAST/openfast/issues/105.

Argument parsing (`CheckArgs`) and one line in FAST_Prog

As a very minor improvement, this makes the help prompt display quickly without that extra second of time and added confusion of telling the user that an input file wasn't given.

But more important supporting information is that systems can now check for a "valid" OpenFAST binary by running `openfast -h` and evaluating the return code. Of course, this doesn't verify that accuracy like running the regression tests, but it verifies that the binary is compatible with the system.

See [GitHub Actions](https://github.com/rafmudaf/openfast/runs/488349517).

---

**TODO**
- [x] Update unit test documentation.


rafmudaf:
cc @bjonkman

bjonkman:
One comment I have is that your changes will alter some functionality in TurbSim (and possibly other codes). The routine was designed to allow a code to specify a default input file name (hence the `INTENT(INOUT)` on the `InputFile` argument). In TurbSim, it is valid to call the code using 

```
TurbSim.exe
```
and it is also valid to call it using 
```
TurbSim.exe InputFileName
```
With this PR, the first syntax will no longer be valid.

My second comment is that this PR does not set the `Arg2` argument, which is supposed to return the last non-flag argument to FAST_Subs.f90's `GetInputFileName` subroutine . It will now be using uninitialized data.

rafmudaf:
Well, here I was so proud of myself :( but that's why I asked! Thanks for pointing this out.

bjonkman:
It does look much cleaner. :) It's just extra complicated because that routine gets called from a lot of different places with different use cases.

rafmudaf:
What does `PRESENT()` mean in the context of an `INTENT(OUT)` variable?
https://github.com/OpenFAST/openfast/blob/dev/modules/nwtc-library/src/NWTC_IO.f90#L1505

rafmudaf:
Oh - `PRESENT()` is in reference to the `OPTIONAL` part not the `INTENT` 👍

modules/nwtc-library/src/NWTC_IO.f90
```diff
@@ -1475,94 +1475,105 @@ SUBROUTINE AllR16Ary5 (  Ary, AryDim1, AryDim2, AryDim3, AryDim4, AryDim5, Descr
...
-   
+   INTEGER                               :: NumArg             ! The number of arguments on the command line.
+   CHARACTER(LEN(InputFile))             :: Arg                ! A command-line argument.
 
       ! Find out how many arguments were entered on the command line.
-   NumArg   = COMMAND_ARGUMENT_COUNT()
-   FirstArg = .TRUE.
+   NumArg = COMMAND_ARGUMENT_COUNT()
 
-   IF ( PRESENT(Arg2) ) Arg2 = ""
```
bjonkman: I think you should leave these lines in so that these (optional) arguments always get initialized, regardless of how many command line arguments exist.

rafmudaf: Good point - done

modules/nwtc-library/src/NWTC_IO.f90
```diff
@@ -1475,94 +1475,105 @@ SUBROUTINE AllR16Ary5 (  Ary, AryDim1, AryDim2, AryDim3, AryDim4, AryDim5, Descr
...
-   IF ( PRESENT(flag) ) flag = ""
+      ! Early return if no arguments given
+   IF ( NumArg == 0 ) CALL INVALID_SYNTAX( 'no command-line arguments given.' )
```
bjonkman: Perhaps you should check `IF (TRIM_LEN(InputFile)==0)` before throwing an error on no command-line arguments; otherwise, just return. That would take care of the issue with allowing a default file name.

rafmudaf: Its not currently used this way, but one could provide a default input file and expect other flags. So I'm thinking to do this at the moment:

```
   IF ( LEN(InputFile) == 0 ) THEN
      Arg = InputFile
   ELSE
      CALL GET_COMMAND_LINE_ARG( 1, Arg, PRESENT(ErrStat))
   END IF
```

Thought its kind of silly because a couple of lines later is 
```
      InputFile = Arg
```

modules/nwtc-library/src/NWTC_IO.f90
```diff
@@ -1475,94 +1475,105 @@ SUBROUTINE AllR16Ary5 (  Ary, AryDim1, AryDim2, AryDim3, AryDim4, AryDim5, Descr
...
+   CALL GET_COMMAND_LINE_ARG( 1, Arg, PRESENT(ErrStat))
 
-   IF ( NumArg .GT. 0 ) THEN
+      ! If the first argument is not one of the switch characters, then assume its the input file
+   IF ( Arg(1:1) /= SwChar .AND. Arg(1:1) /= '-' ) THEN
+      InputFile = Arg
 
-      DO IArg=1,NumArg
+      ! Parse the additional second argument, if it exists
+      IF ( NumArg == 2 .AND. PRESENT(Arg2) ) CALL GET_COMMAND_LINE_ARG( 2, Arg2, PRESENT(ErrStat))
```
bjonkman: Your second argument here (`Arg2`) may or may not be a flag.

I would say that you need to allow for at least 
```
executable.exe -Flag FileName Arg2
executable.exe -Flag FileName
executable.exe FileName Arg2
executable.exe FileName
executable.exe
```
which I think you mostly have here. The previous logic also allowed the flags to be in any position on the line, like 
```
executable.exe FileName -Flag
executable.exe FileName -Flag Arg2
executable.exe FileName Arg2 -IgnoredFlagUnlessItIsHELP -Flag
```
While it is probably not necessary to reproduce that exact logic, it should be documented that it has changed, particularly since you won't be able to throw the `-h` flag anywhere except in the first argument.

modules/nwtc-library/src/NWTC_IO.f90
```diff
@@ -1475,94 +1475,155 @@ SUBROUTINE AllR16Ary5 (  Ary, AryDim1, AryDim2, AryDim3, AryDim4, AryDim5, Descr
...
-         END IF ! ( Arg(1:1) == SwChar )
+      CASE ('H')
+         CALL NWTC_DisplaySyntax( Arg1, ProgName )
+         CALL CLEANUP()
+         RETURN
 
-      END DO ! IArg
+      CASE ('RESTART')
+         IF ( .NOT. FirstArgumentSet .OR. .NOT. SecondArgumentSet ) THEN
+            CALL INVALID_SYNTAX( 'the restart capability requires and first and second argument: <input_file> -restart <checkpoint_file>' )
```
bjonkman: Is there a reason that you are now requiring an extra argument for the restart capability?

The previous syntax was `<name of FAST executable > -restart <root name of checkpoint file without extension>`, i.e., `-restart <checkpoint_file>` with no need to supply `<input_file>`. The input file isn't used in restart, so I'd rather not have to supply an extra argument for it. The checkpoint file then *is* the input file.

rafmudaf: Fixed! Thanks for pointing that out.

rafmudaf:
@bjonkman This isn't as clean as it once was, but it still provides the needed functionality and bug fix while also introducing unit tests. Do you have any comments on the current state of the pull request?

andrew-platt:
I did a quick check of the uses of the CheckArgs routine.  I checked all the ones that I think should work as before:

- [ ] aerodyn/src/AeroDyn_Driver_Subs.f90
- [ ] aerodyn14/src/DWM_driver_wind_farm_sub.f90
- [ ] hydrodyn/src/SS_Radiation_DriverCode.f90
- [ ] openfast-library/src/FAST_Subs.f90
- [ ] servodyn/src/ServoDyn_Driver.f90
- [ ] turbsim/src/TurbSim.f90

The only two that I think could be problematic are SS_Radiation and DWM.

PR #411, opened 2020-03-11
Update cpp API to provide consistent definition of hub position

psakievich:
**Complete this sentence**
THIS PULL REQUEST  IS READY TO MERGE

The C++ API currently defines a hub position that never gets updated or checked against the force location.  This can lead to serious bugs if the users specify a hub position that is quite different than the calculated hub position from elastodyn (i.e. a different height). These changes synchronize the hub position across all processors after OpenFAST is initialized so the hub position is the C++ API is the initial position of velocity/force point for the hub.

### DIFF
--- a/glue-codes/openfast-cpp/src/OpenFAST.H
+++ b/glue-codes/openfast-cpp/src/OpenFAST.H
@@ -161,7 +161,7 @@ class OpenFAST {
   OpenFAST() ;
   
   // Destructor
-  ~OpenFAST() {} ;
+  ~OpenFAST() ;
 
   void setInputs(const fastInputs &);  
 
--- a/glue-codes/openfast-cpp/src/OpenFAST.cpp
+++ b/glue-codes/openfast-cpp/src/OpenFAST.cpp
@@ -194,7 +194,7 @@ void fast::OpenFAST::init() {
       if(positionChange){
         std::vector<double> temp(globTurbineData[i].TurbineHubPos);
         MPI_Bcast(globTurbineData[i].TurbineHubPos.data(), 3, MPI_DOUBLE, get_procNo(i), mpiComm);
-        if(isDebug()){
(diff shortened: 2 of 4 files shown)


gantech:
The C++ API right now provides two functions

`getApproxHubPos` - Can be used to determine which processor to place the openfast instance for this turbine. This is available to use **before** openfast is instantiated for this turbine. 
https://github.com/OpenFAST/openfast/blob/master/glue-codes/openfast-cpp/src/OpenFAST.cpp#L443

`getHubPos` - Can be used to determine the exact location of the hub at any given simulation time. This is available to use only **after** openfast is instantiated for this turbine. 
https://github.com/OpenFAST/openfast/blob/master/glue-codes/openfast-cpp/src/OpenFAST.cpp#L452

May be we could add documentation to clarify the difference between the two? I don't see a reason to use `globTurbineData[i].TurbineHubPos` directly.

gantech:
May be we should get rid of this approximate hub position all together. The original reason for this was to place the OpenFAST instance of the turbine on a processor containing the hub. But may be this will become irrelevant soon.

psakievich:
@gantech I'd be fine getting rid of it.  I'm planning to deprecate it as an input value in nalu-wind since we won't need it any more.

psakievich:
I will update my code to remove all references to `getApproxHubPos`

rafmudaf:
@gantech could you take a look and approve or comment on this?

gantech:
@psakievich Are you ok with removing this pull request if you're removing all references to `getApproxHubPos`  in your refactor?

psakievich:
@gantech I will close the PR if you want to deprecate this feature/leave it the way it is. I'm not using it because of its not parallel consistent. I would probably find it useful if it were parallel consistent, but I can work around it.

PR #428, opened 2020-03-26
Add a version flag and other printout updates

rafmudaf:
**Complete this sentence**

This pull request adds support for a version flag in the `CheckArgs` subroutine. Passing the lower case or upper case form of `-v` or `-version` will print the version info of the program being run.

The method to get version information was modified to allow a version string to be passed in the command line via CMake. This is useful in cases where the source code is downloaded as a tarball rather than cloned with git. In this case, there is no `.git` directory and therefore a version isn't automatically set.

Finally, the routines and formatting for printing general information to the console are made more explicit and consistent. A compiler description is also added to the compile information section of the standard printout.

None

NWTC Library, drivers and glue codes

GItHub Actions.


rafmudaf:
Here's an example of the version when the source code is downloaded outside of git: https://github.com/Homebrew/homebrew-core/pull/52107/checks?check_run_id=535445810#step:3:1596

This is an automatic build inside Homebrew. In this case, an OpenFAST binary distributed through that package manager would be unversioned.

bjonkman:
I haven't read through all of your changes, but am wondering if you have you tried to use the `.gitattributes` file for some of this? I added the 
```
CreateGitVersion.bat ident export-subst
```
line in it so that when you download a zip file from GitHub, it replaces a string in the .bat file and Visual Studio builds can still tell us what version it originated with. Might be nice to just have a header file with this info in it so either CMake or Visual Studio could access it.

rafmudaf:
@bjonkman It hadn't occurred to me to use `.gitattributes`, but I like the idea of using that to create a version header file and consolidating all of the versioning infrastructure.

rafmudaf:
@bjonkman I think the suggestion to create a header file will work, but I haven't seen how to get the latest tag information with gitattributes. This is required in order to recreate out current version. One option is to hard code the latest tag and then append the current git info to the end. In any case, I'd like to think a bit more for a better solution to that.

As for this pull request, do you have any other comments?

PR #437, opened 2020-04-16
Feature/update if w hawc

ewquon:
**Complete this sentence**

Add an additional option to the `WindProfile` parameter: -1 is allowable so that the mean profile is not added on. Use case would be when a user wants to directly supply inflow planes that include both the mean wind and turbulence, e.g., when using sampled output from SOWFA and using it as inflow data.


andrew-platt:
Also added the ability to shift the HAWC wind.  This is added into the InflowWind input files as the last line in the `Mean wind profile parameters (added to HAWC-format files)`

```
         0    InitPosition(x) - Initial offset in +x direction (shift of wind box)
```

PR #456, opened 2020-06-05
Properly support yaw rate command integration from controller

### LINKED ISSUE #25: Bug in ServoDyn When Controlling Yaw from DLL
In routine CalculateStandardYaw of ServoDyn.f90, the yaw angle (position) command when the yaw rate is commanded from a Bladed-style DLL is calculated by adding an increment of YawRateCom*DT to the current yaw angle. However, the yaw angle command should be defined as the integral of the yaw rate command, regardless of the current yaw angle (i.e. the yaw angle command must become a state of ServoDyn). As it is currently implemented (incorrectly), when the yaw rate commanded from a Bladed-style DLL is zero, the yaw angle is commanded to be current yaw angle, effectively eliminating the yaw spring stiffness, yielding large yaw errors. I would not use yaw control from Bladed-style DLL controllers until this bug is fixed.

See the following forum topic for more information: https://wind.nrel.gov/forum/wind/viewtopic.php?f=30&t=1844&p=9256.

andrew-platt:

Integration of the yawrate command from the DLL is now properly supported.
This has been tested by @nikhar-abbas and @ewquon

Issue #25 

This resolves issue #25.  When a DLL was controlling the yaw, the position was incorrectly integrated.

No test cases are affected by this PR as none of the glue-code DLL cases use yaw control.

Closes #25

### DIFF
--- a/modules/servodyn/src/ServoDyn.f90
+++ b/modules/servodyn/src/ServoDyn.f90
@@ -467,9 +467,8 @@ SUBROUTINE SrvD_Init( InitInp, u, p, x, xd, z, OtherState, y, m, Interval, InitO
 
       !............................................................................................
       ! yaw control integrated command angle
-      !  -- this really should be in OtherStates, but it is calculated in calcoutput where otherstates is intent(in)
       !............................................................................................
-   m%YawPosComInt = 0.0_ReKi
+   OtherState%YawPosComInt = p%YawNeut 
    
    
       !............................................................................................
@@ -2833,7 +2832,8 @@ SUBROUTINE Yaw_CalcOutput( t, u, p, x, xd, z, OtherState, y, m, ErrStat, ErrMsg 
    ! Calculate standard yaw position and rate commands:
    !...................................................................
 
-   CALL CalculateStandardYaw(t, u, p, m, YawPosCom, YawRateCom, m%YawPosComInt, ErrStat, ErrMsg)
+   YawPosComInt = OtherState%YawPosComInt    ! get state value.  We don't update the state here.
+   CALL CalculateStandardYaw(t, u, p, m, YawPosCom, YawRateCom, YawPosComInt, ErrStat, ErrMsg)
          
    !...................................................................
    ! Override standard yaw control with a linear maneuver if necessary:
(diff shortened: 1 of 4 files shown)


modules/servodyn/src/ServoDyn.f90
```diff
@@ -2954,7 +2973,7 @@ SUBROUTINE Yaw_UpdateStates( t, u, p, x, xd, z, OtherState, m, ErrStat, ErrMsg )
 
       IF ( .not. OtherState%BegYawMan )  THEN  ! Override yaw maneuver is just beginning (possibly again).
 
-         CALL CalculateStandardYaw(t, u, p, m, YawPosCom, YawRateCom, ErrStat, ErrMsg)
+         CALL CalculateStandardYaw(t, u, p, m, YawPosCom, YawRateCom, OtherState%YawPosComInt, ErrStat, ErrMsg)
```
bjonkman, line 2976: It looks like you could potentially be updating the OtherState%YawPosComInt twice in this call (i.e., taking two steps instead of just one).

modules/servodyn/src/ServoDyn.f90
```diff
@@ -464,6 +464,11 @@ SUBROUTINE SrvD_Init( InitInp, u, p, x, xd, z, OtherState, y, m, Interval, InitO
    OtherState%TTpBrFl = HUGE(OtherState%TTpBrFl) !basically never deploy them. Eventually this will be added back?
    !OtherState%TTpBrFl = InputFileData%TTpBrFl + p%TpBrDT
 
+
+      !............................................................................................
+      ! yaw control integrated command angle
+      !............................................................................................
+   OtherState%YawPosComInt = p%YawNeut
```
bjonkman, line 472: Do we need to take into account the initial yaw position from ElastoDyn here?

modules/servodyn/src/ServoDyn.f90
andrew-platt: I missed this bit of logic.  The second call to `CalculateStandadYaw` is during an override yaw manuever and should only be called once while `OtherState%BegYawMan` is false.  If we are doing an over-ride of the yaw from the DLL, I would expect the DLL commanded yaw should be completely out of the loop then, so maybe the first first `CalculateStandardYaw should be in an `else` condition on `t >= p%TYawManS`.

PR #486, opened 2020-07-06
 MoorDyn bug fix for tensions changing part-way through simulation.

mattEhall:
- Tension amplitudes were sometimes changing, typically at 512 s.
 - This was likely caused by use of single-precision time variable with MD_Input_ExtrapInterp.
 - All time variables are now double precision, solving the problem.

**Complete this sentence**
THIS PULL REQUEST __ IS __ READY TO MERGE

Time variables in MoorDyn now always double precision.

https://wind.nrel.gov/forum/wind/viewtopic.php?f=4&t=1156&start=45#p14866

No impact aside from eliminating tension output bug.


andrew-platt:
Since all regression tests are run in double precision, no test cases are affected (`ReKi` is double precision during regression tests).

PR #488, opened 2020-07-08
Fix bug that could cause incorrect loads transfer in Line2-to-Line2 or Line2-to-Point mapping

bjonkman:
**Complete this sentence**

This pull request fixes a problem identified in the implementation of the mesh mapping line2-to-line2 and line2-to-point loads-transfer algorithm. The algorithm to create the augmented source mesh would split an element but--in certain (fairly rare?) instances--incorrectly calculate the position of the new node, effectively increasing the load being transferred.

None

This may impact any modules that have distributed loads output. This includes any models that use
- AeroDyn15 to transfer loads to ElastoDyn or BeamDyn, or
- HydroDyn to transfer loads to SubDyn.

I have attached a plot showing a case where this bug was triggered. The red line shows the smooth loads expected from AeroDyn's output mesh; the blue line indicates how the loads input to the structural code were being calculated prior to this bug fix:
problem2

The regression tests look the same (with the exception of the numerical instability in the SWRT case), indicating that this bug probably wasn't triggered in any of these test cases.


jjonkman:
I've discussed this previously unknown bug and fix with @bjonkman and approve this pull request.

andrew-platt:
Good catch!  I think I've seen this bug once with a BeamDyn test case, but thought it was just something odd with BeamDyn.

Issue #504, opened 2020-07-19
Compilation warning in NWTC_IO.f90


sayerhs:
Building OpenFAST on Linux with GCC 7.4.0 on ORNL Summit issues the following warning:

```
[  1%] Building Fortran object modules/nwtc-library/CMakeFiles/nwtclibs.dir/src/NWTC_IO.f90.o
/ccs/proj/cfd142/exawind/exawind-2020-07/source/openfast/modules/nwtc-library/src/NWTC_IO.f90:2248:10:

       USE iso_fortran_env
          1
Warning: Use of the NUMERIC_STORAGE_SIZE named constant from intrinsic module ISO_FORTRAN_ENV at (1) is incompatible with option -fdefault-real-8
```

**OpenFAST commit**: 5aacf65c

```
5aacf65c Merge pull request #447 from andrew-platt/f/vc
```

**System information:**

```
sayerhs@login1:~$ lsb_release -d ; uname -a ; gfortran --version
Description:	Red Hat Enterprise Linux Server release 7.6 (Maipo)
Linux login1 4.14.0-115.21.2.el7a.ppc64le #1 SMP Thu May 7 22:22:31 UTC 2020 ppc64le ppc64le ppc64le GNU/Linux
GNU Fortran (GCC) 7.4.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

**CMake configure command**
```
cmake 
  -DCMAKE_INSTALL_PREFIX:PATH=/ccs/proj/cfd142/exawind/exawind-2020-07/install/gcc/openfast 
  -DCMAKE_BUILD_TYPE=RELWITHDEBINFO
  -DBUILD_SHARED_LIBS:BOOL=OFF 
  -DFPE_TRAP_ENABLED:BOOL=ON 
  -DUSE_DLL_INTERFACE:BOOL=ON 
  -DBUILD_OPENFAST_CPP_API:BOOL=ON 
  -DYAML_ROOT:PATH=/autofs/nccs-svm1_proj/cfd142/exawind/exawind-2020-07/spack/opt/spack/linux-rhel7-power9le/gcc-7.4.0/yaml-cpp-0.6.2-6gmppkg63e3mubizsarxetpsqavgoqex 
[...]

PR #507, opened 2020-07-20
Fix a compiler warning regarding the iso_fortran_env module

rafmudaf:
When compiling `nwtclibs`, the a recent addition that uses the `iso_fortran_env` module leads to a compiler warning regarding an incompatibility in another defined parameter in that module. This fix is limited to the compiler warning since the referenced module parameter was not used.

#504

NWTC_IO

### DIFF
--- a/modules/nwtc-library/src/NWTC_IO.f90
+++ b/modules/nwtc-library/src/NWTC_IO.f90
@@ -2245,8 +2245,8 @@ END SUBROUTINE CheckR16Var
 !>
    SUBROUTINE DispCompileRuntimeInfo()
      
-      USE iso_fortran_env
-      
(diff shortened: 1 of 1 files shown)


andrew-platt:
Good idea.  I've been wondering about that warning.

andrew-platt:
I think we have a problem with old compilers.  Nick just ran into it with Intel Fortran Compiler 17.0.2.187 and Visual Studio Community 2015.

sayerhs:
@andrew-platt I believe this should fix issues with `ISO_FORTRAN_ENV` https://github.com/OpenFAST/openfast/pull/512

andrew-platt:
@sayerhs Yes that should fix it.  I see that is still listed as a draft.  Is there anything further required before we can merge it?

Issue #89, opened 2018-02-15
Implement the RanLux Pseudo-Random Number in HydroDyn For Wave Generation


jjonkman:
HydroDyn invokes an intrinsic pseudo-random number generator (pRNG) to derive the internal wave kinematics time series from the wave frequency and direction spectra.  However, the intrinsic pRNG is different across different compilers and operating systems, making cross comparisons of time-domain response difficult.

TurbSim also uses a pRNG to derive wind time series from the wind spectrum (among other uses).  In addition to an intrinsic pRNG, TurbSim also has a built-in pRNG named RanLux (implemented in Fortran source code).  If RanLux where moved to the NWTC_Library, it could be used by FAST modules.  It is recommended to add the RanLux pRNG to the NWTC_Library and use it as an optional pRNG within HydroDyn.  (The intrinsic pRNG is initialized with 2 seeds; the RanLux pRNG is initialized with 1 seed and is selected by setting the second random seed to the RanLux keyword.)  Once implemented, when the RanLux pRNG is selected in HydroDyn, it will then be possible to have identical wave kimenatics time series to enable cross comparisons of time-domain responses across different compilers and operating systems.

rafmudaf:
Using a HydroDyn standalone test case with irregular waves (WaveMod 2), I've found that the wave elevation calculation differs between GNU and Intel compilers, as shown in the plot below.

![compiler_diff](https://user-images.githubusercontent.com/13797903/87964988-ac167900-ca80-11ea-9b36-3f38f260aa97.png)

PR #508, opened 2020-07-21
Add RANLUX to NWTC Library / HydroDyn

rafmudaf:
It has been observed that HydroDyn wave kinematics are subject to influence from the compiler and operating system. This is due to the use of an intrinsic pRNG in the `Waves` module. This pull request introduced a module providing random number generation subroutines to the NWTC Library. This was taken from a similar TurbSim module that includes various methods for pRNG generation. Here, I've only included RANLUX, but the module is flexible and can be extended to use other methods.

Also included is a HydroDyn module-level regression test. Initially, the test uses the Fortran intrinsic `RANDOM_NUMBER` subroutine, and the results are inconsistent between Intel and GNU compilers. After adding RANLUX, the test case passes for both compilers.

The TurbSim module and its random number routines have not been modified.

#89 

NWTC Library and HydroDyn

The HydroDyn input file now supports choosing the RANLUX pRNG, but fully supports using the Fortran intrinsic subroutine. If the user provides a second seed, the Fortran intrinsic is used. If "RANLUX" is provided instead, that pRNG is used.

This should remove some of the variations we have seen in compiler/machine sensitivities in the OpenFAST regression tests. Once this is accepted, I'll update the appropriate regression tests to use this method in another pull request.

Using `RANDOM_NUMBER`, this is the wave elevation calculation from GNU and Intel compilers:
compiler_diff

Using RANLUX, the both align
compiler_same


rafmudaf:
@andrew-platt could you please review this when you have a chance?

andrew-platt:
Did any of the visual studio projects get updated?

rafmudaf:
Negative, good catch.

PR #454, opened 2020-06-02
Fix for compiler_version() not available with old intel versions

ebranlard:
**Complete this sentence**
THIS PULL REQUEST IS READY TO MERGE

Should fix the issue discussed here:
https://github.com/OpenFAST/openfast/commit/59a520b4e8ee871e34edc997f7fc8e49423917c7#r39157954

PR #512, opened 2020-07-23
Introduce checks for Fortran 2008 features

### LINKED ISSUE #502: Compilation error on PowerPC64 due to unsupported real kind
OpenFAST fails to build on IBM PowerPC64 architecture using GNU GCC 7.4.0 ([ORNL Summit](https://www.olcf.ornl.gov/summit/)) due to unsupported real kind used for `QuKi`. The compiler error message is 

```
    REAL(DbKi), INTENT(IN)       :: DblNum
            1
Error: Kind -2 not supported for type REAL at (1)


    REAL(QuKi), INTENT(IN)     :: x             ! input
            1
Error: Kind -2 not supported for type REAL at (1)
```

I can workaround for now by using the following patch. However, this is not a robust fix. Given the coupling using C++ API, I don't believe `-DDOUBLE_PRECISION:BOOL=OFF` is a viable option. 

```diff
diff --git a/modules/nwtc-library/src/SingPrec.f90 b/modules/nwtc-library/src/SingPrec.f90
index c8b4bca3..5864d8ff 100644
--- a/modules/nwtc-library/src/SingPrec.f90
+++ b/modules/nwtc-library/src/SingPrec.f90
@@ -35,7 +35,11 @@ INTEGER, PARAMETER              :: B2Ki     = SELECTED_INT_KIND(  4 )
 INTEGER, PARAMETER              :: B4Ki     = SELECTED_INT_KIND(  9 )           !< Kind for four-byte whole numbers
 INTEGER, PARAMETER              :: B8Ki     = SELECTED_INT_KIND( 18 )           !< Kind for eight-byte whole numbers

+#ifdef __GFC_REAL_16__
 INTEGER, PARAMETER              :: QuKi     = SELECTED_REAL_KIND( 20, 500 )     !< Kind for 16-byte, floating-point numbers
+#else
+INTEGER, PARAMETER              :: QuKi     = SELECTED_REAL_KIND( 20, 120 )     !< Kind for 16-byte, floating-point numbers
+#endif
 INTEGER, PARAMETER              :: R8Ki     = SELECTED_REAL_KIND( 14, 300 )     !< Kind for eight-byte floating-point numbers
 INTEGER, PARAMETER              :: SiKi     = SELECTED_REAL_KIND(  6,  30 )     !< Kind for four-byte, floating-point numbers
```

**OpenFAST commit**: 5aacf65c

```
5aacf65c Merge pull request #447 from andrew-platt/f/vc
```

**System information:**

```
sayerhs@login1:~$ lsb_release -d
Description:	Red Hat Enterprise Linux Server release 7.6 (Maipo)
sayerhs@login1:~$ uname -a
Linux login1 4.14.0-115.21.2.el7a.ppc64le #1 SMP Thu May 7 22:22:31 UTC 2020 ppc64le ppc64le ppc64le GNU/Linux
sayerhs@login1:~$ gfortran --version
GNU Fortran (GCC) 7.4.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

**CMake configure command**
```
cmake 
  -DCMAKE_INSTALL_PREFIX:PATH=/ccs/proj/cfd142/exawind/exawind-2020-07/install/gcc/openfast 
  -DCMAKE_BUILD_TYPE=RELEASE 
  -DBUILD_SHARED_LIBS:BOOL=OFF 
  -DFPE_TRAP_ENABLED:BOOL=ON 
  -DUSE_DLL_INTERFACE:BOOL=ON 
  -DBUILD_OPENFAST_CPP_API:BOOL=ON 
  -DYAML_ROOT:PATH=/autofs/nccs-svm1_proj/cfd142/exawind/exawind-2020-07/spack/opt/spack/linux-rhel7-power9le/gcc-7.4.0/yaml-cpp-0.6.2-6gmppkg63e3mubizsarxetpsqavgoqex 
  -DHDF5_ROOT:PATH=/autofs/nccs-svm1_proj/cfd142/exawind/exawind-2020-07/spack/opt/spack/linux-rhel7-power9le/gcc-7.4.0/hdf5-1.10.4-5jqpjwnfoupcjas2fnjbh54t3wkiorq7 
  -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=ON 
  -DCMAKE_INSTALL_RPATH_USE_LINK_PATH:BOOL=ON ..
```

sayerhs:
- Add new compile-time definition HAS_FORTRAN2008_FEATURES
- Add new macro in CMake to check for features and enable the flag
- Add ifdef statements as necessary to protect F2008 usage

See #502 for a discussion motivating the proposed change. Impacts CMake configuration logic and NWTC_Library. 

Closes #502

### DIFF
--- a/cmake/OpenfastFortranOptions.cmake
+++ b/cmake/OpenfastFortranOptions.cmake
@@ -74,6 +74,27 @@ macro(set_fast_fortran)
 endmacro(set_fast_fortran)
 
 #
+# CHECK_F2008_FEATURES - Check if Fortran2008 features are available
+#
+macro(check_f2008_features)
+  include(CheckFortranSourceCompiles)
+  check_fortran_source_compiles(
+    "program test
+     use iso_fortran_env, only: compiler_version, real32, real64, real128
+     integer, parameter :: quki = real128
+     integer, parameter :: dbki = real64
+     integer, parameter :: reki = real32
+
+     end program test"
+     HAS_FORTRAN2008
+     SRC_EXT F90)
+   if (HAS_FORTRAN2008)
+     message(STATUS "Enabling Fortran 2008 features")
+     add_definitions(-DHAS_FORTRAN2008_FEATURES)
+   endif()
+endmacro(check_f2008_features)
+
+#
 # SET_FAST_GFORTRAN - Customizations for GNU Fortran compiler
 #
 macro(set_fast_gfortran)
@@ -109,6 +130,7 @@ macro(set_fast_gfortran)
      set(CMAKE_Fortran_FLAGS_DEBUG "${CMAKE_Fortran_FLAGS_DEBUG} -fopenmp" )
   endif()
 
+  check_f2008_features()
 endmacro(set_fast_gfortran)
 
 #
@@ -144,6 +166,8 @@ macro(set_fast_intel_fortran_posix)
      set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -qopenmp")
      set(CMAKE_Fortran_FLAGS_DEBUG "${CMAKE_Fortran_FLAGS_DEBUG} -qopenmp" )
   endif()
+
+  check_f2008_features()
 endmacro(set_fast_intel_fortran_posix)
 
 #
@@ -177,4 +201,5 @@ macro(set_fast_intel_fortran_windows)
      set(CMAKE_Fortran_FLAGS_DEBUG "${CMAKE_Fortran_FLAGS_DEBUG} /qopenmp" )
   endif()
 
+  check_f2008_features()
 endmacro(set_fast_intel_fortran_windows)
--- a/modules/nwtc-library/src/NWTC_IO.f90
+++ b/modules/nwtc-library/src/NWTC_IO.f90
@@ -2244,9 +2244,9 @@ END SUBROUTINE CheckR16Var
 !=======================================================================
 !>
    SUBROUTINE DispCompileRuntimeInfo()
-     
+#ifdef HAS_FORTRAN2008_FEATURES
       USE iso_fortran_env, ONLY: compiler_version
-
+#endif
       CHARACTER(200) :: compiler_version_str
       CHARACTER(200) :: name
       CHARACTER(200) :: git_commit, architecture, compiled_precision
@@ -2263,7 +2263,7 @@ END SUBROUTINE CheckR16Var
          compiled_precision = 'unknown'
       END IF
 
-#if defined(__INTEL_COMPILER) && (__INTEL_COMPILER<1800)
+#ifndef HAS_FORTRAN2008_FEATURES
       compiler_version_str = 'Intel(R) Fortran Compiler '//num2lstr(__INTEL_COMPILER)
 #else
       compiler_version_str = compiler_version()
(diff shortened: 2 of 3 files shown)


modules/nwtc-library/src/NWTC_IO.f90
```diff
@@ -2263,7 +2263,7 @@ SUBROUTINE DispCompileRuntimeInfo()
          compiled_precision = 'unknown'
       END IF
 
-#if defined(__INTEL_COMPILER) && (__INTEL_COMPILER<1800)
+#ifndef HAS_FORTRAN2008_FEATURES
```
sayerhs, line 2266:
```suggestion
#if defined(__INTEL_COMPILER) && !defined(HAS_FORTRAN2008_FEATURES)
```

bjonkman, line 2266: I'm a little concerned that a non-Intel compiler could get into this ifndef branch and then have trouble because `__INTEL_COMPILER` isn't defined on the next line.

cmake/OpenfastFortranOptions.cmake
```diff
@@ -73,6 +73,27 @@ macro(set_fast_fortran)
...
+     integer, parameter :: quki = real128
+     integer, parameter :: dbki = real64
+     integer, parameter :: reki = real32
+
+     end program test"
+     HAS_FORTRAN2008
+     SRC_EXT F90)
+   if (HAS_FORTRAN2008)
+     message(STATUS "Enabling Fortran 2008 features")
+     add_definitions(-DHAS_FORTRAN2008_FEATURES)
```
sayerhs, line 93:
```suggestion
     add_compile_definitions(-DHAS_FORTRAN2008_FEATURES)
```

Use modern CMake

modules/nwtc-library/src/NWTC_IO.f90
sayerhs: @bjonkman Please see the suggested change I had put in after submitting this pull request.

sayerhs: Never mind, this isn't robust either.

bjonkman: @sayerhs, yes, that helps. However, if it's not an Intel compiler, we could still get into the #else branch here when `!defined(HAS_FORTRAN2008_FEATURES)`, and then `compiler_version()` wouldn't be defined.

sayerhs: Pushed another commit which I think covers the different use-cases: 1. we have Fortran2008 features, 2. We don't have F2008 but are using an Intel compiler, 3. neither... so we don't print anything regarding compiler. #454 also had an issue where it would have tried to access `compiler_version` on a non-Intel compiler which might not have had Fortran2008.

sayerhs: Resolving this conversation to allow discussion on the newer commit.

andrew-platt:
Also fixes bug reintroduced in #507 (https://github.com/OpenFAST/openfast/pull/507#pullrequestreview-453778639)

PR #534, opened 2020-08-20
Add error check in FAST_OpFM_Init

ewquon:
**Complete this sentence**

Add error check after call to `FAST_InitializeAll_T` in `FAST_OpFM_Init`. Not sure how I ended up with an input file missing the CalcSteady line, but if it can happen to me it can happen to someone else.

### DIFF
--- a/modules/openfast-library/src/FAST_Library.f90
+++ b/modules/openfast-library/src/FAST_Library.f90
@@ -497,6 +497,10 @@ subroutine FAST_OpFM_Init(iTurb, TMax, InputFileName_c, TurbID, NumSC2Ctrl, NumC
    ExternInitData%NumActForcePtsTower = NumActForcePtsTower
 
    CALL FAST_InitializeAll_T( t_initial, 1_IntKi, Turbine(iTurb), ErrStat, ErrMsg, InputFileName, ExternInitData )
+
+   IF ( ErrStat >= AbortErrLev ) THEN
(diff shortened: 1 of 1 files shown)


modules/openfast-library/src/FAST_Library.f90
```diff
@@ -497,6 +497,10 @@ subroutine FAST_OpFM_Init(iTurb, TMax, InputFileName_c, TurbID, NumSC2Ctrl, NumC
    ExternInitData%NumActForcePtsTower = NumActForcePtsTower
 
    CALL FAST_InitializeAll_T( t_initial, 1_IntKi, Turbine(iTurb), ErrStat, ErrMsg, InputFileName, ExternInitData )
+
+   IF ( ErrStat >= AbortErrLev ) THEN
+      CALL ProgAbort( "Error in FAST_OpFM_Init:FAST_InitializeAll_T" // TRIM(ErrMsg) )
+   END IF
```
bjonkman, line 504: Wouldn't it be better to return to the calling routine after line 511 instead of abruptly ending the program here? That way memory can be deallocated and files closed.

andrew-platt: Yes.  Good idea.  @ewquon updated this.

andrew-platt:
Not sure why the waves cases are failing.  @rafmudaf any thoughts?

rafmudaf:
Excellent question @andrew-platt. I've restarted the tests to see if that changes anything. They passed after my RanLux pull request and then failed when you changed something in InflowWind Driver, so that doesn't make much sense.

PR #536, opened 2020-08-25
Minor clarification in the API_Change.rst

andrew-platt:
Ready to merge.

Very minor note on the line number for the changes in the API for the airfoil table file. Since the airfoil table reader can handle an arbitrary number of header line comments, line numbers aren't very sensible. Instead we will note the new line as 4\* with the note `*non-comment line count`.

Non-commentLineCount


andrew-platt:
Thanks @kshaler for point out the ambiguity in the previous description.

PR #538, opened 2020-08-29
Updates to AeroDyn

bjonkman:
**Complete this sentence**
THIS PULL REQUEST __IS NOT__ READY TO MERGE

_Items to address before merging this code:_
- [x] need to check that the cp/cavitation code still works (is there a test for this anywhere?)
- [x] @andrew-platt and/or @ebranlard should verify that the changes to UA in the FVW module make sense. I changed some logic because the dynamic lift/drag coefficients weren't used after calculating them from UA, but perhaps that was not intended to work that way. (See the email I sent the two of you on 31-Jul-2020.)
- [x] update the documentation:
  - [x] significant changes to the AeroDyn driver
- [x] update r-test results

## Feature or improvement description
This pull request updates the structure of the AeroDyn module that was required to add new linearizable dynamic stall and dynamic inflow models. In the interest of getting the code changes that affect future pull requests in OpenFAST sooner, the two new models have been disabled until they can be tested and documented further.

### Other modifications include:
- The unsteady aero `filtCutOff` input value is now specified as a reduced frequency instead of Hz. **The default has changed from 20 Hz to 0.5.**
- Added code to turn off BEM if TSR <= 1. When TSR is below 2, it blends the BEM and non-BEM solution so that this is a gradual turning off of BEM. The `GeomPhi` output channel from the AeroDyn blade nodes indicates how much of the solution is BEM and how much is a geometric (non-BEM) solution.
- The AeroDyn driver now outputs only 1 file. It also allows a time-series input file.
- The `FrozenWake` option should work better. It may have been broken by https://github.com/OpenFAST/openfast/pull/373.
- The azimuth channel output from AeroDyn now is limited to [0, 360] degrees.
- Added a `DBEMTau1` output channel to AeroDyn, which outputs the `tau1` value used in the DBEMT module.

- Changes to NWTC Library
   - Fixed a fatal issue if that occurred when calling WrOver with more than `ConRecL`-1 characters.
   - Added a warning if the airfoil files (or any other code that uses the NWTC_Library ParseVar routines) read a variable that is wider than the field where it is stored.
- Cleaned up the Visual Studio solution files contained in the `vs-build` directory.

## Related issue, if one exists
- https://github.com/OpenFAST/openfast/issues/304

## Impacted areas of the software
<List any modules or other areas which should be impacted by this pull request. This helps to determine the verification tests.>
AeroDyn, linearization, NWTC Library

## Test results, if applicable
Cases that use AeroDyn will see some changes. Those changes should be very minor if unsteady aero is not used and the TSR is larger than 2. Cases that use unsteady aero may see some larger changes due to the slightly different filter that is implemented.


ebranlard:
That looks good to me, thanks for that. Right now the UA is not enabled with OLAF (an error is thrown in `UA_Init_Wrapper`), this is likely why we were overriding the Cl dynamic with the Cl static. 

Here's some changes that we will need to do on our side after this merge:
- add the variables `Vx_elast_dot`, `Vy_elast_dot`, `omega_z` to FVW_inputs and set them in `SetInputsForFVW` like BEM. 
- Make sure `UA_Init_Wrapper` perform the same steps as in BEM (should actually be fine). We should extract `BEMT_Set_UA_InitData` and `BEMT_CheckInitUA` so that they could be used by both BEM and FVW.
- Perform UA states integration within `UA_UpdateStateWrapper` based on several UA input times (see `BEMT.f90`)
       - Extract `SetInput_for_UA` into common code or more likely, make our own version of it
      - Make our own version of `CalculateInputsAndOtherStatesForUA`
- Check `SetOutputsFromFVW` so that it uses `FVW%m_UA`, and maybe it should call our own version of SetInput_for_UA and the dynamic coefficients are used.

ebranlard:
I don't see any issue with the cavitation check. It seems that it was not allowed with dynamic stall (AFAeroMod_BL_unsteady), I'm not sure why, but I believe it's fine to keep it that way. I can see that the code of the check (around line 1250 of AeroDyn) has not changed. Let me know if I'm missing something.

I'm not sure why some of the r-test fail. Hopefully the errors are just small numerical changes. Do you know what it could be?

bjonkman:
Thanks for checking the cavitation. I don't have that in my original code base and don't have a test for it, so it's good to have someone else take a look.

I think all of the r-tests with AD (15) are failing. I suspect most of them just have small differences because the loops in the code have been completely reordered, and some calculations are done in a slightly different order. I did see a few 5MW-model tests that had some bigger changes. I attributed those to the different UA filter. Unfortunately, we don't have a good way to visualize the failing tests here.

ebranlard:
ok, thanks a lot for the clarifications!

rafmudaf:
@bjonkman I made a few commits to your branch so that you can now inspect the results of the regression tests from GitHub. If the reg tests fail, the test files including the html-based plots are posted to the GitHub Actions workflow artifacts: https://github.com/bjonkman/openfast/actions/runs/336987286. The html files are in `glue-codes/openfast/<case-name>/<case-name>.html`.

rafmudaf:
One easy way to open all of these at once on a Mac is this bash command:
```
open `find . -name *.html`
```
I don't know the cmd prompt form, but I bet you do.

andrew-platt:
is there a change in the way the skewed wake is calculated and applied?

From my quick look at the results, I'm seeing some changes in the loads from AD15 in the 5MW models that I'm not entirely certain about.  This has the effect of changing the blade pitch a little in response.  This can be seen in the root forces in the `5MW_Land_DLL_WTurb` model.

bjonkman:
I think the differences are a result of the different unsteady aero filter that is now used. I think we need to run this with the UA filter calculated the old way and compare results. For the few models that use AD15 without UA turned on, I don't see differences.

andrew-platt:
Setting the `filtCutOff` to 20 Hz (old default) brings the 5MW models very close to the original, except for the BeamDyn case (there may be other issues triggered there).

The only other case that might be problematic is the `SWRT_YFree_VS_EDG01`.  This shows a relatively large change in the nacelle yaw angle.

Screen Shot 2020-11-02 at 4 08 02 PM

bjonkman:
@andrew-platt , I reverted the changes for the new UA filter (i.e., I made these changes: https://github.com/bjonkman/openfast/commit/3142b2fd96e24d8b6b980748e1d3c453e947ac97), and this is what I get for SWRT_YFree_VS_EDG01/SWRT_YFree_VS_EDG01.html#NacYawP
image

I'll take a look at the results over [here](https://github.com/bjonkman/openfast/runs/1349704167) when it's finished and make sure that is the same as what I get locally.

bjonkman:
Going through all of the manual regression test locally (with the exception of a couple of cases that didn't run because the python script didn't copy the input files and one that ran out of memory), I see only minor numerical differences between this branch using the old UA filter and previous results--including the BeamDyn and SWRT cases @andrew-platt mentioned. So, it is the UA filter that is causing differences.

bjonkman:
I was able to look at the artifacts from here: https://github.com/bjonkman/openfast/runs/1349704167 (the code from this PR with the old UA filter put back in).

The most noticeable difference was in the `UAE_Upwind_Rigid_WRamp_PwrCurve` case:
image

This difference is caused by the TSR dropping below 2. In the new code, BEM starts to turn off when TSR is 2 and is turned off completely when TSR is below 1. Here is a plot of the TSR from that case:
image

I also plotted TSR of another case that had some minor differences, `AWT_WSt_StartUp_HighSpShutDown`. That one also has the TSR below 2 where those differences occurred.
image
image

andrew-platt (commented):
This all looks good to me.  I confirmed that any differences in the results were due to the newer UA filter (as seen in rerunning with the old filter), and the change in the default cutoff frequency for the UA filter.

So, I recommend we rerun baselines and merge this.

jjonkman:
@1900360 -- It looks like you posted a similar question here: https://github.com/OpenFAST/openfast/issues/808; I'll respond there.

PR #547, opened 2020-09-17
[BugFix] cmake configuration for  Linux + Intel + Debug

andrew-platt:
@ewquon  discovered that the cmake setup for debug with Intel Fortran on linux is incorrect.  An incorrect flag was used for allowing recursive routine calls in the SLATEC library qk61 solvers.

This does not affect any test cases.


rafmudaf (approved):
This looks good to me. Thanks for addressing my comment.

PR #550, opened 2020-09-29
Make channel length consistent between C and Fortran sides of interface

bjonkman:
Fixes a run-time error in the Simulink interface that was introduced with increased channel-name lengths in https://github.com/OpenFAST/openfast/pull/373.

Fixes https://github.com/OpenFAST/openfast/issues/548

Simulink interface and possibly the C++ interface (depending on if it does anything with the channel names/units)

This does not change any of the current test results.


andrew-platt (approved):
Thanks for finding this!

PR #560, opened 2020-10-06
Added test for BD_diffmtc (5-node problem)

michaelasprague:
Expanded unit test for BD_diffmtc; BD_diffmtc evaluates shape functions and their derivatives at points in [-1,1]
Converted single-subroutine test to module with two test routines.
Added a 5-node test problem more closely aligned with usage. 
Also fixed call to diffmtc in original 2-node problem
Ready to be merged; all tests passed

### DIFF
--- a/modules/beamdyn/tests/test_BD_diffmtc.F90
+++ b/modules/beamdyn/tests/test_BD_diffmtc.F90
@@ -1,15 +1,12 @@
-@test
-subroutine test_BD_diffmtc()
-    ! branches to test
-    ! - 2 node, 1 element
-    
+module test_BD_diffmtc
+
     use pFUnit_mod
     use BeamDyn
     use NWTC_Num
     use test_tools
-    
+
     implicit none
-    
+
     integer                    :: n, i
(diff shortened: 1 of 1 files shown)


rafmudaf (approved):
This looks good to me.
Is Mathematica the only practical choice for a script to check the results? Octave or Python would be more convenient, not may not have a LengendreP function.

ashesh2512 (approved):

modules/beamdyn/tests/test_tools.F90
```diff
@@ -132,7 +144,27 @@ type(BD_ParameterType) function simpleParameterType() RESULT(p)
         end do
         
     end function
-    
+   
+    subroutine SimpleParameterType_TearDown(p)
```
rafmudaf, line 148: It would be more accurate to call this `ParameterType_TearDown`. The "simple" part of that is only a reference to the subroutine that creates a basic instance of this data type. By the time we are cleaning up, this may be far from a "simple" `BD_ParameterType`.

rafmudaf (changes requested):

PR #576, opened 2020-10-22
BeamDyn Curve

michaelasprague:
Work-in-progress pull request

I removed all of the cubic-spline infrastructure that was used as the initial fit to user-input Keypoint data for XYZ and twist. 

A reason for this change is that using a cubic-spline fit for initial/reference geometry calculations left an inherent discrepancy between that and the LSFE current reference line calculation.  Further, with the cubic-spline fit, a user could possibly get a bizarre LSFE fit for a given spline fit.

I replaced that cubic-spline fit with a least-squares fit with the target Legendre spectral finite elements (LSFE).  User inputs are unchanged, but the process for fitting the LSFE to the data is as follows:
- Input kp user-defined keypoints defining XYZ+twist and LSFE order p
- If p < kp, perform a least-squares fit of the target p-order LSFE
- If p > kp, let q = kp - 1 and perform a least-squares fit of q-order LSFE; once fit, use that LSFE to define the nodes of the higher-order, p-order LSFE

Routines still need to be checked/updated for multi-element usage, especially for trapezoidal-rule quadrature.

The largest changes occur in the routine InitializeNodalLocations, which was a nearly complete rewrite.  I've added a new unit test for that routine that tests a curved beam set of keypoints/twist: test_InitializeNodalLocations.F90

BeamDyn unit tests pass.
A number of regression tests fail -- namely those where the new LSFE-least squares fit gives a different interpolation of XYZ-Twist data.  For the 5MW-blade problems, I think the source of the test discrepancies is in the new interpolation of the non-uniform twist.   The changes here should not affect problems where XYZ-Twist is constant along the blade.


modules/beamdyn/src/BeamDyn.f90
```diff
@@ -3237,161 +3303,84 @@ END SUBROUTINE BD_GyroForce
...
-       DO j=1,kp_member(i)-1
-           temp_id = temp_id + 1
-           dist_to_member_start = dist_to_member_start + segment_length(temp_id)
-           segment_eta(temp_id) = dist_to_member_start/member_length(i)
-       ENDDO
-   ENDDO
-
-END SUBROUTINE BD_SegmentEta
-
+!SUBROUTINE BD_diffmtc( nodes_per_elem,GLL_nodes,QPtN,nqp,Shp,ShpDer )
```
ashesh2512, line 3306: @michaelasprague Was leaving the commented out code intentional ?

ashesh2512 (commented):
I have gone through the code changes in a broad sense, and the changes look fine. I wanted to comment on the regression test discrepancies. The pull request results in failures of the 5MW dynamic tests and the curved beam test. The failures in 

1) For the 5mw Dynamic case, the dynamic profile matches - 
Screen Shot 2020-11-18 at 1 00 26 PM

The closeup reveals difference in peaks. @michaelasprague Is it fair to interpret the differences here to be a result of the difference in resolutions of capturing the reference axis between the old and proposed approach - 
Screen Shot 2020-11-18 at 1 00 44 PM

2) For the curved beam test, I observe discrepancies too, however there is no twist in this example. Do we expect these differences as a result of the difference in capturing the reference axis ? 
Screen Shot 2020-11-18 at 1 22 50 PM

modules/beamdyn/src/BeamDyn.f90
andrew-platt: There isn't any reason to leave the commented `BD_diffmtc` in BeamDyn.f90; it has been relocated to BeamDyn_Subs.f90

andrew-platt (approved):

rafmudaf (approved):

PR #578, opened 2020-10-26
InflowWind Updates

bjonkman:
This pull request IS ready to merge.

Before merging,
- [x] check that new InflowWind features that NREL added have not been overwritten
- [x] update r-test input files and results

### InflowWind driver:
- functionality:
  - added option to convert wind files to different file formats
  - files listed in the driver input file will be relative to the file they are specified in
- input file changes:
  - added section for file conversion options
  - moved dt after timeStart (before summary stuff)
  - removed inflowWind setup section header

### InflowWind:
- functionality:
  - added base FF wind module so TurbSim, Bladed, and HAWC2 formats can use common code (all three formats use the same base code for interpolation, mean wind speed, etc)
  - added code to include vertical flow angle (to rotate FF wind boxes vertically, too), but this angle must be 0 for linearization
  - added option for native Bladed file formats (scale by standard deviation, add mean wind-speed  profile)
  - returns 0 wind speed at and below the ground height (z=0) instead of producing an error
  - added `Wind*VelXY`, `Wind*VelMag`, and `WindAngXY` output channels
- input files changes:
  - added a new line to read `VFlowAng` (upflow angle) in the InflowWind input file
  - uniform wind files now allow a 9th column to specify the the upflow angle. If this column is missing, it assumes the upflow angle is 0 degrees.

### Other:
- added Windows Visual Studio files to build InflowWind driver
- added some example input files and the start of some InflowWind documentation files (documented the new features.... it will need to be reorganized a bit when the previous InflowWind documentation is added)

None

InflowWind, Windows Visual Studio build, documentation, cmake

- [x] compare results here: 
One case fails. All of the channels that depend on turbine dynamics have a sharp increase in the error plots at 25.75 seconds, which then dies out.  The differences are still small here, though: max L2 norm is 2.0808e-04, and you really have to zoom in to see differences in the time series. The inflow channels at the hub has 0 error. Here is the channel with the max error: 
image


andrew-platt:
After merging the PR to this branch, the `5MW_TLP_DLL_WTurb_WavesIrr_WavesMulti` case will need rebaselining.  It is the only case that fails.

PR #579, opened 2020-10-27
GitHub Actions Rework

rafmudaf:
**Complete this sentence**

This is a major overhaul of the existing GitHub Actions infrastructure. It's a shift from running the full pipeline through a Bash script in Docker to leveraging the Workflows and Actions features in GitHub Actions. Some of the primary features are:

- No tests are run when a branch name begins with "docs/"
- On failing runs, the test files are posted to the run's Artifacts for further inspection
- This will work between forks whereas the existing workflow does not
- New build targets added for tests so that we do not need to compile code that isn't being tested

None

GitHub Actions and CMake targets for unit tests and regression tests

https://github.com/rafmudaf/openfast/actions


andrew-platt (approved):
This is a much cleaner approach.  From my read through, it looks like it should work well (I don't have a good way to test this, but I know @rafmudaf has done extensive testing on it).

PR #588, opened 2020-11-13
Add NWTC Library infrastructure for parsing string inputs

rafmudaf:

This pull request includes infrastructure for parsing arrays of strings as inputs. This feature will be used in various ongoing projects and eventually incorporated into most OpenFAST physics modules.

None

NWTC Library IO module

See GitHub Actions

### DIFF
--- a/modules/nwtc-library/src/NWTC_IO.f90
+++ b/modules/nwtc-library/src/NWTC_IO.f90
@@ -4445,14 +4445,25 @@ END SUBROUTINE CheckR16Var
    RETURN
    END SUBROUTINE PremEOF
 !=======================================================================
-   SUBROUTINE InitFileInfo( StringArray, FileInfo )
+   SUBROUTINE InitFileInfo( StringArray, FileInfo, ErrStat, ErrMsg )
 
-      CHARACTER(1024), DIMENSION(:), INTENT(IN) :: StringArray
-      TYPE(FileInfoType), INTENT(OUT) :: FileInfo
+      CHARACTER(*), DIMENSION(:), INTENT(IN   ) :: StringArray
+      TYPE(FileInfoType),         INTENT(  OUT) :: FileInfo
+      INTEGER(IntKi),             INTENT(  OUT) :: ErrStat
+      CHARACTER(*),               INTENT(  OUT) :: ErrMsg
+
+      CHARACTER(*), PARAMETER :: RoutineName = 'InitFileInfo'
       INTEGER :: i, n_lines
 
-      n_lines = size(StringArray)
+      ErrStat = ErrID_None
+      ErrMsg  = ""
 
+      IF (.NOT. ALLOCATED(StringArray)) THEN
+         CALL SetErrStat( ErrID_Fatal, 'Attempting to use StringArray before it is allocated.' , ErrStat, ErrMsg, RoutineName )
+         RETURN
+      END IF
+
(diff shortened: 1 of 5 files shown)


modules/nwtc-library/src/NWTC_IO.f90
```diff
@@ -4350,6 +4444,29 @@ SUBROUTINE PremEOF ( Fil , Variable, TrapErrors, ErrMsg )
...
+
+      FileInfo%NumLines = n_lines
+      FileInfo%NumFiles = 1
+
+      ALLOCATE( FileInfo%Lines(n_lines) )
+      ALLOCATE( FileInfo%FileLine(n_lines) )
+      ALLOCATE( FileInfo%FileIndx(n_lines) )
+      ALLOCATE( FileInfo%FileList(n_lines) )
+
+      FileInfo%Lines = StringArray
```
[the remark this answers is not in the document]
rafmudaf: Updated!

modules/nwtc-library/src/NWTC_IO.f90
```diff
@@ -4350,6 +4444,40 @@ SUBROUTINE PremEOF ( Fil , Variable, TrapErrors, ErrMsg )
...
+      END IF
+
+      n_lines = size(StringArray)
+      FileInfo%NumLines = n_lines
+      FileInfo%NumFiles = 1
+
+      ALLOCATE( FileInfo%Lines(n_lines) )
+      ALLOCATE( FileInfo%FileLine(n_lines) )
+      ALLOCATE( FileInfo%FileIndx(n_lines) )
+      ALLOCATE( FileInfo%FileList(n_lines) )
```
andrew-platt, line 4473: change to `ALLOCATE( FileInfo%FileList(1) )` since we only have one file.

rafmudaf: Ok good catch

modules/nwtc-library/src/NWTC_IO.f90
```diff
@@ -3320,6 +3321,99 @@ SUBROUTINE OpenUOutfile ( Un, OutFile, ErrStat, ErrMsg )
...
+!> This subroutine parses the specified line of text for AryLen CHARACTER values.
+!! Generate an error message if the value is the wrong type.
+!! Use ParseAry (nwtc_io::parseary) instead of directly calling a specific routine in the generic interface.   
+   SUBROUTINE ParseChAry ( FileInfo, LineNum, AryName, Ary, AryLen, ErrStat, ErrMsg, UnEc )
+
+         ! Arguments declarations.
+
+      INTEGER, INTENT(IN)                    :: AryLen                        !< The length of the array to parse.
+
+      CHARACTER(1024), INTENT(OUT)            :: Ary(AryLen)                  !< The array to receive the input values.
```
bjonkman, line 3367: I get a little nervous when I see the character strings passed as a defined length here: ` CHARACTER(1024), INTENT(OUT)` here. I've fixed a lot of memory issues over the years due to this kind of specification. It's much less prone to error if you use `CHARACTER(*), INTENT(OUT)` because the calling routine can specify whatever length it wants. If you want this length tied to a specific length, I'd use a parameter for it instead of 1024.

rafmudaf: Thanks for pointing this out, @bjonkman, I've made the update.

PR #589, opened 2020-11-14
ElastoDyn: Allow one-blade turbine model

ebranlard:
Ready to be merged.

Allows `NumBld=1` in ElastoDyn

ElastoDyn

The only changes needed from what I could see were about disactivating some output channels. The total number of DOF being reduced, if statements that relied on the Teeter DOF being at position 22 needed to be changed since this index is out of bounds for a 1 bladed rotor (max number of DOF being 18). 

Extending to more blades (>3) would require a rework of the outputs and the DOF indexing.


jjonkman:
Thanks, @ebranlard.  These ElastoDyn changes all make sense to me.  I'm a bit concerned that there are likely necessary changes outside of ElastoDyn that require changes as well (given that NumBl is an initialization output of ElastoDyn for use by AeroDyn, ServoDyn, and the C++ interface).  Did you check these modules for any necessary changes as well?

ebranlard:
Thanks for the review. 

I haven't seen any issue with AeroDyn, apart from hard coded number of blade inputs in the input file (3 blade files), but only then. The same is found for ServoDyn, where most of the `InputFileData` variables are hardcoded to 3 blades (`TipManS`, `PitManRat`, `BlPitchF`), but the rest of code loops through NumBl, or the min between 3 and NumBl. To be safe, I've added a bit more "array" bounds when I thought it would either be necessary, or to increase readability in ServoDyn. 

I've tested this with AeroDyn and ServoDyn for 1,2 and 3 blades, and the results seemed to make sense. I wouldn't be surprised if there is a corner case that I've missed, but we can hopefully easily spot and fix such array bound. Hopefully the r-test will pass..

jjonkman (approved):
Sounds good.  I'll approve.

andrew-platt (approved):
Looks good to me as well.

rafmudaf:
@ebranlard Do you have 1-blade test case?

ebranlard:
I don't really have a test (I mostly did this to have only one blade for the elliptical wing of OLAF, which is "inelastic". 

An easy and more or less relevant one would be to simply duplicate this one, and changing the number of blade to 1 in ElastoDyn input file (that will affect AeroDyn and ServoDyn): 
https://github.com/OpenFAST/r-test/tree/dev/glue-codes/openfast/5MW_Land_DLL_WTurb

andrew-platt:
The test case for this will be added in PR #648.

PR #590, opened 2020-11-14
VTK folder and Vx sign for OLAF

ebranlard:
Ready to merge

 - Vx sign was wrong for aero outputs per node with OLAF
-  Folder fvw_vtk was not in the right location when openfast called from a different folder, no vtk files were produced

OLAF


andrew-platt (approved):

andrew-platt:
Note for our future selves:
For the `Vx` sign, this change now aligns the equation for the FVW output with that for the BEMT (code block above it).  The signs used for `Vx` in the standard outputs match between FVW and BEMT, so this is expected here.

PR #595, opened 2020-11-16
Disable gfortran stack-reuse compiler option

andrew-platt:
This PR is ready for merging

No test cases are expected to change as gfortran 7.4 is used for testing on GitHub actions.

This PR disables the `-fstack-reuse` option for gfortran.  In some cases gfortran 9.1.0 (and possibly others) will not correctly figure out the scope of a variable within a routine and it's nested subroutine or subfunction.  This can result in the same stack location used for a variable in scope and another variable in scope that the compiler thinks is out of scope.  This is seen in c35dc09b using gfortran 9.1.0 on mac.

**Details of case c35dc09b**
The subroutine `ParseInputFileInfo` contains the function `Failed` that accesses the `ErrStat` variable in the parent routine.  The variable `CurLine` is used only in the `ParseInputFileInfo` parent routine.  However, gfortran 9.1.0 on mac with `-O3` optimzation assigns the same stack location to both variables, resulting in `ErrStat` getting overwritten by `CurLine` and OpenFAST failing with an unknown error id.  While this case can be remedied by including `CurLine=CurLine` in the contained `Failed` function, it may be an indicator that gfortran is making this kind of error elsewhere, which may be affecting the numerical stability of OpenFAST.  For safety reasons, I propose disabling this optimization option in gfortran.

Code snippet from this commit:
```
print*,'Before setting CurLine=4: ',ErrStat2,'     ---> ErrStat: ',ErrStat,'      ---> Curline: ',Curline
   CurLine = 4
print*,'After  setting CurLine=4: ',ErrStat2,'     ---> ErrStat: ',ErrStat,'      ---> Curline: ',Curline
   call ParseVar( FileInfo_In, CurLine, 'Echo', InputFileData%Echo, ErrStat2, ErrMsg2 )   ! Note: ParseVar increments CurLine
print*,'After  call to Parsevar:  ',ErrStat2,'     ---> ErrStat: ',ErrStat,'      ---> Curline: ',Curline
```

Resulting output:
```
 Before setting CurLine=4:            0      ---> ErrStat:            0       ---> Curline:            0
 After  setting CurLine=4:            0      ---> ErrStat:            4       ---> Curline:            4
 After  call to Parsevar:             0      ---> ErrStat:            5       ---> Curline:            5

 FAST_InitializeAll:SrvD_Init:

 FAST encountered an error during module initialization.
  Simulation error level: Unknown ErrID

  Aborting OpenFAST.
```

No prior issue

Seen in one isolated instance, but may be present in numerics elsewhere in the code that we have never been able to isolate.

I cannot find any known bug reports in the `gcc` bug reports.

### DIFF
--- a/cmake/OpenfastFortranOptions.cmake
+++ b/cmake/OpenfastFortranOptions.cmake
@@ -107,6 +107,10 @@ macro(set_fast_gfortran)
   #set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -ffree-line-length-none -cpp -fopenmp")
   set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -ffree-line-length-none -cpp")
 
+  # Disable stack reuse within routines: issues seen with gfortran 9.x, but others may also exhibit
+  #   see section 3.16 of https://gcc.gnu.org/onlinedocs/gcc-9.2.0/gcc.pdf
+  set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -fstack-reuse='none'")
+
   # Deal with Double/Single precision
   if (DOUBLE_PRECISION)
     add_definitions(-DOPENFAST_DOUBLE_PRECISION)
--- a/cmake/OpenfastFortranOptions.cmake
+++ b/cmake/OpenfastFortranOptions.cmake
@@ -109,6 +109,7 @@ macro(set_fast_gfortran)
 
   # Disable stack reuse within routines: issues seen with gfortran 9.x, but others may also exhibit
   #   see section 3.16 of https://gcc.gnu.org/onlinedocs/gcc-9.2.0/gcc.pdf
+  #   and https://github.com/OpenFAST/openfast/pull/595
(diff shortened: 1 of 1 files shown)


andrew-platt:
I ran a couple of tests to check what impact this had on memory usage with the following glue code regression tests:

- `5MW_Land_DLL_WTurb` case:                  0.05% increase in maximum memory usage
- `5MW_OC4Semi_WSt_WavesWN` case:     3.0%.  increase in maximum memory usage

ebranlard:
I wonder if we could remove the single quotes in (`-fstack-reuse='none'` ).
The single quotes leads to compilation error on windows. (I'm using MSYS2 / msys64to get gfortran 10.2 and blas). 
Removing the quotes makes things work fine. 

The main issue I've noticed is that due to this extra flag, my blas libraries are not found (but there is no error message related to the flag at this stage) and I can't get anywhere in the compilation. It's only if I somehow hack the cmake files with my blas libraries, that then cmake goes through and I can start the compilation. Then I can see that it's the fortran flag that is problematic.  So I wonder if we should set the Fortran flags *after* looking for blas/lapack/mkl. 

Anyway, if you are fine with removing the quotes, that'd be great!

andrew-platt:
A quick check with gfortran on the mac indicates removing the quote should be fine.

PR #597, opened 2020-11-19
AD/AA: Add new TE definition, improve airfoil thickness calculation, simplify input

ptrbortolotti:
The PR should be ready to be merged
The docs are updated
The appropriate r-tests have been modified and a PR is open 
https://github.com/OpenFAST/r-test/pull/36

This PR impacts the aeroacoustic module.
Emmanuel B. and I changed the way the trailing edge geometry is read in and we got rid of one input file. This simplifies the inputs and the code and reduces the risk of errors. I also recoded the way the code computes the airfoil thickness at 1% and at 10% chord, which was very prone to bugs as it only supported airfoil geometries defined counterclockwise. The new implementation supports both clockwise and counterclockwise defined airfoils.

None

Aeroacoustics only


ebranlard (approved):

ebranlard:
It all makes sense to me, good job also adding to the documentation!

andrew-platt:
See https://github.com/OpenFAST/r-test/pull/36 for r-test updates.

rafmudaf:
@andrew-platt @ptrbortolotti @ebranlard  The r-test commit is correctly updated here and the AeroAcoustics test case is failing. Is this expected? The regression test results files can be inspected here: https://github.com/OpenFAST/openfast/actions/runs/375312343.

ebranlard:
I had a look at the differences for the failing test case and they appear to be small enough and justified given the changes that Pietro made.

ptrbortolotti:
I am not so sure where to check the differences, but I did recode the way the code computes the leading edge shape for the Simplified Guidati model, and I trust the new code much more, so this should be good to go

rafmudaf:
This currently hangs when compiled and run on Windows with Intel compiler.

ptrbortolotti:
and are we sure that this happens because of my changes? @ebranlard , I may need your help here. I don't see anything suspicious, but I'm definitely not the fortran guru who can state something like this

rafmudaf:
No, I'm not sure its due to your changes. I plan to test and evaluate this further soon, but I did want to mention what is the current hold up on merging this.

ebranlard:
I was able to compile Pietro's branch with Release x64 on windows and I could run Pietro's test case.
Maybe it's an issue with your virtual machine?

ptrbortolotti:
@rafmudaf @ebranlard , I've just figured that there is another problem with the code (specifically with the way the turbulence intensity is passed to the Amiet model). me fixing the new issue won't solve any compiling errors, but maybe we can wait to merge this before I've figured how to fix the other problem...

PR #602, opened 2020-11-24
Vectorize a section of VariousWaves_Init

rafmudaf:
This pull request improves the vectorization potential of a set of initialization lines in the Waves submodule of HydroDyn. This is relatively small performance improvement as it saves only about 3 seconds of initialization time in the offshore 5MW cases. Its also a fixed-cost improvement since it is in the init; that is, the speed up is similar whether the simulation duration is 60 seconds or 1 second. In any case, faster is faster :)

I've also added commented flags for getting profiling and performance information from the Intel compiler in the CMake configuration. See the attached optimization reports for details regarding the vectorization improvements by the compiler due to this change. These are best viewed as a side-by-side diff.

One question is whether the bounds of the loop are meaningful in the current version. The outer loop goes from 0 to `InitOut%NStepWave-1`, but all of the variables accessed in the loop are allocated to `0:InitOut%NStepWave` in their first dimension. Is the last index of the first dimension required to be empty or uninitialized?

None

Waves submodule of HydroDyn: `VariousWave_Init` subroutine

This was funded by Intel through their designation of NREL as an Intel Parallel Computing Center (IPCC). The objective was to improve the vectorization of the offshore module by using the Intel compiler and related tools. We wanted to get as much speed up as possible without making significant changed to the OpenFAST source code.

[current_Waves.f90.optrpt.log](https://github.com/OpenFAST/openfast/files/5591450/current_Waves.f90.optrpt.log)
[incoming_Waves.f90.optrpt.log](https://github.com/OpenFAST/openfast/files/5591537/incoming_Waves.f90.optrpt.log)

### DIFF
--- a/cmake/OpenfastFortranOptions.cmake
+++ b/cmake/OpenfastFortranOptions.cmake
@@ -173,6 +173,22 @@ macro(set_fast_intel_fortran_posix)
   endif()
 
   check_f2008_features()
+
+  ### Intel profiling flags
+
+  # https://software.intel.com/content/www/us/en/develop/documentation/fortran-compiler-developer-guide-and-reference/top/compiler-reference/compiler-options/compiler-option-details/optimization-report-options/qopt-report-qopt-report.html
+  # phases: vec, par, openmp
+  # set(CMAKE_Fortran_FLAGS_RELWITHDEBINFO "${CMAKE_Fortran_FLAGS_RELWITHDEBINFO} -qopt-report-phase=vec,openmp -qopt-report=5")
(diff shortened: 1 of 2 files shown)


andrew-platt:
The last point in the array can be initialized to zero.  It gets set to value of the first point in the array [here](https://github.com/OpenFAST/openfast/pull/602/files#diff-e0a70aaaab90986316064d234fe6a6272f968935bfb932235407fc92fe53ad7eR2005).

andrew-platt:
Nice find!  This looks good to me!

PR #606, opened 2020-12-02
[BugFix] array bounds issue introduced in PR #602

andrew-platt:
There was an error in the array bounds with PR #602 that would cause a runtime error when compiled in debug mode.

No test cases are affected.

### DIFF
--- a/modules/hydrodyn/src/Waves.f90
+++ b/modules/hydrodyn/src/Waves.f90
@@ -1939,13 +1939,13 @@ SUBROUTINE VariousWaves_Init ( InitInp, InitOut, ErrStat, ErrMsg )
          ELSE
             ! The elevation of the point defined by WaveKinzi(J) must lie between the seabed and the mean sea level (inclusive)
 
-            InitOut%WaveDynP(:,J  ) = WaveDynP0B(:,K)
-            InitOut%WaveVel (:,J,1) = WaveVel0Hxi(:,K)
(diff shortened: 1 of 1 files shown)


modules/hydrodyn/src/Waves.f90
```diff
@@ -1939,13 +1939,13 @@ CASE ( 1, 10 )          ! Plane progressive (regular) wave; the wave spectrum is
...
             ! The elevation of the point defined by WaveKinzi(J) must lie between the seabed and the mean sea level (inclusive)
 
-            InitOut%WaveDynP(:,J  ) = WaveDynP0B(:,K)
-            InitOut%WaveVel (:,J,1) = WaveVel0Hxi(:,K)
-            InitOut%WaveVel (:,J,2) = WaveVel0Hyi(:,K)
-            InitOut%WaveVel (:,J,3) = WaveVel0V(:,K)
-            InitOut%WaveAcc (:,J,1) = WaveAcc0Hxi(:,K)
-            InitOut%WaveAcc (:,J,2) = WaveAcc0Hyi(:,K)
-            InitOut%WaveAcc (:,J,3) = WaveAcc0V(:,K)
+            InitOut%WaveDynP(0:InitOut%NStepWave-1,J  ) = WaveDynP0B( 0:InitOut%NStepWave-1,K)
```
andrew-platt, line 1942: The first index on the arrays on the left runs from `0:InitOut%NStepWave` instead of `0:InitOut%NStepWave-1` like the right

PR #610, opened 2020-12-10
Infrastructure Improvements

rafmudaf:
This pull request does a few minor things:
- Adds some resolution for when to run the GH Actions workflow (i.e. don't compile and run tests when the only thing that changed is the documentation)
- Removes unused CI configuration files ([au revoir, Travis](https://www.jeffgeerling.com/blog/2020/travis-cis-new-pricing-plan-threw-wrench-my-open-source-works))
- Prevents variable tracking for large Fortran modules

Testing and CMake

One question is whether disabling variable tracking should only happen in Release mode. @andrew-platt have any thoughts? Relevant documentation for GNU Fortran is [here](https://gcc.gnu.org/onlinedocs/gcc/Debugging-Options.html). My thought is that if `-g` is on, we probably want this enabled.


glue-codes/openfast/CMakeLists.txt
```diff
@@ -14,8 +14,8 @@
 # limitations under the License.
 #
 
-add_executable(openfast
-  src/FAST_Prog.f90)
+add_executable(openfast src/FAST_Prog.f90)
+set_source_files_properties(src/FAST_Prog.f90 PROPERTIES COMPILE_FLAGS "-fno-var-tracking -fno-var-tracking-assignments")
```
sayerhs, line 18: heh... I was just griping about this on eagle. Great addition!

andrew-platt (commented):
Are the compile flags `"-fno-var-tracking -fno-var-tracking-assignments"` specific to the gfortran compiler?  Will this be an issue with windows + ifort?

glue-codes/openfast/CMakeLists.txt
rafmudaf: Yes, I've also added this to both of those modules.

rafmudaf: @sayerhs do you have any input on whether this should always be disabled or if only for Release mode (`-O3`)? I think I see this warning with `-O2` as well, but RelWithDebInfo includes `-g` so I think we may want variable tracking there.

andrew-platt:
I think we should only add the no-var-tracking when not using the `-g` option.  Probably safe to do for `-O2` and higher.

rafmudaf:
Ok I've added a check for GNU compiler and CMAKE_BUILD_TYPE=Release mode.

glue-codes/openfast/CMakeLists.txt
sayerhs: @rafmudaf My vote would be to **not** add `-fno-var-tracking...` flags for `Debug` build, but use it for `Release` and `RelWithDebInfo`. With `-O2 -g` a lot of variables are removed anyway, so I am not sure it is going to affect much.

rafmudaf:
I've left variable tracking enabled for RelWithDebInfo for now just to be conservative. We can easily add that to the if-statement, if needed.

PR #537, opened 2020-08-26
Flexible floating platforms

HaymanConsulting:
**Complete this sentence**
THIS PULL REQUEST IS READY TO MERGE!!!

**Feature or improvement description**
This pull request adds the ability to model flexible floating platforms with multiple potential flow bodies and flexible connections.  A significant rewrite of the HydroDyn Morison module and large modifications to SubDyn was required.

- Major revisions to SubDyn
    - New input file format
    - Guyan damping
    - Guyan load correction
    - Raleigh damping
    - Cables
    - Rigid links
- Major revisions to HydroDyn
    - New method for calculating buoyancy
    - multiple bodies for potential flow methods
    - revision of internal data handling for potential flow models (WAMIT)
    - significant rewrite of the Morison module
    - new input file format
    - new driver level test cases
- Glue code changes
    - coupling of SubDyn to HydroDyn for flexible floating platforms

**Impacted areas of the software**
HydroDyn
   - calculation methods
   - input file format
SubDyn
   - calculation methods
   - input file format
Glue-code
   - coupling between SD and HD

**Test results, if applicable**
r-test baselines need updating pending final verification (current branch: **f/TCF-mods-Dev**)
- [x] Extract new input motion time series for the `modules/hydrodyn/hd_5MW*` floating cases using the comparable fully coupled OpenFAST regression test results.
- baseline rerun:
   - [ ] linux-gnu
   - [ ] linux-intel
   - [x] macos-gnu
   - [ ] windows-intel
- [x] update the NBodyMod example cases -- these are not run in testing, but are examples for format
- [x] add the subdyn driver level test cases

**To Do**
- [x] Update the `api_change.rst` with changes to the HD input file
- [x] Whats failing with the regression tests?
   - All regression tests now run (2020.12.21).  
   - The `hd_OC3tripod_offshore_fixedbottom_wavesirr` case fails with
       ```
        ErrMsg returned: HydroDyn_CalcOutput:Morison_CalcOutput:The lowest element of a Morison member
    has become partially submerged!  This is not allowed.  Please review your model and create a
    discretization such that even with displacements, the lowest element of a member does not
    become partially submerged.```
    This case was revised so this no longer is an issue (heave was set to -3m which didn't make any sense for a fixed bottom case).
   - The channel names from HydroDyn changed, so comparisons should be made only with common named channels (testing cannot currently generate plots automagically when names are different)
   - The buoyancy calculations have changed considerably from before, leading to very small differences in floating cases.
[...]

PR #615, opened 2020-12-18
HD/Morison performance improvement, NWTC Lib and BD bug fixes

bjonkman:
This pull request increases the efficiency of HydroDyn's Morison CalcOutput routine (based on profiling results), fixes a few minor debugging issues that seem to be related to the latest Intel compiler suite I installed, and updates some minor issues I found in NWTC Library. 

none

HydroDyn, NWTC_Library, BeamDyn

**The details:**

- **HydroDyn**
  - HydroDyn's Morison CalcOutput routine was updated so that the interpolation routiness it uses are more efficient. The waveTime (X) array is checked only one time, and it uses the calculated slope for all of the corresponding (Y) arrays (in nested loops)  In my tests, this change resulted in a little more than 1% speed improvement.
  - I also updated some HydroDyn code so that my newest Intel debugger does not complain about uninitialized variables or passing unallocated arrays.
  -  I removed a (non-standard) tab from the Waves.f90 source file.

- **NWTC_Library**
  - I regenerated the parts of the NWTC_Library that are generated by the registry to ensure consistency between the Registry include files and the code.
  - I added an `#else` statement in `DispCompileRuntimeInfo` to prevent any chance of using uninitialized variables there.
  - I added a `NWTC_SizeOfNumWord` parameter to use in place of hard-coded values in the input-file parsing routines, for consistency between Envision and NREL code bases.
  - I fixed a bug in `ReadFASTbin`, where trying to read from uncompressed binary files would have resulted in garbage or a seg fault.
  - I changed the error number if the `LoadDynamicLibProc` routines ever fail, to allow calling routines to determine which procedure failed to load (this is potentially used in ServoDyn with loading non-legacy DLLs)

- **BeamDyn**
  - The error status variables weren't initialized in the `BD_Interp_Pos_CRV` routine. If the compiler initialized the error status variable to a large random number, the code would fail.

These changes do not affect the regression tests: https://github.com/bjonkman/openfast/actions/runs/413886628


modules/nwtc-library/src/SysIFL.f90
```diff
@@ -468,7 +468,7 @@ FUNCTION dlSym(handle,name) BIND(C,NAME="dlsym")
          DLL%ProcAddr(i) = dlSym( DLL%FileAddrX, TRIM(DLL%ProcName(i))//C_NULL_CHAR )  !the "C_NULL_CHAR" converts the Fortran string to a C-type string (i.e., adds //CHAR(0) to the end)
 
          IF(.NOT. C_ASSOCIATED(DLL%ProcAddr(i))) THEN
-            ErrStat = ErrID_Fatal
+            ErrStat = ErrID_Fatal + i - 1
```
andrew-platt, line 471: This is an interesting statement.  Does this mean that `ErrStat` can be larger than `ErrID_Fatal` since `NWTC_MAX_DLL_PROC = 3` in `NWTC_Base.f90`?  I guess that's useful for figuring out which ProcAddr failed.

bjonkman: Yes. With the non-legacy DLL, the user is allowed to have an init routine, but it is not necessary. So, if the procedure doesn't exist, there needs to be some logic to avoid calling it and to call the normal controller routine instead. See https://github.com/OpenFAST/openfast/blob/dev/modules/servodyn/src/BladedInterface.f90#L401. I forgot to give you this code when I updated ServoDyn last time. Oops.

andrew-platt: Thanks for the clarification.  That makes sense.

andrew-platt (approved):
This might cause conflicts with #537, but I would like to merge this before then.

It all looks good to me.

PR #627, opened 2021-01-11
[BugFix] AD15: Fix logic for setting size of arrays for airfoil tables

andrew-platt:
This is ready for merging

If multiple input tables exist within an airfoil input file, the logic for setting the size of the Coefs array would result in extra columns for all tables > 1.  This caused segfaults within the CubicSplineInitM routine with certain compilers (gfortran 9).

This problem was discovered by @ptrbortolotti 

This fixes a segmentation fault that could occur when multiple airfoil tables were included in a single airfoil input file.

Only the AirFoilInfo submodule of AeroDyn15 is affected.

No test cases change with this request.

### DIFF
--- a/modules/aerodyn/src/AirfoilInfo.f90
+++ b/modules/aerodyn/src/AirfoilInfo.f90
@@ -257,6 +257,16 @@ CONTAINS
             return
          end if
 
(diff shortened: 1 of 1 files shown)


ptrbortolotti:
thank you Andy, that was a really quick fix!

rafmudaf (approved):

Issue #620, opened 2020-12-30
Runtime error when linearizing with dev branch


dzalkind:
**Bug description**
I get either a segmentation fault or runtime error (depending on compilation) when trying to linearize on the dev branch with 

```
---------------------- FEATURE SWITCHES AND FLAGS ------------------------------
1                      CompElast   - Compute structural dynamics (switch) {1=ElastoDyn; 2=ElastoDyn + BeamDyn for blades}
1                      CompInflow  - Compute inflow wind velocities (switch) {0=still air; 1=InflowWind; 2=external from OpenFOAM}
2                      CompAero    - Compute aerodynamic loads (switch) {0=None; 1=AeroDyn v14; 2=AeroDyn v15}
1                      CompServo   - Compute control and electrical-drive dynamics (switch) {0=None; 1=ServoDyn}
0                      CompHydro   - Compute hydrodynamic loads (switch) {0=None; 1=HydroDyn}
0                      CompSub     - Compute sub-structural dynamics (switch) {0=None; 1=SubDyn; 2=External Platform MCKF}
0                      CompMooring - Compute mooring system (switch) {0=None; 1=MAP++; 2=FEAMooring; 3=MoorDyn; 4=OrcaFlex}
0                      CompIce     - Compute ice loads (switch) {0=None; 1=IceFloe; 2=IceDyn}
```

and only the `GenDOF` is enabled in ElastoDyn.

**To Reproduce**
<Update the following list with your specific information.>
1. Compile dev branch with gfortran
2. Run lin_0.fst case in [iea_semi_lin.zip](https://github.com/OpenFAST/openfast/files/5755902/iea_semi_lin.zip)

**Expected behavior**
When running this same case (and deleting a line in lin_0_InflowWind.dat) on the master branch, the linearization completes as expected.

**Screenshots, if applicable**
<Add screenshots to help explain your problem.>
![image](https://user-images.githubusercontent.com/65573423/103384320-3c86ca00-4ab3-11eb-85cb-62887e79859a.png)

**OpenFAST Version**
<Please provide as much detail as possible including git commit. The best information is a screenshot of the OpenFAST system description that prints when running OpenFAST:>

```

 **************************************************************************************************
 OpenFAST

 Copyright (C) 2020 National Renewable Energy Laboratory
 Copyright (C) 2020 Envision Energy USA LTD

 This program is licensed under Apache License Version 2.0 and comes with ABSOLUTELY NO WARRANTY.
[...]

PR #634, opened 2021-01-14
Fix HydroDyn linearization matrix multiplication using unallocated arrays

bjonkman:
HydroDyn was calling linearization routines for SS_Radiation and SS_Excitation without checking if those modules were ever initialized. This resulted in doing some matrix multiplications with unallocated matrices.

#620

Linearization with the HydroDyn module

The Windows Intel compiler seems to return an empty matrix, but some other compilers give a seg fault.

I haven't tested this, so someone who is more familiar with HydroDyn should review before merging.

This should not change any results.


andrew-platt (approved):
This looks correct to me.  Thanks for fixing this!

PR #607, opened 2020-12-07
ServoDyn Structural control submodule (formerly TMD)

andrew-platt:
This pull request is ready to merge

This update includes significant contributions from the following people at the Univeristy of Massachusetts in Amherst.

- Matt Lackner (@matthewlackner, lackner@ecs.umass.edu, [https://mie.umass.edu/faculty/matt-lackner](https://mie.umass.edu/faculty/matt-lackner) )
- William LaCava
- Meghan Glade
- Semyung Park

**Feature or improvement description**
The TMD module within ServoDyn has been updated and renamed to Structural Control (StC).  This module is used to simulate tuned mass dampers attached to a wind turbine. This includes the following updates:

- Multiple StC elements at Nacelle, Tower, Blades, and/or floating platform (each StC element is a unique instance of the StC module).
  - Specified within the ServoDyn input file.
  - Blade StC is identical on each blade (one set of blade StC's at applied per StC module instance)
- StC options:
  - Vertical DOF for an StC element
  - Tuned Liquid Column Damper (TLCD) option
  - Force time series input at an StC element
- Updated ServoDyn file parsing and input file passing

**Impacted areas of the software**

- ServoDyn input file format
-Modeling of tuned mass dampers

**To complete before PR**

- [x] Update ServoDyn input files in r-test
- [x] User documentation on StC
- [x] User documentation on ServoDyn

**Test results, if applicable**
Currently no test cases exist for the StC module.

PR #637, opened 2021-01-15
Use default runner and GFortran 10

rafmudaf:
This pull request reconfigures GitHub Actions to use the default GitHub-hosted runner (Ubuntu 20.04) and drops the need to use Docker images with OpenFAST preinstalled. This pull request also updates the GNU baselines with GFortran 10.

None

This primarily impacts the test suite. However, other pull requests will benefit from the improved build time from upgrading the GFortran version. The Docker-dependent pipeline also had a memory limitation when compiling the FAST Library, so that may be fixed here.

The tests pass, but more interesting is the time to completion improvement between this configuration and the previous one: [1h 32m 43s](https://github.com/rafmudaf/openfast/actions/runs/488381720) vs [2h 8m 37s](https://github.com/OpenFAST/openfast/actions/runs/486587421).


andrew-platt:
Thanks! This should also now allow regression tests PR #607 to run in GH actions (memory issues with gcc-7.4).

PR #643, opened 2021-01-20
AD: Support primary input parsing through file and strings

andrew-platt:
This pull request is ready for merging.

This PR changes the method that the AD15 primary input file is handled.  The file is now read into a data structure using the `ProcessComFile`, then parsed using the `ParseVar` commands.  This also allows direct passing of the AD15 primary input file data structure to avoid IO.

Passing of other files in AD15 is not yet enabled.  So those files (FVW, blade files, airfoil tables, etc) are still read.  Similar changes would be required to enable passing those files.  This PR is being submitted now with only the passing of the primary AD15 file so as to not get too far out of sync with the rest of the code base.  The other passing of other files may be added at a later date.

This only affects the handling of the AD15 primary input file.  No other code should be affected, and no test cases should be affected by this change.

This is part of a larger effort of the Integration task within the ATLANTIS-WEIS program.

**Testing**
No additional testing is added with this PR.


ebranlard (approved):
This looks good to me. 320 lines of code less to parse an input file, and the possibility to read to read it from the fileinfo, with better error handling, that's definitely worth it! 

One thing I would consider (for this pull request or another), is that when an input is missing,  provide a link to the user as to where to find the latest format. Maybe the dev branch of  readthedocs api_change (but that might not be easy to read), or the dev-branch of r-test (not ideal neither, but would definitely be up to date). Up to you!

andrew-platt:
Good point.  It can be really hard for an outside user to know where the latest input file is located.  I'm not sure where we would want to point them to.  I wonder if we could get the git hash for the r-test the build was linked to, and maybe a generic location (r-test/dev?) for when it gets compiled without git (we do something similar for the git hash that appears at the startup of OpenFAST).

rafmudaf (approved):
Looks great @andrew-platt good work!

PR #647, opened 2021-01-25
Add ability to turn unsteady aero back on during a simulation

bjonkman:
This pull request allows unsteady aero (UA) to turn back on after it has been turned off. Previously, when the angle of attack exceeded a certain threshold, UA was turned off for that node for the rest of the simulation. The new algorithm blends the steady solution with the unsteady solution; it uses:
- only the **steady** solution for all angles of attack above the `UACutout` value listed in the airfoil file
- only the **unsteady** solution for all angles of attack 5 degrees below `UACutout`
- a (cosine) blend of the steady and unsteady solutions within 5 degrees of `UACutout`

Some other changes included in this PR:
- I updated the UA driver code to output a few more channels and write to the output file inside the UA module instead of separately in BEM, FVW, and UA driver. This means that the `UA_OUTS` preprocessor directive is used only in UA now.
- I tweaked the fix for https://github.com/OpenFAST/openfast/pull/627 so that if there are multiple tables and UA isn't used, there won't be an error if some of the airfoil tables are missing UA data.
- I fixed some minor things to avoid compiler warnings:
  - a missing comma in a TurbSim format statement
  - a 0-based array index in HydroDyn (in unused code)
  - `R8Ki` vs `DbKi` in MAP++ for the double-precision C-Fortran interface
 

none

AeroDyn

```
The following tests FAILED:
	  3 - AWT_YFree_WSt (Failed)
	  7 - AOC_YFree_WTurb (Failed)
	  8 - AOC_YFix_WSt (Failed)
	 12 - WP_VSP_ECD (Failed)
	 13 - WP_VSP_WTurb (Failed)
	 14 - SWRT_YFree_VS_EDG01 (Failed)
```
The failing cases 3, 7, 8, 12, and 14 previously had `Warning: Turning off Unsteady Aerodynamics due to high angle of attack` in their log files (i.e., printed to the console file). After rerunning those cases with additional aerodynamic outputs, I checked that the changes in output were indeed related to the outputs of unsteady aero.

Case 13 had a step or two where the angle of attack exceeded `UAcutout - 5 degrees`, thus slightly changing results for that case. You have to zoom in quite a bit to see differences (if we output at every time step, it might be more noticeable)

Some example plots:
Case 3: AWT
image

Case 7: AOC
image

Case 8: AOC
image

Case 12: WP ECD: This case looks the worst, especially in the acceleration outputs. However, I believe that is a result of the UA model itself, and not the new code. One option would be using a lower `UAcutout` value ; we have also seen that `UAMod=2` seems to give better results in some of these extreme wind cases.
image
image

Case 13: WP
image

Case 14: SWRT
image


jjonkman (approved):
I've discussed this enhancement with Bonnie and I'm generally happy with the approach.  There are some new spikes that show up in these test results, but that is likely because of how the models are set up (likely not very realistically).  The tests could be changed to eliminate these spikes, e.g,. by switching UAMod, changing the UA parameters, or lowering the UACutOut.

ebranlard (approved):
It looks good to me. It'd be nice to merge this soon so that we can push a vortex code release (that will need some merging with this).

PR #648, opened 2021-01-26
OLAF improvements for WEIS

ebranlard:
This pull request is not ready to be merged. 

The following features are included in this pull request:
 - Dynamic stall into OLAF
 - Core radius as a continuous state
 - Additional core radius options (based on chord, or section length)
 - Runge-Kutta 4 time integration scheme
 - Over-cycling improved and supported by Euler1 and RK4
 - Preliminary guidelines documented
 - Possibility to export velocity field on grid/planes/lines/points
 - Placement of vorticity and adjusting core size for tip and root vortex 
 - Bug fixes for exports

The features have been documented in OLAF's documentation. The changes to the input file (gridOutputs) are optional. 

**Related**
Requires #647 and #589

AeroDyn/OLAF

Changes in EllipticalWing test case are negligible and down to numerics. 

**Checklist**
- [x] merge #647
- [x] merge #589
- [x] additional features
- [x] Test case for ElastoDyn 1 blade
- [x] review


modules/aerodyn/src/AirfoilInfo.f90
```diff
@@ -683,7 +657,7 @@ SUBROUTINE ReadAFfile ( AFfile, NumCoefs, InCol_Alfa, InCol_Cl, InCol_Cd, InCol_
                CALL ParseVarWDefault ( FileInfo, CurLine, 'filtCutOff', p%Table(iTable)%UA_BL%filtCutOff, 0.5_ReKi, ErrStat2, ErrMsg2, UnEc )
                   CALL SetErrStat( ErrStat2, ErrMsg2, ErrStat, ErrMsg, RoutineName )
                   
-               p%ColUAf    = NumCoefsTab + 1 ! column for f_st
+               p%ColUAf    = NumCoefsIn + 1 ! column for f_st
```
andrew-platt, line 660: Is this a regression?  We ran into issues with multiple tables and corrected it with PR #627, but this PR seems to be changing it back to the problematic code.

andrew-platt (commented):
This PR may be undoing PR #627.

modules/aerodyn/src/AirfoilInfo.f90
bjonkman: @andrew-platt , this is code from my recent UA pull request #647. I don’t think this is actually changing it back to problematic code. It is just modifying it in a different way. Note that the number of coefficients passed as an argument is no longer intent(out).

andrew-platt: Thanks for the clarification. I'll look at the logic more closely.

Do you know if this code can handle one of the airfoil files containing more tables than the other files?  What I had found in #627 was that the numcoefs wouldn't get reset between the reading of the tables and the array would get overstepped (perhaps my solution actually introduced other issues).

bjonkman: Yes. In the AD input file, you must specify which columns are in (all) the airfoil tables. Each table must include at least those numbers of columns. Those inputs determine the number of columns being read from the file... you could theoretically have more columns in the tables, but they wouldn't be used. When adding the new HGM unsteady aero model, we need to calculate 2 more columns to be interpolated with alpha, but I messed up the logic if there were multiple tables in each file. So now, the number of columns as determined from the AD file (`NumCoefsIn`) is fixed, but then each table in each airfoil file has the number of columns calculated independently (`NumCoefsTab` is either (1) the number from AD or (2) the number from AD + 2). The 3D spline coefficients for each table are now allocated based on the size of the coefficients array instead of the `NumCoefs` variable, so you don't have to do the check that the arrays are the same size later. 

With this method, you do not have to specify UA coefficients for each table unless you are actually using a UA model. (e.g., you could specify UA coefficients for the first table but not the second and it would still run for cases where UA was not used.) I didn't want to have AFI checking UA stuff any more than it needs to, and it's a little clearer to me what that NumCoefs in the calling routine is actually doing if we don't modify it in this subroutine.

modules/aerodyn/src/FVW_VortexTools.f90
```diff
@@ -51,6 +52,127 @@ module FVW_VortexTools
...
+         iTip =n+1
+         iRoot=1
+         EpsTip   = Eps(1,iTip-1)
+         EpsRoot  = Eps(1,iRoot)
+      endif
+   endsubroutine PlaceTipRoot
+
+
+   !> Flatten/ravel a 3D grid of vectors (each of size n)
+   subroutine FlattenValues(GridValues, FlatValues, iHeadP)
```
andrew-platt, line 145: Wouldn't it be possible to use the Fortran intrinsic `reshape` to do what `FlattenValues` and `DeflateValues` do?

```
FlatValues(:,:) = reshape( GridValues, (/ size(gridvalues,1), size(gridvalues,4)*size(gridvalues,3)*size(gridvalues,2) /) )
```
(or something along those lines -- I don't remember the exact syntax)

andrew-platt (approved):
I didn't check through the full logic of the over cycling, but I think I have a rough understanding of how it works.

I also noticed a few `print*` and `save` variables, but as far as I can tell, those are all debug related so they shouldn't pose issues in normal use.

andrew-platt:
We may need to merge r-test/dev into the r-test/f/weis_olaf branch to clear the conflicts there.

modules/aerodyn/src/FVW_VortexTools.f90
ebranlard: From my teaching and experience I've avoided using intrinsic functions for large datasets, some of them are using the stack and fail without reporting errors. I've been told these simple loops are easy for the compiler to optimize. But 
in that case you might be right, and I don't know which one would perform best!!! This function has a little "bonus" that allows me to flatten several grids into one array by calling this function several times for different arrays, changing iHeadP. I'm using this to pack all the control points from all the user-grids into one array that is then sent to the induced velocity function.

andrew-platt:
Oops.  I didn't update the elliptical wing results before merging.  :(

I'll setup a new PR for that update.

ebranlard:
I should have updated them in:
    https://github.com/OpenFAST/r-test/tree/f/weis_olaf
For linux..

andrew-platt:
I see the updates for the linux-gnu results there, but not for the other 3. Do you have some updates for the intel ones that didn't get pushed?  I can handle the macos-gnu one.

andrew-platt:
Issue #742 may be related to the missing updates for the Windows baselines for the EllipticalWing regression test.

PR #666, opened 2021-02-24
Fix tower force nodes on restart

gantech:

Fix actuator force nodes on the tower on a restart if TwrAero is set to False

OpenFAST C++ API for actuator type applications

This only affects the C++ API and since this doesn't involve any regression tests, none of them are affected.

Updated by @rafmudaf 4/13/22:
I've rebased this commit onto the current dev branch and force-pushed. For reference, the original commit hash was 1ba33f4a70d51df11f004f0b18e0d289ab0459d6 on @gantech's fork.

### DIFF
--- a/glue-codes/openfast-cpp/src/OpenFAST.cpp
+++ b/glue-codes/openfast-cpp/src/OpenFAST.cpp
@@ -7,7 +7,7 @@
 
 int fast::OpenFAST::AbortErrLev = ErrID_Fatal; // abort error level; compare with NWTC Library
 
-//Constructor 
+//Constructor
 fast::fastInputs::fastInputs():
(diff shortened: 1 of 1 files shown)


sayerhs (approved):
@gantech Probably should add this for `fast::trueRestart` also?

rafmudaf:
@gantech I've recently been working on incorporating a regression test for the C++ interface based on the work you started a few years ago. See it [here](https://github.com/rafmudaf/openfast/tree/cpp_api_test). This is a good opportunity to expand the tests to catch this sort of bug, so let's work together to expose this bug through a test.

gantech (commented):
Looks good.

PR #670, opened 2021-02-26
Enable regression tests for the C++ API

rafmudaf:
This pull request add the testing infrastructure to create regression tests for the C++ API. It adds a single test that runs an existing OpenFAST regression test through the C++ API with the C++ driver code and compares the results to the baselines in the OpenFAST r-test repository.

This pull request also adds consistent formatting to the C++ code. The changes to `OpenFAST.cpp` are only formatting changes. 

C++ API, C++ driver code, regression test configuration.

The included test is a simple test to ensure the C++ API compiles and runs in a basic state. The included infrastructure should be used to generate new tests for expanded testing.

Primarily, the next step is to establish a test for #666.


gantech (approved):
This is awesome and a necessary first step to catch all errors here.

sayerhs:
@rafmudaf 

- when did OpenFAST move to C++17 standard?
- OpenFAST C++ driver is still part of the install target right? So even if we only use the library to link to nalu-wind, installation step would still fail. So this is a breaking change.

PR #677, opened 2021-03-05
BD Driver bug fix and expand BeamDyn unit testing

michaelasprague:
Added unit test test_BD_QuadraturePointData.F90, which tests multiple components required to calculate quantities at quadrature points; also tests components found in BD_QuadraturePointDataAt0.

Fixed a bug in Driver_Beam.f90

Rewrote logic in BeamDyn_Subs.f90:BD_ComputeIniNodalCrv, which appears to give the same answer but was more intuitive. 

All unit tests pass. 

BeamDyn regression tests will pass, except for bd_5MW_dynamic_gravity_Az90, which needs GlbRotBladeT0 = True in bd_driver.inp


modules/beamdyn/src/Driver_Beam.f90
```diff
@@ -98,8 +98,7 @@ PROGRAM BeamDyn_Driver_Program
       
       ! initialize the BD_InitInput values not in the driver input file
    BD_InitInput%RootName = TRIM(BD_Initinput%InputFile)
-   BD_InitInput%RootDisp = matmul(transpose(BD_InitInput%RootOri),BD_InitInput%GlbPos) - BD_InitInput%GlbPos
-   BD_InitInput%RootVel(1:3) = matmul(BD_InitInput%RootOri, Cross_Product( BD_InitInput%RootVel(4:6), BD_InitInput%GlbPos ))  ! set translational velocities based on rotation and GlbPos.
+   BD_InitInput%RootDisp = 0.d0
```
andrew-platt, line 101: With this change, how do you simulate starting the blade at a different azimuth?

modules/beamdyn/src/Driver_Beam.f90
```diff
@@ -98,8 +98,7 @@ PROGRAM BeamDyn_Driver_Program
       
       ! initialize the BD_InitInput values not in the driver input file
    BD_InitInput%RootName = TRIM(BD_Initinput%InputFile)
-   BD_InitInput%RootDisp = matmul(transpose(BD_InitInput%RootOri),BD_InitInput%GlbPos) - BD_InitInput%GlbPos
-   BD_InitInput%RootVel(1:3) = matmul(BD_InitInput%RootOri, Cross_Product( BD_InitInput%RootVel(4:6), BD_InitInput%GlbPos ))  ! set translational velocities based on rotation and GlbPos.
```
andrew-platt, line 102: By removing the rotational velocity input, how are you doing the initialization of a rotating blade?  If there is a hub radius, these terms will be non-zero from the OpenFAST glue code.

michaelasprague: Rotational velocity is initialized in BD_ReadDvrFile:

`   InitInputData%RootVel(1:3) = cross_product(InitInputData%RootVel(4:6),InitInputData%GlbPos(:))`

modules/beamdyn/src/Driver_Beam.f90
michaelasprague: The displacement is relative to its initial position.   Part of my verification was to show that results were the same as those produced by openfast at the different azimuth.

modules/beamdyn/src/Driver_Beam.f90
andrew-platt: Ok.  This makes sense to me now.

modules/beamdyn/src/BeamDyn_Subs.f90
```diff
@@ -1045,22 +1046,59 @@ SUBROUTINE BD_ComputeIniNodalCrv(e1, phi, cc, ErrStat, ErrMsg)
...
-
-   e2(3)    = -(e1(1)*COS(PhiRad) + e1(2)*SIN(PhiRad))/e1(3)
-   Delta    = SQRT(1.0_BDKi + e2(3)*e2(3))
-   e2(1)    = COS(PhiRad)
-   e2(2)    = SIN(PhiRad)
-   e2       = e2 / Delta
-   Rr(:,1)  = e2
-   Rr(:,2)  = Cross_Product(e1,e2)
+   ! Note that e3 corresponds with the Z-axis direction in our formulation.  For Beam theory, this would be the x-axis.
+   !Rr(:,3)  =  e3(:)
```
rafmudaf, line 1059: @michaelasprague Do we need to keep this commented code?

rafmudaf (approved):

modules/beamdyn/src/BeamDyn_Subs.f90
andrew-platt: I think we can safely toss lines 1058:1068 -- there's no reason in my opinion to keep them.

ashesh2512 (approved):
Other than some of the comments already addressed, looks all good to me!

modules/beamdyn/src/BeamDyn_Subs.f90
andrew-platt: I removed them.

PR #686, opened 2021-03-11
Bug fix for regression tests without C++ API enabled

rafmudaf:
Currently, the test list in CTest includes the C++ API regression test even if the C++ API is not configured to be built with CMake. Therefore, when building only OpenFAST and running `ctest` (no additional arguments), CTest attempt to run the C++ API test but fails since it is not compiled.

Regression tests


andrew-platt (approved):
It works as expected now!

PR #687, opened 2021-03-19
[BugFix] incorrect pitch/roll moments on tapered elements crossing water line

andrew-platt:
** This pull request is ready to be merged**

There were two negative signs missing in the calculation of the centroid for tapered members crossing the still water line, and one incorrect variable name.  The negative signs were missing in the original implementation plan.

This had been reported by a collaborator on an ARPA-E ATLANTIS project.

HydroDyn module.  Only tapered members crossing the still water surface are affected

Added a very simple test case of a tapered cylinder (0.1m taper over 40 m) with a prescribed pitch motion.

Original results showing pitching moment that was way too large:
hd_taper_HydroMyi--orig

Corrected results in this PR:
hd_taper_HydroMyi--corrected

### DIFF
--- a/modules/hydrodyn/src/Morison.f90
+++ b/modules/hydrodyn/src/Morison.f90
@@ -2687,7 +2687,7 @@ SUBROUTINE Morison_CalcOutput( Time, u, p, x, xd, z, OtherState, y, m, errStat, 
          z2      = pos2(3)
          r1      = mem%RMG(i  )                         ! outer radius element nodes including marine growth
          r2      = mem%RMG(i+1)
-         dRdl_mg = mem%dRdl_mg(i)                                    ! mass of element including marine growth
+         dRdl_mg = mem%dRdl_mg(i)                                    ! Taper of element including marine growth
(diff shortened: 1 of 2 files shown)


andrew-platt:
@mattEhall, take a look at the code change for updating the plan.  This now matches your python script (I can upload the pages of algebra if you want the proof).

Should we add some of that derivation into the docs at some point?

HaymanConsulting (approved):
Looks like in addition to the negative signs (which, as you state, did not appear in the implementation plan), there was an additional typo in the code, where the value of r1 was used instead of rh.  The implementation plan does, indeed, call out rh in that particular term.

I'll leave the verification of the final form of the equation to Matt H.

mattEhall:
These corrections look good, Andy. They match what I have in a Python verification script and match the corrected derivation. The trends in the plot of corrected moments look right. Thanks for tracking the errors down!

jjonkman (approved):
Thanks for tracking down this bug!  This is one part of the implementation plan that I was not able to check in detail.

PR #689, opened 2021-03-22
Enable types generation in CI

rafmudaf:
This pull request configures OpenFAST to build the OpenFAST Registry and use it to generate the types-files used throughout OpenFAST. Currently, the OpenFAST Registry is not tested in the automated system.

None.

GitHub Actions

I chose to add this test to the single precision compile test in order to allow developers to commit changes to the types-files and use the automated regression test as a tool to check the results. If this change were added to other GitHub Actions tests, users would be required to use the registry in their development workflow while changing types-files. I find it much simpler to make changes to the Fortran types-files directly while and then regenerate the types with the registry before submitting the pull request.


andrew-platt:
Good idea!

Issue #695, opened 2021-03-26
OpenfFAST 2.5.0 AeroDyn 15 and ElastoDyn problem with Matlab and Simulink run


eletroinf:
Hi
I am facing problems in running my simulation (working fine in OpenFAST 2.3.0) in OpenFast 2.5.0.
I am using Matlab/Simulink 2018b and I compiled OpenFast with Visual Studio and Intel Fortran Compilers.

Any suggestion about how to solve this?

That is what I am getting in Matlab 2018b output:

IPC active simulation

 **************************************************************************************************
 OpenFAST

 Copyright (C) 2021 National Renewable Energy Laboratory
 Copyright (C) 2021 Envision Energy USA LTD

 This program is licensed under Apache License Version 2.0 and comes with ABSOLUTELY NO WARRANTY.
 See the "LICENSE" file distributed with this software for details.
 **************************************************************************************************

 OpenFAST-Unversioned from 718d46f707d78e85edf1b49d3b1a63e8e23e1aae
 Compile Info:
  - Compiler: Intel(R) Fortran Compiler 1910
  - Architecture: 64 bit
  - Precision: single
  - Date: Mar 26 2021
  - Time: 10:29:37
 Execution Info:
  - Date: 03/26/2021
  - Time: 11:16:52-0300

 OpenFAST input file heading:
     FAST Certification Test #13: WindPACT 1.5 MW Baseline with many DOFs with VS and VP and FF
     turbulence.

 Running ElastoDyn.
 Nodal outputs section of ElastoDyn input file not found or improperly formatted.
 Running AeroDyn.
 Nodal output section of AeroDyn input file not found or improperly formatted.
 Warning: Turning off Unsteady Aerodynamics because C_nalpha is 0. (node 1, blade 1)
 Warning: Turning off Unsteady Aerodynamics because C_nalpha is 0. (node 2, blade 1)
 Warning: Turning off Unsteady Aerodynamics because C_nalpha is 0. (node 1, blade 2)
 Warning: Turning off Unsteady Aerodynamics because C_nalpha is 0. (node 2, blade 2)
 Warning: Turning off Unsteady Aerodynamics because C_nalpha is 0. (node 1, blade 3)
 Warning: Turning off Unsteady Aerodynamics because C_nalpha is 0. (node 2, blade 3)
 Running InflowWind.

  OpenFAST completed.

Error using test13_fast_v10_automated_test (line 45)
Error due to multiple causes.
Caused by:
[...]

Issue #692, opened 2021-03-23
OpenFAST Simulink compilation in dev branch crashing.


aidowu:
I want to ask if you can give a timeline for when the Simulink OpenFAST version (at least up to the TCF MOD2 merge) in the development branch will be ready? 

We tried compiling currently as is, to get started on our work, pending the time the updates gets completed. While the OpenFAST compiled for the command line works, the Simulink compilation keeps crashing MATLAB.

andrew-platt:
How are you compiling it, which compiler, and which platform?

aidowu:
I am compiling following the steps listed in the attached file which I found on this forum. I am compiling for windows x64. It does compile and I generate the S-Function from the compilation but when I try running it, it crashes MATLAB. I have followed the steps in the attached document to compile the main branch and that runs well.
[OpenFAST_Simulink.pdf](https://github.com/OpenFAST/openfast/files/6192553/OpenFAST_Simulink.pdf)

andrew-platt:
Are there any errors from the `create_MATLAB_SFunc.m` script?

aidowu:
There are no errors. Here is a picture of the directory containing the compiled files.
![image](https://user-images.githubusercontent.com/9618674/112214573-bb445c80-8bf5-11eb-9c74-27329ff4e0e8.png)

andrew-platt:
What version of the Intel Fortran compiler are you using with Visual Studio?

Which Simulink example case are you running?

Also, I'm not entirely sure what you are asking about in terms of timeline.  The TCF MODS2 branch was merged in January, but did not contain any updates for control channels.  If you are asking about the cable control channels, there is a branch in development associated with this pull request: #664.

aidowu:
I am using Intel Parallel Studio X17 and I didn't run any of the Simulink example cases but ran a project case that worked with the main dev Simulink compilation. I will go run an example case to see if I can narrow down where the issue might be.

Looking through the pull request you mentioned above, I'll modify my question by asking if the Simulink version in the development branch has been validated for the cable control updates? That way I can know if the error is from my path or if it might be a bug?

andrew-platt:
Did you update the input file set for your project?  Several input files have changed between the main branch and the development branch.  There had been some error handling issues that were causing MATLAB to crash without telling the user what occurred -- all the known ones have been corrected in the development branch.

The development branch does not currently contain any of the cable control functionality.  We hope to merge that in by the end of the month (#664).

aidowu:
[...]

PR #702, opened 2021-03-31
Simulink

bjonkman:
This fixes an issue where the Simulink interface assumed channel names being passed did not include the time channel. This was likely introduced when Envision gave NREL code that contained changes for the upcoming steady-state solver.

Fixes "size of NumOutputs is invalid" errors in #692 and #695

Simulink interface

There are no regression tests for the Simulink interface, so this error was not caught and will not change any other results.


andrew-platt:
Thanks for finding that!  I was really confused as to where this issue was coming from.

andrew-platt (approved):
I verified this corrects the issue on my `f/StrucCtrl_controls` branch.

PR #703, opened 2021-04-02
[BugFix] unallocated AD%y%Rotors when AD14 is used with cpp interface

### LINKED ISSUE #692: OpenFAST Simulink compilation in dev branch crashing.

andrew-platt:
When using the cpp interface with AD14, the `FillOutputAry_T` routine tried to pass `Turbine(iTurb)%AD%y%Rotors(1)` to `FillOutputAry`, but the `AD%y%Rotors` was not allocated.  This bug was introduced with PR #672.

The logic within `FillOutputAry` and the variables passed into it have been modified somewhat similarly to how BeamDyn and IceDyn outputs are handled.

NOTE: this set of routines is only called from the FAST_Library, so this issue would only appear with the cpp interface when AD14 is used (I only know of one Simulink example case that this would affect)

*This PR is ready to merge*

The example Simulink case `Run_Test01_SIG.m` used AD14 with Simulink.  A segmentation fault occurred with this example case (this case is not included in the regression tests).

Close #692

### DIFF
--- a/modules/openfast-library/src/FAST_Subs.f90
+++ b/modules/openfast-library/src/FAST_Subs.f90
@@ -4762,15 +4762,9 @@ SUBROUTINE WriteOutputToFile(n_t_global, t_global, p_FAST, y_FAST, ED, BD, AD14,
    IF ( y_FAST%WriteThisStep )  THEN
 
          ! Generate glue-code output file
-        if (allocated(AD%y%rotors)) then
          CALL WrOutputLine( t_global, p_FAST, y_FAST, IfW%y%WriteOutput, OpFM%y%WriteOutput, ED%y%WriteOutput, &
-               AD%y%rotors(1)%WriteOutput, SrvD%y%WriteOutput, HD%y%WriteOutput, SD%y%WriteOutput, ExtPtfm%y%WriteOutput, MAPp%y%WriteOutput, &
+               AD%y, SrvD%y%WriteOutput, HD%y%WriteOutput, SD%y%WriteOutput, ExtPtfm%y%WriteOutput, MAPp%y%WriteOutput, &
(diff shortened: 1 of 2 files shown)


ebranlard (approved):
Thanks for finding this!

modules/openfast-library/src/FAST_Subs.f90
```diff
@@ -4959,10 +4953,12 @@ SUBROUTINE FillOutputAry(p_FAST, y_FAST, IfWOutput, OpFMOutput, EDOutput, ADOutp
       END IF            
       
       IF ( y_FAST%numOuts(Module_AD) > 0 ) THEN
-         indxLast = indxNext + SIZE(ADOutput) - 1
-         OutputAry(indxNext:indxLast) = ADOutput
-         indxNext = IndxLast + 1
-      END IF
+         do i=1,SIZE(y_AD%Rotors)
+            indxLast = indxNext + SIZE(y_AD%Rotors(i)%WriteOutput) - 1
```
bjonkman, line 4957: If `size(y_AD%Rotors)` is ever more than 1, we should also put an `if (allocated(y_AD%rotors(i)%WriteOutput)` around these calculations before using the size function. If one rotor has no outputs but others do, it could cause a seg fault otherwise.

andrew-platt: Good point. I might as well add that now.

I had initially skipped such checking here since there are a lot of other places in the glue code that will need updates for more than one rotor.

bjonkman: Yeah, that will be a fun job later, too. :) Since this one actually uses a loop on the size of the rotors now, it caught my attention.

rafmudaf (approved):

PR #720, opened 2021-04-16
inflowWind C-bound interface and python wrapper

nrmendoza:
Ability to couple inflowWind to other c-based codes through a c-bound interface (specifically python). There is a python example driver code that illustrates how to use interface; the python-c library, and a c-bound wrapper for InflowWind.f90. This is not dependent on any other PRs.

None

inflowWind module only. uses NWTC IO library. Does not affect any existing functionality - only adds to the existing IFW code base. 

Part of the ATLANTIS CCT2 project (collaborating with Sandia to merge HD, MD, and IFW into Julia-based OWENS)

New regression tests for this IFW library, but existing tests should not be affected. New scripts have been demonstrated to work for WindTypes = 1, 2, 3.

**Tasks**
- [x] Update Visual Studio build system
- [x] Add regression test for inflowWind library interface 
- [x] Port testing over to automated testing
- [x] cmake build for inflowWind library for Windows


andrew-platt:
Todo (AP):
- [ ] move the `archive` folder to r-test.  Will need to add a .bts file (or use one of the 5MW ones for this regression test)
- [ ] remove the one commit for the moordyn related stuff (it goes into the next PR).

andrew-platt (approved):

modules/inflowwind/python-lib/inflowwind_library.py
```diff
@@ -0,0 +1,206 @@
...
+        self.error_message = create_string_buffer(1025)
+
+        # This buffer for the channel names and units is set arbitrarily large
+        # to start.  InflowWind only has a maximum of 9 outputs at present, but
+        # may be expanded.  Channel name and unit lengths are currently hard
+        # coded to 20 (this must match ChanLen in NWTC_Base.f90).
+        self._channel_names = create_string_buffer(20 * 4000)
+        self._channel_units = create_string_buffer(20 * 4000)
+
+        self.dt = c_double(0)               # InflowWind must be passed
```
andrew-platt, line 69: Is this correct?  If we don't set `dt` in the driver that calls this, an error of `must be real number, not c_double` when `dt` is passed on the call it `inflowwind_init_c`.

andrew-platt:
After discussion with @rafmudaf, it probably makes sense to move the `class` for writeoutput from the example driver in the r-test to the `inflowWind_Library.py`

- [ ] move class for writing to the library for simplicity.

andrew-platt (approved):

rafmudaf (commented):
Whats the difference between `modules/inflowwind/src/IfW_C.f90` and `modules/inflowwind/src/IfW_C_Binding.f90`? Is `IfW_C` something left over from my pull request to @andrew-platt?

Issue #719, opened 2021-04-16
Issues compiling AeroDyn Driver for #688 & OpenFAST\dev branch


mayankchetan:
**Bug description**

I initially tried compiling the AeroDyn Driver using #688, but the `AeroDyn_Driver_Types.f90` file does not generate correctly while compiling AeroDyn Driver via the visual studio solution. Next, I tried compiling the AeroDyn Driver using the `dev` branch and had issues with the `Registry.lib` file.

**To Reproduce**

Steps to reproduce the behavior for #688:
1. Clone https://github.com/ebranlard/openfast/tree/f/driver .
2. Compile AeroDyn Driver using `AeroDyn_Driver.sln`.
3. Run generated `AeroDyn_Double_Driver.exe`, but no outputs/does not run.
4. Clear solution and re-compile.
5. Error while generating `AeroDyn_Driver_Types.f90`.

![image](https://user-images.githubusercontent.com/4620557/114970345-de9ea800-9e3f-11eb-8eea-5a61c2d54e96.png)

Steps to reproduce the behavior for `OpenFAST\dev` branch:
1. Clone `dev` branch
2. Compile AeroDyn Driver using `AeroDyn_Driver.sln` 
3. Error while generating `Aerodyn Driver`

![image](https://user-images.githubusercontent.com/4620557/115037120-c22c5b00-9e93-11eb-9533-62d1194afb5b.png)

**Expected behavior**

- No errors while generating `AeroDyn_Driver_Types.f90` for #688.
- No errors while generating AeroDyn Driver for `dev` branch.

**Screenshots, if applicable**

`.txt` appears where it's not expected

![image](https://user-images.githubusercontent.com/4620557/114970562-4b19a700-9e40-11eb-824e-9197e5bfaa33.png)

`AeroDyn_Driver_Types.f90` doesn't complete generating via `FAST registry`

![image](https://user-images.githubusercontent.com/4620557/114970585-553ba580-9e40-11eb-8b13-67171bcbc7dd.png)

**OpenFAST Version**

- https://github.com/ebranlard/openfast/tree/f/driver
- `dev`

**System Information (please complete the following information):**
 - OS: Windows 10 [20H2]
 - Compiler: [Intel OneAPI Fortran Compiler](https://software.intel.com/content/www/us/en/develop/tools/oneapi/hpc-toolkit/download.html) + Visual Studio 2019
 - Compiler settings: Release x64, Release Double x64

**Additional context**
[...]

PR #725, opened 2021-04-19
Updated linking dependencies for newer version of Visual Studio

bjonkman:
Some newer versions of VS want to link a Registry library into executables because it says the project depends on the Registry executable. This should fix that issue.

https://github.com/OpenFAST/openfast/issues/719

Visual Studio build system

This will not affect any existing tests.


andrew-platt (approved):
This looks good to me.  Thanks for the updated VS files!

PR #728, opened 2021-04-22
AllBldOuts more forgiving and support for 0

ebranlard:
Ready to merge

AllBldNdsOuts would fail if the user specified 0 blades or more than the current number of blades. This fix is more forgiving, and returns when the number of outputs is 0 to avoid accessing some unallocated array.

AeroDyn

### DIFF
--- a/modules/aerodyn/src/AeroDyn_AllBldNdOuts_IO.f90
+++ b/modules/aerodyn/src/AeroDyn_AllBldNdOuts_IO.f90
@@ -195,6 +195,11 @@ SUBROUTINE Calc_WriteAllBldNdOutput( p, p_AD, u, m, m_AD, y, OtherState, Indx, E
          ! Initialize some things
       ErrMsg = ''
       ErrStat = ErrID_None
+      ! NOTE: if no blade outputs, we return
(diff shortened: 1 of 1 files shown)


andrew-platt (approved):
Good catch!

bjonkman:
@ebranlard,  I'm trying to understand where it might have accessed an unallocated array. It looks like the loops would go from 1 to 0, which means they would (should?) be skipped. Did you actually experience that behavior, other than just not generating ANY blade output if, for example, you had set this to output 3 blades on a 2 blade turbine?

ebranlard:
If I recall, the issue occurs about line 222, p%BldNd_NumOuts is >0, but p%BldNd_OutParam is not allocated, which is likely the real issue (I didn't take care of it).  I've seen the issue either by setting the BldNd_BladesOut=0 or BldNd_BladesOut>NumBlades. In all cases, I have several channels in OutListAD.

bjonkman:
Thanks! I'll look closer. I'm pretty sure this logic is used in ElastoDyn and BeamDyn, too, so I wanted to double check (though they might not get to that part of the code if there aren't any blades).

Issue #617, opened 2020-12-18
Develop Prescribed Platform Motion capability with aero-elasticity 


Amarelio:
Good evening,

I'm trying to implement a capability in OpenFAST that will allow a user to prescribe periodic platform motions to the simulations. The objective is to run these simulations with aero-elasticity so I'm trying to see how to implement this in a fully coupled framework instead of in the AeroDyn module alone. I'll show below where in the code I'm standing now and where I'm thinking of making the changes. If you could give me some guidance, it would be amazing.

As far as my understanding goes, ElastoDyn is the module that receives the loads and calculates the displacements, velocities and accelerations of the system's DOF, so the motion prescription should be done here. I also went through the code starting from the OpenFAST driver (FAST_Prog.f90) and it lead me to ElastoDyn. In the beginning of ElastoDyn.f90 (line 557), there's this:
``` 
! SEE IF THESE NEED TO BE CALLED (i.e., if UpdateStates was called, these values are already calculated)
   IF ( UpdateValues ) THEN
         ! Update the OtherState data by calculating the derivative...
      !OtherState%HSSBrTrqC = SIGN( u%HSSBrTrqC, x%QDT(DOF_GeAz) )
      CALL ED_CalcContStateDeriv( t, u, p, x, xd, z, OtherState, m, dxdt, ErrStat, ErrMsg ) ! sets m%QD2T = dxdt%QDT
      CALL ED_DestroyContState( dxdt, ErrStat2, ErrMsg2 )
      IF (ErrStat >= AbortErrLev) RETURN
   END IF 
```
and ``` UpdateValues ``` is set to ``` .TRUE. ```.

At the beginning of the function ```ED_CalcContStateDeriv``` (ElastoDyn.f90 line 1825), there's this:
```
         ! set the coordinate system variables:
      CALL SetCoordSy( t, m%CoordSys, m%RtHS, u%BlPitchCom, p, x, ErrStat2, ErrMsg2 )
         call setErrStat(ErrStat2,ErrMsg2,ErrStat,ErrMsg,RoutineName)
         IF (ErrStat >= AbortErrLev) RETURN

      CALL CalculatePositions(        p, x, m%CoordSys,    m%RtHS ) ! calculate positions
      CALL CalculateAngularPosVelPAcc(p, x, m%CoordSys,    m%RtHS ) ! calculate angular positions, velocities, and partial accelerations, including partial angular quantities
[...]

Issue #541, opened 2020-09-08
Wind turbine analysis subjected to earthquake excitation


zz17635:
Greetings!

It seems the "SubDyn" module in OpenFAST always assumes a fixed-base boundary condition during a dynamic analysis. However, I remember reading in a previous article stating that it was possible to perform a wind turbine analysis with earthquake excitation at its base using the FAST software. I wander if it is still an accessible function in the current OpenFAST?

Also, just for a discussion, since the OpenFAST framework bundles a number of different modules in charge of different aspects of wind turbine simulation, how hard would it be to add an additional module to analyse the soil-structure interaction (SSI) using a beam-element model of wind turbine pile (or incorporate the OpenFAST with other software such as OpenSEES for modelling its pile and the associated SSI phenomenon)?

Thank you

jjonkman:
Dear @zz17635,

Regarding your first question, I believe you are referring to the Seismic version of FAST v7, which allowed the user to prescribe motion at the tower base to simulate seismic loading of wind turbines.  Unfortunately, the functionality of Seismic from FAST v7 has not yet been ported over to FAST v8 or OpenFAST.  The Seismic version of FAST v7 is still available if you wish to use it.

Regarding your second question, this question is indeed timely.  NREL has been working on developing a new module (named SoilDyn) for modeling soil-structure interaction. SoilDyn includes options for interfacing to the REDWIN DLLs for a super-element representation of the soil hysteresis, as well as provides an option for a simple 6x6 stiffness matrix (coupled springs model), and a placeholder for future implementation of nonlinear p-y and t-z curves.  SoilDyn has been developed in a feature branch of OpenFAST and pull request will be issued soon to bring this into the Dev branch.

Best regards,

zz17635:
Thank you for the info @jjonkman , it would be interesting to try out FAST v7.

It is quite difficult to find the old versions so I shear the link here for anyone else who might come in here:
[https://wind.nrel.gov/forum/wind/viewtopic.php?f=4&t=2517&p=14833&hilit=FAST+v7+download#p14833](url)

I also found a documentation for seismic analysis in FAST, named _Seismic Loading for FAST_:
[https://pdfs.semanticscholar.org/3343/4fb2c054ca04126c60805e46631b3bd9fb08.pdf](url)

Only, from the link above, the downloaded FAST v7 or v8 seems does not have the certifying test examples as described in the seismic documentation. Is this link directing me to the correct FAST and where could I find relevant examples?

[...]

Issue #317, opened 2019-07-31
Feature Request - Ability to Apply Time-Dependent Point Loads at Various Points on the Wind Turbine Structure


jjonkman:
**Is your feature request related to a problem? Please describe.**
When modeling laboratory-based tests (e.g., forced oscillation of a blade in a test stand, forced oscillation of the tower-base on a seismic shake table, or forced oscillation of a floating platform in a wave tank), there is often a need to apply time-varying point loads (forces, and moments) in an OpenFAST model at one or more locations on the wind turbine structure (e.g., blade tip, tower-top, platform) to mimic the experimental set-up.  It would be great to enhance OpenFAST to support this feature.

**Describe the solution you'd like**
Introduce a new module in OpenFAST that allows the user to specify user time series of 6-component loads (3 forces, 3 moments, dependent on time) at one or more user-specified locations (X,Y,Z).  The module need not have inputs or states.  The module outputs would be the lumped time-dependent loads on a point mesh that the OpenFAST glue / driver code will use mesh mapping to transfer as input to the various structural modules (e.g., ElastoDyn, BeamDyn, SubDyn) as applied loads.

andrew-platt:
In the case of a blade in a test stand, would that be best served through the BeamDyn driver code?

jjonkman:
@andrew-platt -- I agree if you have a BeamDyn model cantilevered to a test stand that the BeamDyn driver would function well.  The feature I am requesting above can be used if you want to excite an ElastoDyn blade model (ElastoDyn does not currently have a driver), a BeamDyn model with free-free boundary conditions (the otherwise cantilevered end in BeamDyn can be modeled as free through the ElastoDyn platform degrees of freedom when BeamDyn is coupled to OpenFAST), a full wind turbine, etc..

andrew-platt:
Thanks for the clarification.  I'll add that the ability to apply time-dependent loads in the BeamDyn driver is also a feature that would be nice to have.

andrew-platt:
This is now possible from the Structural Control module in ServoDyn.  This was merged in with PR #607.  An example test case is given in the OpenFAST glue-code regression test `StC_test_OC4Semi`.  In this case, a time dependent loads applied to all 3 blades simultaneously using the `BStC` option in ServoDyn.

Four options exist in ServoDyn for applying loads:
- `BStC` -- can apply prescribed loads at any point along the blades (either BeamDyn or ElastoDyn)
[...]

PR #739, opened 2021-05-03
[BugFix] Logic error in glue code for StC loads on SD with no HD

andrew-platt:
This PR is ready to merge.

PR #607 introduced the ability to apply Structural control loads to the SubDyn mesh.  However, if HydroDyn was not used with SubDyn, then the SubDyn load meshes for load mapping in the glue code would not be created.  This caused segmentation loads if an StC load was applied to SubDyn without running HydroDyn (i.e. in a simulated earthquake).

#317 
#541
#617 

ServoDyn StC loads with SubDyn, but without HydroDyn.

This came to our attention from an outside collaborator at NTNU who is looking at the ability to apply seismic loads at the base of a jacket structure.  For testing, they were turning of HydroDyn which triggered this segmentation fault.

There are no current test cases for this feature.  We are expecting to add one in a few months once this method of applying seismic loads has been validated.

### DIFF
--- a/modules/openfast-library/src/FAST_Solver.f90
+++ b/modules/openfast-library/src/FAST_Solver.f90
@@ -2345,7 +2345,7 @@ SUBROUTINE FullOpt1_InputOutputSolve( this_time, p_FAST, calcJacobian &
       
       CALL Create_FullOpt1_UVector(u, u_ED%PlatformPtMesh, u_SD%TPMesh, u_SD%LMesh,  &
                u_HD%Morison%Mesh, u_HD%WAMITMesh, u_ED%HubPtLoad, MeshMapData%u_BD_RootMotion, u_Orca%PtfmMesh, &
-               u_ExtPtfm%PtfmMesh, p_FAST )
+               u_ExtPtfm%PtfmMesh, u_SrvD%SStC, p_FAST )
                   
       K = 0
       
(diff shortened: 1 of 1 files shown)


ebranlard (approved):

modules/openfast-library/src/FAST_Solver.f90
```diff
@@ -3091,6 +3092,7 @@ SUBROUTINE U_FullOpt1_Residual( y_ED2, y_SD2, y_HD2, y_BD2, y_Orca2, y_ExtPtfm2,
       !..................
 
          IF ( p_FAST%CompServo == Module_SrvD .and. allocated(y_SrvD%SStC) ) THEN
+print*,'Mapping loads from SrvD to u_SD_LMesh_2'
```
bjonkman, line 3095: is this staying? :)

andrew-platt: oh whoops...

andrew-platt: Removed now.

danielmpedersen:
Dear Andy.
There is another error that has shown up with the error is as follows:
 _Generating incident wave kinematics and current time history.
 Error in call to FFT. The imaginary component at the largest positive frequency must be zero.
 FAST_InitializeAll:HydroDyn_Init:Waves_Init:VariousWaves_Init:Error occured while applying the  FFT to InitOut%WaveElev._
This error comes when I try to simulate EQ, wind and waves. The error does not show up when I only simulate EQ and waves with the same input files. 
I have attached my HydroDyn file but I do not think that the problem lies there. Hope you could help me with this error, since it is urgent for me to do this simulations. 
[Jacket_HydroDynzip.zip](https://github.com/OpenFAST/openfast/files/6566380/Jacket_HydroDynzip.zip)

Issue #743, opened 2021-05-07
Compilation failure with gfortran 11.1.0


fxcoudert:
**Bug description**

Compilation of openfast 2.5.0 from source on macOS with gfortran 11.1.0 (latest gcc release) leads to compilation failure:

```
/tmp/openfast-20210507-37695-1n1r9id/openfast-2.5.0/modules/openfast-library/src/FAST_Solver.f90:3016:39:
 |                      CALL CleanUp()
 |                                   2    
......
 |             do nb=1,size(y_BD_perturb)
 |                                       1
Error: Index variable 'nb' redefined at (1) in procedure 'cleanup' called from within DO loop at (2)
```

**To Reproduce**
1. Compile with `cmake .. -DCMAKE_C_FLAGS_RELEASE=-DNDEBUG -DCMAKE_CXX_FLAGS_RELEASE=-DNDEBUG -DCMAKE_INSTALL_PREFIX=/usr/local/Cellar/openfast/2.5.0_1 -DCMAKE_INSTALL_LIBDIR=lib -DCMAKE_BUILD_TYPE=Release -DCMAKE_FIND_FRAMEWORK=LAST -DCMAKE_VERBOSE_MAKEFILE=ON -Wno-dev -DBUILD_TESTING=OFF -DCMAKE_OSX_SYSROOT=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk -DDOUBLE_PRECISION=OFF -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBLA_VENDOR=OpenBLAS`
2. Then `make openfast`
3. See the error

**OpenFAST Version**
Source is https://github.com/openfast/openfast/archive/v2.5.0.tar.gz

**System Information (please complete the following information):**
 - OS: macOS 11.3
 - Compiler: gfortran 11.1.0
 - Homebrew issue: https://github.com/Homebrew/homebrew-core/pull/74843#issuecomment-834448708
 - Full logs: https://github.com/Homebrew/homebrew-core/pull/74843/checks?check_run_id=2518126449

PR #744, opened 2021-05-07
Avoid gfortran 11 error in FAST_Solver.f90

fxcoudert:
See https://github.com/OpenFAST/openfast/issues/743

This pull request should make it possible to compile openfast with gfortran 11.1.0 (the latest version)


andrew-platt:
Thanks for the PR!  This is indeed a legitimate compiler complaint. 

The `CleanUp` routine is only called when the there is an error that will terminate the `FullOpt1_InputOutputSolve` routine. So in some ways this doesn't matter that it gets reset in the `CleanUp` routine called within a loop using that variable but the compiler doesn't have any way of knowing that.

@rafmudaf, can we merge this into the upcoming v2.6.0 release?

rafmudaf:
@andrew-platt Lets keep it in the 3.0 release since it is also imminent.

I've updated the rules for GitHub Actions to run when changing a pull request and merged that into this branch. After the full suite of tests finished (just for completeness since this is a small change), lets merge.

Thanks @fxcoudert!

PR #746, opened 2021-05-10
Add Missing Attribute for sc_end in the FAST.Farm Super Controller

jjonkman:
The !DEC$ ATTRIBUTES DLLEXPORT was missing for sc_end and has now been added.

Applies only to the super controller of FAST.Farm, built with Intel Fortran on Windows.  This will not effect any of the r-test results.


jjonkman:
I'm cancelling this PR and replacing with https://github.com/OpenFAST/openfast/pull/749.

PR #749, opened 2021-05-11
Add Missing Attribute for sc_end in the FAST.Farm Super Controller

jjonkman:
The !DEC$ ATTRIBUTES DLLEXPORT was missing for sc_end and has now been added.

Applies only to the super controller of FAST.Farm, built with Intel Fortran on Windows. This will not effect any of the r-test results.

### DIFF
--- a/modules/supercontroller/src/SC_DLL.F90
+++ b/modules/supercontroller/src/SC_DLL.F90
@@ -305,6 +305,7 @@ subroutine sc_end ( errStat, errMsg )  bind (C, NAME='sc_end')
 
    implicit                        none
 #ifndef IMPLICIT_DLLEXPORT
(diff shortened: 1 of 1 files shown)


andrew-platt:
For clarification: the `p%UseLegacyInterface = .TRUE. !InputFileData%UseLegacyInterface` change prevents an unnecessary warning during initialization.  The `InputFileData%UseLegacyInterface` was never set, so the DLL initialization could try loading the new interface and giving a warning about it (it was fault tolerant of this and would then try the legacy).

bjonkman:
It looks like `InputFileData%UseLegacyInterface=.true.` was in the original pull request (line 2203 of ServoDyn.f90 in #373). But that line went missing when the input file parsing was changed to use strings (#607).

PR #760, opened 2021-05-19
NWTC_IO: Add InitFileInfo_FromNullCString subroutine

andrew-platt:
This PR is ready to be merged

This routine allows for passing an input file as one long string with `C_NULL_CHAR` delimiters between lines of the input file.
Also added a new interface for `InitFileInfo` to simplify usage.

Related to PR #720 -- this new interface allows simpler construction of the string for the input file on the python side.

No direct impacts on any existing modules.

The `nwtc_library_utest` was updated to account for the updated parsing.

### DIFF
--- a/modules/nwtc-library/tests/test_NWTC_IO_FileInfo.F90
+++ b/modules/nwtc-library/tests/test_NWTC_IO_FileInfo.F90
@@ -18,7 +18,7 @@ subroutine test_initfileinfo()
 
     integer(IntKi) :: error_status
     character(ErrMsgLen) :: error_message
-    character(1024) :: input_strings(num_lines)
+    character(MaxFileInfoLineLen) :: input_strings(num_lines)
     type(FileInfoType) :: file_info_type
     integer :: i
 
@@ -62,7 +62,8 @@ subroutine test_initfileinfo2()
 
     integer(IntKi) :: error_status
     character(ErrMsgLen) :: error_message
-    character(1025) :: input_strings(num_lines)
+    character(MaxFileInfoLineLen*2) :: input_strings(num_lines)
+    character(MaxFileInfoLineLen*2) :: tmpstring
     type(FileInfoType) :: file_info_type
     integer :: i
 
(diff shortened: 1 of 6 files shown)


modules/nwtc-library/tests/test_NWTC_IO_FileInfo.F90
```diff
@@ -62,7 +62,8 @@ subroutine test_initfileinfo2()
 
     integer(IntKi) :: error_status
     character(ErrMsgLen) :: error_message
-    character(1025) :: input_strings(num_lines)
+    character(MaxFileInfoLineLen*2) :: input_strings(num_lines)
```
rafmudaf, line 65: Why the `*2` here?

modules/nwtc-library/tests/test_NWTC_IO_FileInfo.F90
```diff
@@ -73,12 +74,46 @@ subroutine test_initfileinfo2()
...
+                     "line  1"//C_NULL_CHAR//   &
+                     "line  2"//C_NULL_CHAR//   &
+                     "line  3"//C_NULL_CHAR//   &
+                     "line  4"//C_NULL_CHAR//   &
+                     "line  5"//C_NULL_CHAR//   &
+                     "line  6"//C_NULL_CHAR
+    call InitFileInfo( input_string, file_info_type, error_status, error_message )
+    @assertEqual(num_lines, file_info_type%NumLines)
+    @assertEqual(num_files, file_info_type%NumFiles)
+    @assertEqual( 0, error_status )
```
rafmudaf, line 115: It would be good to add a test on the contents of the resulting array after its parsed. You'll probably have to make another array of input strings since `input_string` here is all one string.

modules/nwtc-library/tests/test_NWTC_IO_FileInfo.F90
```diff
@@ -73,12 +74,46 @@ subroutine test_initfileinfo2()
         "line 3",  &
         "line 4"   &
     /)
+    ! make the last character a + so a trim does not reduce this string length
+    tmpstring=input_strings(5)
```
rafmudaf, line 78: Can you explain this more? Why do we need to add the `a` at the end?

modules/nwtc-library/tests/test_NWTC_IO_FileInfo.F90
andrew-platt: Just to make sure it was longer than the standard size.  We could make it `MaxFileInfoLineLen+1` instead.

modules/nwtc-library/tests/test_NWTC_IO_FileInfo.F90
andrew-platt: The parsing in the `InitFileInfo_FromStringArray` was modified to check if the line after stripping of trailing comments was too long instead of checking the length of the array passed in.  This allows us to keep super long comments on a line without issue.  So, to test that this worked, I had to add the `a` at the end of the string so that the trimmed length was long enough.

rafmudaf: Ah got it, thanks.

modules/nwtc-library/tests/test_NWTC_IO_FileInfo.F90
rafmudaf: No problem, this makes sense.

rafmudaf (approved):
This is great, good work @andrew-platt

Issue #830, opened 2021-09-16
Bug report: Incorrect Treatment of MSL2SWL within HydroDyn


jjonkman:
**Bug description**
In OpenFAST, the reference positions of all nodes of all module-level input and output meshes are stored in the inertial frame coordinate system, which for offshore wind turbines is defined as the intersection of the undisplaced tower centerline and the mean sea level (MSL).  These reference positions are used in the mesh-mapping within OpenFAST to transfer motions and loads between modules, including rigid-body behavior between nodes that are offset across modules.

Within HydroDyn, there is an input `MSL2SWL` that defines the offset between the MSL and the still water level (SWL), e.g., to represent the change in water depth associated with tidal variations or storm surge.  This change in water depth does not change the inertial frame coordinate system, which is relative to MSL regardless.  However, in HydroDyn, it is important to know the position of hydrodynamic nodes relative to the SWL, e.g. for the calculation of local wave kinematics and hydrostatic pressure.  It has been discovered that the module-level input and output meshes within HydroDyn stored the reference position relative to SWL instead of relative to MSL.  This allowed HydroDyn to use the proper relative water depth in its internal calculations, but leads to offset in the mesh mappings between HydroDyn and other modules like ElastoDyn and SubDyn.  Thus, any module-to-module coupling will result in errors when `MSL2SWL` is nonzero.

Until this bug is fixed, we recommend setting `MSL2SWL` = 0.

**To Reproduce**
Any model with nonzero setting of `MSL2SWL` will produce incorrect results when HydroDyn is coupled to other OpenFAST modules like ElastoDyn and SubDyn.

**Expected behavior**
HydroDyn needs to be updated so that the reference position of its meshes is relative to MSL, but with the internal calculations of wave kinematics and hydrostatic pressure relative to SWL.

@HaymanConsulting is currently working to fix this bug.

**OpenFAST Version**
It appears that this bug has been around for many years.

jjonkman:
Closed by PR #831.

jjonkman:
I'm reopening this issue.  I believe PR #831 fixed the issue in HydroDyn, ElastoDyn, and SubDyn, however, it appears that the offset between MSL and SWL is not properly captured in the mooring modules (MAP++, MoorDyn, and FEAMooring).  The effect of MSL2SWL needs to be captured within the mooring modules as well.

PR #831, opened 2021-09-18
HydroDyn Input/Output meshes: change from SWL to MSL for consistency with OF glue code

HaymanConsulting:
This pull request is ready to merge.  Please examine why AeroDyn regression is failing for MacOS [@rafmudaf ].  We should add a future test for MSL2SWL.

In OpenFAST, the reference positions of all nodes of all module-level input and output meshes are stored in the inertial frame coordinate system, which for offshore wind turbines is defined as the intersection of the undisplaced tower centerline and the mean sea level (MSL). These reference positions are used in the mesh-mapping within OpenFAST to transfer motions and loads between modules, including rigid-body behavior between nodes that are offset across modules.

Within HydroDyn, there is an input MSL2SWL that defines the offset between the MSL and the still water level (SWL), e.g., to represent the change in water depth associated with tidal variations or storm surge. This change in water depth does not change the inertial frame coordinate system, which is relative to MSL regardless. However, in HydroDyn, it is important to know the position of hydrodynamic nodes relative to the SWL, e.g. for the calculation of local wave kinematics and hydrostatic pressure. It has been discovered that the module-level input and output meshes within HydroDyn stored the reference position relative to SWL instead of relative to MSL. This allowed HydroDyn to use the proper relative water depth in its internal calculations, but leads to offset in the mesh mappings between HydroDyn and other modules like ElastoDyn and SubDyn. Thus, any module-to-module coupling will result in errors when MSL2SWL is nonzero.

Bug report: Incorrect Treatment of MSL2SWL within HydroDyn [#830](https://github.com/OpenFAST/openfast/issues/830)

HydroDyn, Morison

None

This feature was never part of the OpenFAST test suite

### DIFF
--- a/modules/hydrodyn/src/HydroDyn_Input.f90
+++ b/modules/hydrodyn/src/HydroDyn_Input.f90
@@ -4003,10 +4003,7 @@ SUBROUTINE HydroDynInput_ProcessInitData( InitInp, ErrStat, ErrMsg )
 
       InitInp%Morison%MGTop    = -999999.0
       InitInp%Morison%MGBottom =  999999.0
-      do I = 1,InitInp%Morison%NMGDepths
-            ! Adjust the depth values based on MSL2SWL
-         InitInp%Morison%MGDepths(I)%MGDpth = InitInp%Morison%MGDepths(I)%MGDpth - InitInp%Morison%MSL2SWL
-      end do
+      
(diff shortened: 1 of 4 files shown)


jjonkman (approved):
I've reviewed these changes with @HaymanConsulting and the implementation looks good to me.

Issue #833, opened 2021-09-20
AeroDyn Driver runtime error


rafmudaf:
**Bug description**
Since merging #688, the AeroDyn driver fails to run when compiled in debug mode on macOS with GNU Fortran. The specific error is given below.

```
test 36
    Start 36: ad_timeseries_shutdown

36: Test command: /Users/rmudafor/miniconda3/bin/python3 "/Users/rmudafor/Desktop/of/reg_tests/executeAerodynRegressionCase.py" "ad_timeseries_shutdown" "/Users/rmudafor/Desktop/of/build/modules/aerodyn/aerodyn_driver" "/Users/rmudafor/Desktop/of/reg_tests/.." "/Users/rmudafor/Desktop/of/build/reg_tests/modules/aerodyn" "0.00001" "Darwin" "GNU"
36: Test timeout computed to be: 5400
36: At line 579 of file /Users/rmudafor/Desktop/of/modules/aerodyn/src/AeroDyn_Driver_Subs.f90
36: Fortran runtime error: Allocatable argument 'dvr' is not allocated
36: 
```

**To Reproduce**
<Update the following list with your specific information.>
1. Compile with GNU Fortran 11.2 on macOS
2. Run the following regression test case:
```
ctest -R ad_timeseries_shutdown -VV
```
3. This should fail with the error above.

**OpenFAST Version**
The help-flag in AeroDyn Driver does run and provides this version output:

```
(base) >>mbp@~/Desktop/of/build (dev)$ ./modules/aerodyn/aerodyn_driver -h

 **************************************************************************************************
 AeroDyn_driver

 Copyright (C) 2021 National Renewable Energy Laboratory
 Copyright (C) 2021 Envision Energy USA LTD

 This program is licensed under Apache License Version 2.0 and comes with ABSOLUTELY NO WARRANTY.
 See the "LICENSE" file distributed with this software for details.
 **************************************************************************************************

 AeroDyn_driver-v3.0.0-207-gfef5b7ec
 Compile Info:
  - Compiler: GCC version 11.2.0
  - Architecture: 64 bit
  - Precision: double
  - OpenMP: No
  - Date: Sep 17 2021
  - Time: 16:38:18
 Execution Info:
  - Date: 09/20/2021
  - Time: 15:21:28-0500


  Syntax is:

     AeroDyn_driver [-h] <InputFile>

  where:

     -h generates this help message.
     <InputFile> is the name of the required primary input file.

[...]

PR #835, opened 2021-09-20
GH Actions: Add OS/Compiler matrix to debug job

rafmudaf:
This pull request adds an operating system and compiler matrix to the regression-test-debug job in the automated testing system, GitHub Actions. This gap became clear when a compile-time error was introduced but not caught in the automated tests due to a lack of compiling with GNU Fortran 11.

#833 

Automated tests

We can see the results of the automated tests on this commit: https://github.com/rafmudaf/openfast/runs/3656409377?check_suite_focus=true


rafmudaf:
In addition to adding a build with Gfortran 11, we also get a build on macOS 11.

PR #839, opened 2021-09-23
Fix Bug in FAST.Farm Causing Wake Bounce-Back

jjonkman:

All tests are expected to pass.

For wake planes that are skewed relative to domain boundaries (e.g., due to nacelle-yaw), the weighting used in the spatial averaging to calculate the meandering velocity of wake planes when `Mod_Meander` = 3 could end up being negative as wake planes were leaving the domain.  But the code always assumed that the weighting was positive.  This bug fixes that.  

This solves the issue reported on our forum dated July 06, 2021: https://wind.nrel.gov/forum/wind/viewtopic.php?f=13&t=2117&start=225.  

The Ambient Wind and Wake Effects (AWAE) module of FAST.Farm.


rafmudaf:
@jjonkman Is this situation covered in the existing FAST Farm related tests?

jjonkman:
@rafmudaf -- Not really.  The FAST.Farm r-test only tracks the OpenFAST output.  The bug was most easily seen in the visualization output, although the OpenFAST output could be effected for the wake bounce back was significant.  The bug would only manifest itself if the wake planes where skewed relative to the boundary of the low-resolution domain and would only be seen in the OpenFAST output if the domain boundary was close to a wind turbine rotor.

ebranlard (approved):

rafmudaf:
Yes, that looks like a good fix compared to the plot shown in the forum. Thanks for the additional information.

PR #664, opened 2021-02-22
ServoDyn: StC controls, and cable controls for MD and SD

andrew-platt:
This PR is **not** ready for merging.  Feedback is requested from end users

### Feature description
This pull request adds the following capabilities:

- Cable control feed-through from a Bladed-style controller to the MoorDyn and SubDyn modules.  This feature is needed for several projects under the ATLANTIS program.
- Cable control feed-through from an external input (such as Simulink)
- Aerodynamic flow control actuation control -- the ability to turn on a simple model was added, along with the ability to turn off feed-through from a Bladed-style controller
- Control channel groups for 
- Extended avrSWAP array for future development
- Summary file from ServoDyn that includes information about channel allocations in communication with a Bladed-style controller.

### Related issue, if one exists
#646 -- status request about this project
#565 -- MoorDyn initial cable implementation
#537 -- SubDyn cable capabilities added
#395 -- Flap controls added to AD15

### Impacted areas of the software
This change affects the ServoDyn module and the cable controls added to MoorDyn and SubDyn in prior pull requests.

### Additional supporting information

- StC control channel feed-through from the controller to the StC (Structural Control) sub-module of ServoDyn for independent and omnidirectional dampers is now completed (`StC_DOF_MODE`=1 or 2).  A control channel set can be requested from each of the StC input files.  The StC instance will then pass the local displacement and velocity in X, Y, and Z to the controller.  The controller then can change the stiffness, damping, brake force, and external force in each of the local X, Y, and Z directions.  This is then passed to each StC that requested a control channel.  This is related to #607.

### Test results
Input files for ServoDyn change with this PR (additional lines).  There are no specific test cases for this set of features as no publicly available models using these features.

Now that summary files are produced by ServoDyn, it might be worth rerunning the baselines and adding those files to the regression test results.

- [x] Update summary files from ServoDyn

### Checklist of items to address

- [x] Bugfix for SubDyn to enable cable elements (commented out currently due to rotation issue) @ebranlard 
- [x] Update documentation for missing terms in equations (Brake force, External Force, Force Table).
- [x] Update results for StC test case

PR #902, opened 2021-11-01
Fixes for Intel in debug mode

andrew-platt:
**This PR is ready for merging, pending review**

The Intel compiler (ifort (IFORT) 2021.4.0 20210910 from the OneAPI install on Ubuntu) throws runtime errors without these changes.

1. In _ServoDyn_, the `TRANSFER` function was throwing an error at runtime about the result of the `TRANSFER` not matching the size of the `C_CHAR` array on the LHS.  Adding the optional size seems to fix this (probably not necessary for `avcOUTNAME`, but added to eliminate confusion... or from my failure to note it wasn't necessary at the time). 
2. Passing the various unallocated `y%WriteOutput` arrays to the `WrOutputLine` resulted in the following error at runtime in debug:
```
forrtl: severe (408): fort: (8): Attempt to fetch from allocatable variable WRITEOUTPUT when it is not allocated

Image              PC                Routine            Line        Source
FAST.Farm          0000000007220FC2  Unknown               Unknown  Unknown
FAST.Farm          0000000000AF4F99  fast_subs_mp_writ        4948  FAST_Subs.f90
FAST.Farm          0000000000AD84BB  fast_subs_mp_fast        4016  FAST_Subs.f90
FAST.Farm          0000000000AD79B2  fast_subs_mp_fast        3934  FAST_Subs.f90
FAST.Farm          0000000000867D0F  fastwrapper_mp_fw         504  FASTWrapper.f90
FAST.Farm          00000000007ED72D  fast_farm_subs_mp        1711  FAST_Farm_Subs.f90
FAST.Farm          000000000089027F  MAIN__                    109  FAST_Farm.f90
FAST.Farm          0000000000406E12  Unknown               Unknown  Unknown
libc-2.31.so       00007FE06F5E00B3  __libc_start_main     Unknown  Unknown
```

Related conversation on PR #664:  https://github.com/OpenFAST/openfast/pull/664/files/2d5a2cd5fc456a77a6dc6cca444e3e0763d5548c..41579a7ed30e0f05c47ebd5260b47852eee85d8d#diff-bc64b5f80cb30c32b06ed3e52051ba117d0f3e765f375de9cd5bc7a81c6e0006

No changes in software operation for other compilers or compile settings is expected.

No test results change with this.  And no, I'm not writing a unit test for this...

### DIFF
--- a/modules/elastodyn/src/ElastoDyn.f90
+++ b/modules/elastodyn/src/ElastoDyn.f90
@@ -2460,18 +2460,18 @@ END SUBROUTINE Alloc_CoordSys
 SUBROUTINE SetBladeParameters( p, BladeInData, BladeMeshData, ErrStat, ErrMsg )
 !..................................................................................................................................
 
-   TYPE(ED_ParameterType),        INTENT(INOUT)  :: p                                   !< The parameters of the structural dynamics module
-   TYPE(BladeInputData),          INTENT(IN)     :: BladeInData(:)                      !< Program input data for all blades
-   TYPE(ED_BladeMeshInputData),   INTENT(IN)     :: BladeMeshData(:)                    !< Program input mesh data for all blades
-   INTEGER(IntKi),                INTENT(OUT)    :: ErrStat                             !< Error status
-   CHARACTER(*),                  INTENT(OUT)    :: ErrMsg                              !< Error message
+   TYPE(ED_ParameterType),                      INTENT(INOUT)  :: p                 !< The parameters of the structural dynamics module
+   TYPE(BladeInputData),         ALLOCATABLE,   INTENT(IN)     :: BladeInData(:)    !< Program input data for all blades
+   TYPE(ED_BladeMeshInputData),  ALLOCATABLE,   INTENT(IN)     :: BladeMeshData(:)  !< Program input mesh data for all blades
+   INTEGER(IntKi),                              INTENT(OUT)    :: ErrStat           !< Error status
+   CHARACTER(*),                                INTENT(OUT)    :: ErrMsg            !< Error message
 
       ! Local variables:
-   REAL(ReKi)                                    :: x                                   ! Fractional location between two points in linear interpolation
-   INTEGER(IntKi )                               :: K                                   ! Blade number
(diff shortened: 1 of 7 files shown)


andrew-platt:
@bjonkman, do you mind weighing in on the allocate issue?  I think you've seen something similar before and likely have a better solution than what I put in.

bjonkman:
@andrew-platt , your solution should work fine. An alternative method would be to add the `allocatable` attribute on the variable definitions in the `WrOutputLine` routine. I used that method in HydroDyn when I was debugging in there a while back, but don't think I've actually pushed any of those changes to my OpenFAST fork.

Out of curiosity, have you checked what is in those two arrays after the `transfer` function in ServoDyn when the source character string is shorter than the output character array? I'm guessing they had some undefined values before this change. Are they now repeating the source? or blank? or ???

andrew-platt:
@bjonkman, the transfer was keeping the `c_null_char` after all.  I made a mistake in my quick code for counting the number of null characters by stepping through the array from index 1 to `len(avcOUTNAME)` instead of `size(avcOUTNAME)`.  The transfer is populating the rest of the array with null characters, in this case 12584 null characters following 17 characters in the name for a total of 12601 charcters.

so the code
```
   CHARACTER(KIND=C_CHAR)                    :: avcOUTNAME(p%avcOUTNAME_LEN)         ! OUTNAME (in: Simulation RootName; out: string for logging channels Name:Units;)
   avcOUTNAME  = TRANSFER( TRIM(dll_data%RootName)//C_NULL_CHAR,   avcOUTNAME, p%avcOUTNAME_LEN )

do i=1,p%avcOUTNAME_LEN
   if(avcOUTNAME(i) == C_NULL_CHAR)   cnt=cnt+1
enddo
print*,'avcOUTNAME: ',p%avcOUTNAME_LEN,' Null chars: ',cnt
write(CU,*) avcOUTNAME(1:p%avcOUTNAME_LEN)
```

produces:
```
 avcOUTNAME:        12601  Null chars:        12584
 LESinflow.T1.SrvD


[output truncated]
```

rafmudaf (approved):

Issue #914, opened 2021-11-10
Unexpected results when PtfmRefztRot <>0


rdamiani:
**Bug description**
When the BEM hydrodynamic solution is attained with a reference frame not aligned with the expected global inertial frame in OpenFAST, the parameter PtfmRefztRot could be used to rotate results from BEM to the OpenFAST coordinate system. Yet, simple decay simulations associated with a floater and PtfmRefztRot=180-deg do not appear to reproduce the correct response.  This might have implications for regular simulations results as well, and it appears to be a bug. 

**To Reproduce**
1. Run two simple test cases with any floater and potential flow mode, e.g., OC4 Semi Test Case from r-test directory, with no wind or wave enabled, and with an initial PtfmPitch of 5 deg, and with PtfmRefztRot=0-deg and 180-deg, respectively.
2. Examine PtfmPitch output by the simulations and check associated frequency
3. Examine differences

**Expected behavior**
In a potential flow solution, a still water pitch-decay case should return the same frequency response when the floater is rotated 180 deg. 

**Screenshots, if applicable**
<Add screenshots to help explain your problem.>
 
 Results for a barge
![image](https://user-images.githubusercontent.com/8195507/141185066-fed97a5e-0242-4212-8c17-4cae093d317b.png)

Results for the OC4 Semi
![image](https://user-images.githubusercontent.com/8195507/141193088-c423d681-a3fd-4732-9934-d6e3f49c6e35.png)

**OpenFAST Version**
**************************************************************************************************
OpenFAST

Copyright (C) 2021 National Renewable Energy Laboratory
Copyright (C) 2021 Envision Energy USA LTD

This program is licensed under Apache License Version 2.0 and comes with ABSOLUTELY NO WARRANTY.
See the "LICENSE" file distributed with this software for details.
**************************************************************************************************

OpenFAST-v3.0.0-4-g8c054308-dirty
Compile Info:
 - Compiler: Intel(R) Fortran Compiler 1910
 - Architecture: 64 bit
 - Precision: single
 - OpenMP: No
 - Date: Jul 01 2021
 - Time: 20:11:24
Execution Info:
 - Date: 11/10/2021
 - Time: 12:32:23-0700

**System Information (please complete the following information):**
 - OS: Windows 10
 - Compiler: Intel

**Additional context**
<Add any other context about the problem here.>

andrew-platt:
[...]

PR #915, opened 2021-11-10
[bugfix] Incorrect reference frame used in HD for WAMIT/WAMIT2

andrew-platt:
This PR is ready to merge

**Bug description**
An incorrect reference frame was applied to mesh used for the potential flow bodies when rotated using the `PtfmRefztRot` input in the main HydroDyn input file.  This mesh reference orientation should have been set to the identity.

@rdamiani found this bug in testing a proprietary floating platform design including both Morison and potential flow elements.  He had noticed that rotating the `WAMIT` body produced a different period in a pitch free decay test.  It is expected that this frequency would be identical for an axisymmetric `WAMIT` body.  See #914 for plots.

For further testing, a very simple test using a purely potential flow (`WAMIT` body) floating platform from the ITI Barge test case (`5MW_ITIBarge_DLL_WTurb_WavesIrr` regression test).  Turning off waves and wind, a free decay test from -5 degrees initial pitch is performed.

Before fix:
Screen Shot 2021-11-10 at 2 08 33 PM

After bug fix:
Screen Shot 2021-11-10 at 2 12 34 PM

#914 

Potential flow (WAMIT) bodies in HydroDyn that are rotated will be affected.  No regression tests currently exist using this feature.

### DIFF
--- a/modules/hydrodyn/src/HydroDyn.f90
+++ b/modules/hydrodyn/src/HydroDyn.f90
@@ -1584,7 +1584,7 @@ SUBROUTINE HydroDyn_Init( InitInp, u, p, x, xd, z, OtherState, y, m, Interval, I
             CALL SetErrStat(ErrStat2,ErrMsg2,ErrStat,ErrMsg,RoutineName)
 
          do iBody = 1, p%NBody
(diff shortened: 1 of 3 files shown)


andrew-platt:
Closes #914

jjonkman (approved):
I talked through this @andrew-platt and agree with this fix.

PR #919, opened 2021-11-12
Segment treecode

rcorniglion:

Adds a segment-based treecode algorithm based on the current particle-based treecode algorithm.

It removes the need to choose a number of approximating particles:
- the moments in the treecode algorithm are computed from the existing vortex lines
- the direct evaluations of induced velocity are done on the existing vortex line rather than approximating particles

A new option is added for VelocityMethod in the FVW input file. The segment treecode is VelocityMethod=4

The method gives results equivalent to the existing treecode using several particles but has a computational load similar to the existing treecode with only one particle. It improves the accuracy compared to the one particle method while it has a similar computational cost. The computational cost is slightly higher than the one particle method due to the more demanding moment computation.

OLAF in AeroDyn, documentation

**Tasklist**
 - [x] Update documentation
 - [x] Update r-test

**Tests**
Switching from regular segment to tree segment in the Quad-Rotor tests case had the following change in result (negligible) requiring to change the baselines. 
image

### DIFF
--- a/docs/source/user/aerodyn-olaf/ExampleFiles/ExampleFile--OLAF.dat
+++ b/docs/source/user/aerodyn-olaf/ExampleFiles/ExampleFile--OLAF.dat
@@ -30,7 +30,7 @@ False   FWShedVorticity    Include shed vorticity in the far wake {default: fals
 False   TwrShadowOnWake    Include tower flow disturbance effects on wake convection {default:false} [only if TwrPotent or TwrShadow]
 0       ShearModel         Shear Model {0: No treatment, 1: Mirrored vorticity, default: 0}
 ------------------- SPEEDUP OPTIONS -----------------------------------------------------------
-2       VelocityMethod     Method to determine the velocity {1:Biot-Savart Segment, 2:Particle tree, default: 1}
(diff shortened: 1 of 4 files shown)


ebranlard:
Thank you Remi for sharing this! We've discussed this together a couple of months ago, and this is a great contribution thank you for doing this. 

Can you update the documentation to mention this change?  You can edit:
https://github.com/rcorniglion/openfast/blob/segment-treecode/docs/source/user/aerodyn-olaf/InputFiles.rst#speedup-options  
and also change the example input file:
https://github.com/rcorniglion/openfast/blob/segment-treecode/docs/source/user/aerodyn-olaf/ExampleFiles/ExampleFile--OLAF.dat

For the r-test:
I would suggest to use your new segment tree for the QuadRotor test case. (https://github.com/OpenFAST/r-test/blob/dev/modules/aerodyn/ad_QuadRotor_OLAF/OLAF.dat). No need to add a new test case, but we might need to update the baseline results. We will also need to change all the OLAF input files of the repository to indicate the new option in the description. 

Let me know if you need help for any of these. If you don't have the time to do it, you can give me access to your repo and I'll take care of it. 

Emmanuel

ebranlard:
I've talked with Remi and took care of the remaining actions.  This pull request is now ready to be merged.

ebranlard (approved):

PR #920, opened 2021-11-14
Update in Computing Default Unsteady Airfoil Coefficients

bjonkman:

In addition to some minor code cleanup (comments removed, hard-coded version numbers and dates removed), this pull request fixes some issues with computing the default UA parameters in the airfoil files that can occur with "unusual" polars (particularly those polars that aren't technically linear in the supposed linear region). The documentation of those parameters has also been slightly modified, and a new section on how the default values are calculated was added..

UA's default airfoil parameters, documentation 

This should not change any test results because the default UA-AFI parameters aren't used in any current tests.


bjonkman:
@ebranlard, I added some documentation on how the default values are computed, if you want to review that.

docs/source/user/aerodyn/theory_ua.rst
```diff
@@ -465,13 +487,79 @@ Inputs
...
+
+After computing ``cd0``, the :math:`C_n` curve is computed by
+
+.. math::
+       C_{n}(\alpha) = C_l(\alpha) \cos\alpha + \left(C_d(\alpha) - c_{d0}\right) \sin\alpha
+
+The slope of the :math:`C_n` curve is computed as follows:
+   
+.. math::
+       C_{n}^{Slope}\left(\frac{\alpha_{i+1} + \alpha_i}{2}\right) = \frac{C_n(\alpha_{i+1}) - C_n(\alpha_i)}{\alpha_{i+1} - \alpha_i}
```
ebranlard, line 514: Is this computed for a range of angle of attack and then few indices? How is this range determined? I'm guessing it's +/- 20 deg?

docs/source/user/aerodyn/theory_ua.rst
```diff
@@ -465,13 +487,79 @@ Inputs
...
+``Cn1`` is the value of :math:`C_n(\alpha)` at the smallest value of :math:`\alpha` where :math:`\alpha >= \alpha^{Upper}` and the separation function, :math:`f_{st}(\alpha)` = 0.7.
+
+``Cn2`` is the value of :math:`C_n(\alpha)` at the largest value of :math:`\alpha` where :math:`\alpha <= \alpha^{Lower}` and the separation function, :math:`f_{st}(\alpha)`  = 0.7.
+  
+``Cn_offset`` is the average value of the :math:`C_n` curve at ``alphaUpper`` and ``alphaLower``:
+
+.. math::
+       C_{n}^{offset} = \frac{C_n\left(\alpha^{Lower}\right) + C_n\left(\alpha^{Upper}\right)}{2}
+
+``C_nalpha`` is defined as the maximum slope of the smoothed :math:`C_n` curve, :math:`C_{n,smooth}^{Slope}` between :math:`\pm20` degrees angle of attack.
```
ebranlard, line 545: From the definition of Cnalpha or Clalpha it seems the local slope is used (between i and i+1). I wonder if it would be better to use the slope between i and i0 where i0 if the index of alpha0. That the definition used in the Riso-R-1354 report equation 16.

ebranlard (approved):

ebranlard:
Thanks for implementing this. It looks good to me. I have some small comments on the calculation of the slope.

docs/source/user/aerodyn/theory_ua.rst
bjonkman: The slope is computed for NumAlf-1 points (i.e., between all consecutive AoA specified in the polar). The smoothed values of this curve can have some weird behavior at the ends, but `alphaUpper` and `alphaLower` are far enough from the ends that it shouldn't matter.

docs/source/user/aerodyn/theory_ua.rst
bjonkman: We could change that. I was trying to make it somewhat consistent with the method for computing `C_lalpha`, but it's used in different UA models, so could certainly be done differently.

PR #922, opened 2021-11-17
[bugfix] Minor bugfix in AirfoilInfo

rcorniglion:
fullySeparate was not used at the right place and could be used when not defined.

Ready to be merged.


andrew-platt (approved):
@rcorniglion Good catch!  That is definitely a bug.  Thanks for contributing!

PR #931, opened 2021-12-01
Implementation of the curled-wake model in FAST.Farm

ebranlard:
Not ready to merge

Implement the curled wake model into FAST.Farm
- Uses a Cartesian grid as output of WakeDynamics and within AWAE. for both polar and curled wake 
- UpdateStates of WakeDynamics is modified to accommodate curled wake using if statements throughout
- Outputs of WakeDynamics are in cartesian coordinates and on cartesian grid
- FAST.Wrapper returns Cq, psi_skew and chi_skew 

FAST.Farm, AWAE, WakeDynamics

- Original curled wake paper: [here](https://wes.copernicus.org/articles/6/555/2021/)
- Paper regarding this implementation: [here](https://doi.org/10.1002/we.2785)

We expect small changes in the r-test sine we don't use the filtered velocity anymore to compute dx and convect, and beasue we use a Cartesian grid to communicate with AWAE. 

**Checklist**
- [x] Document input file changes. Add link to paper. [Live documentation](https://ebranlard-openfast.readthedocs.io/en/f-ff-curled-wake/source/user/fast.farm/index.html)
- [x] Review the filtering of skew angle and azimuth angle
of it)
- [x] Set default value of f_c based on the a=1/3, Uinf = 8, and a rotor radius based on the wake plane radius 
- [x] Finalize implementation of plane outputs (- base name of VTK files, - Comment out the old wake-based write outputs for the velocity deficits and eddy viscosity values, - Ensure that the VTK-based output for the velocity deficits and eddy viscosity values also works for polar) 
- [x] Additional turbine outputs (filtered skew, psi, gamma)
- [x] Document outputs
- [x] Add OpenMP parallelization of AWAE High Res
- [x] Add OpenMP parallelization of WD update states (and perhaps calculate output?)
- [x] Add an r-test for the curled wake  (based on the yaw steering test case with one of the existing turbulent wind data files)


bjonkman:
FYI: at the moment it looks like some of the FAST.Farm source code in this PR has white-space changes (different line endings?) on every line. For example, see FAST_Farm_Subs.f90.

ebranlard:
Damned, I just saw that, I think it's just in the last commit. I'm not sure how it happened and how I can easily reverse that, let me know if you have a good hint. I'll try to revert and force push this commit.

ebranlard:
Fixed... I used `set ff=unix` in vim on those files. I might have mingled with `git config --global core.autocrlf true` before this problematic commit, but that was to actually try to avoid this issue as I could see vim not being happy with the file endings... I'll keep an eye out in my next commits to make sure it doesn't happen again. Thanks for spotting it!

ebranlard:
Change of convection velocity slightly affect the test results (for LESInflow): 
image

image

Test cases will be rebaselined

Issue #799, opened 2021-07-30
Bug report


pelljam:
Possible bug with AeroDyn UA

There are a couple of fields of `UA_DiscreteStateType` that, to my nonexpert eye, looked like they might not be getting updated correctly. The one I am wondering about are: `K3prime_q_minus1` and `Kprimeprime_q_minus1`

I expected that they might be updated in `UA_UpdateDiscOtherState`, i.e. something like is done for what looks like a very similar field:

`xd%Kprime_q_minus1(i,j)  = KC%Kprime_q`

But I really don't have a good handle on this stuff, so there is a very good chance I've got this wrong and so apologies if I've wasted your time here!

ebranlard:
Hi @pelljam, 

I had a quick look at the code, and I agree with you, the two variables you mentioned `K3prime_q_minus1` and `Kprimeprime_q_minus1` don't seem to be  updated, and I suspect they might remain zero. 

If they do represent a quantity at the previous time step, they should be updated as you mentioned, as: 
```
xd%K3prime_q_minus1(i,j) = KC%K3prime_q
xd%Kprimeprime_q_minus1(i,j) = KC%Kprimeprime_q
```

It seems though that `K3prime_q` and `Kprimeprime_q` are not stored in KC and might need to be added there. 

I'll have to look at the unsteady aero manual to see if these variables indeed need updating or if they are simply also based on a "zero" previous value. Have you looked into it?

pelljam:
Hi @ebranlard 

No I have not looked at the theory in detail. I'm just going off how these fields are used in the `ComputeKelvinChain` subroutine, i.e. in a way that looks like they require the state to be tracked, for example:

`K3prime_q       = Get_ExpEqn( BL_p%b5*beta_M_Sqrd*KC%ds, 1.0_ReKi, xd%K3prime_q_minus1(i,j),  BL_p%A5*(KC%q_f_cur - q_f_minus1), 0.0_ReKi )`

But then as you say, the _current_ value is assigned to a local var and not a field of `KC`. If the intent is for `K3prime_q_minus1` to always be zero, this would be a strange choice for the variable name (however there could be some legacy reason for that).

However, I'm far from an expert so all I can really point out is that it looked a little strange to me and it might need some further consideration.

Sorry I can't be more helpful.

James

pelljam:
Hi again @ebranlard,

I wondered if there had been any progress with resolving this issue?

Thanks, James

ebranlard:
[...]

PR #932, opened 2021-12-02
UA: some states were never updated

ebranlard:
Apparent bug in UA where states are never updated. 

#799

UA (AeroDyn)

**Tests**
Tests using AFAeroMod=2 and UAMod=2,3 are expected to fail: 
- ad_Kite_OLAF
- OpenFAST r-tests.


rafmudaf:
The difference in `ad_Kite_OLAF` is very small

Screen Shot 2022-04-20 at 6 16 26 PM

ebranlard:
Yes, I'd recommend updating the baselines as I do think this is a bug (having a small impact though).

PR #982, opened 2022-01-25
[BugFix] AD15 nacelle reference position was set to hub position

andrew-platt:
This PR **is** ready for merging

The AeroDyn 15 nacelle reference position was set to the hub reference position.

The effect of this is likely very small since mesh mapping at the glue code accounts for any location differences, and aero loads on the nacelle from AD15 would be nearly identical calculated at the correct nacelle location or at the hub.  However, buoyancy loads on the nacelle were applied at the hub rather than nacelle location, but mapped to the nacelle in the glue code which would result in an unexpected moment.

#957 

This likely only affects buoyancy loads on the nacelle the MHK feature under development.

### DIFF
--- a/modules/aerodyn/src/AeroDyn.f90
+++ b/modules/aerodyn/src/AeroDyn.f90
@@ -1031,7 +1031,7 @@ subroutine Init_u( u, p, p_AD, InputFileData, InitInp, errStat, errMsg )
       if (errStat >= AbortErrLev) return
             
          ! set node initial position/orientation
-      position = InitInp%HubPosition
+      position = InitInp%NacellePosition
       position(1:2) = 0
       call MeshPositionNode(u%NacelleMotion, 1, position, errStat2, errMsg2, orient=InitInp%NacelleOrientation)
          call SetErrStat( errStat2, errMsg2, errStat, errMsg, RoutineName )
--- a/modules/aerodyn/src/AeroDyn_Registry.txt
+++ b/modules/aerodyn/src/AeroDyn_Registry.txt
@@ -42,6 +42,7 @@ typedef	^	RotInitInputType	ReKi	HubPosition	{3}	-	-	"X-Y-Z reference position of
 typedef	^	RotInitInputType	R8Ki	HubOrientation	{3}{3}	-	-	"DCM reference orientation of hub"	-
 typedef	^	RotInitInputType	ReKi	BladeRootPosition	{:}{:}	-	-	"X-Y-Z reference position of each blade root (3 x NumBlades)"	m
 typedef	^	RotInitInputType	R8Ki	BladeRootOrientation	{:}{:}{:}	-	-	"DCM reference orientation of blade roots (3x3 x NumBlades)"	-
+typedef	^	RotInitInputType	R8Ki	NacellePosition	{3}	-	-	"X-Y-Z reference position of hub"	m
 typedef	^	RotInitInputType	R8Ki	NacelleOrientation	{3}{3}	-	-	"DCM reference orientation of nacelle"	-
 typedef	^	RotInitInputType	IntKi	AeroProjMod   	-	0	-	"Flag to switch between different projection models"	-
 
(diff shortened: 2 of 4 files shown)


andrew-platt:
@bjonkman, could you review this?

jjonkman (approved):
Makes sense to me.

modules/aerodyn/src/AeroDyn.f90
```diff
@@ -1031,7 +1031,7 @@ subroutine Init_u( u, p, p_AD, InputFileData, InitInp, errStat, errMsg )
       if (errStat >= AbortErrLev) return
             
          ! set node initial position/orientation
-      position = InitInp%HubPosition
+      position = InitInp%NacellePosition
       position(1:2) = 0
```
bjonkman, line 1035: This mesh was originally created to transfer nacelle drag loads, which are calculated at the tower center line at hub height with the orientation of the nacelle. If you want this to be the actual nacelle position, you should probably also remove the line setting position(1:2)=0.

andrew-platt: Thanks!  I'll do that.

ebranlard: It would be great to add the drag loads at some point though. Should this mesh have two points then ? A nacelle center and an aerodynamic center?

andrew-platt: It seems like the aero center should be handled within AD15.  So I would expect a single mesh point is necessary for the loads to ED.

bjonkman: Maybe two points on the mesh would be better.... I feel like we had a conversation with Hannah about this when I first sent her the code to add the nacelle mesh to AeroDyn. Did she alter her equations to be calculated at this different point???

andrew-platt: It looks like Hannah is referencing against the "nacelle reference point", which is given as the tower top in the documentation.  There are some offsets given somewhere center of mass offset that are then used for the resultant buoyancy forces: https://github.com/OpenFAST/openfast/pull/957/files#diff-dd7b98a4c7eb64ad7503601ab8789a1861273ede4896c0c161a447e05549d8b6R2037

bjonkman: I do recall having a conversation with Jason and Hannah, discussing if we should have multiple points on that mesh or change the reference position, and I feel like they accounted for the difference in position in her equations in AD. But, I'd have to defer to her regarding what was actually implemented and pushed to OpenFAST.

hkross: Just saw this. The buoyancy calculations for the nacelle assume a nacelle reference position aligned with the tower centerline and vertically with the hub, which is how the nacelle position was defined previously. I assume this means I need to adjust the buoyancy calculation now? Where is the new nacelle reference position? @andrew-platt

andrew-platt: The nacelle reference position from the ElastoDyn `NacelleLoads` and `NacelleMotion` mesh is located at `(0,0,TwrHt)`.  This is what is passed to AD15 for the Nacelle initial position.  So this should work as expected with the MHK buoyancy updates.

hkross: Ok, I see. I think the key thing is to make sure it is clear in the documentation that the nacelle center of buoyancy input by the user should be relative to the tower top position. I believe Jason already suggested that "nacelle reference position" be changed to "yaw bearing", so that should take care of it.

We do have a calculation to move the correction for the tower top (since it is not exposed to fluid) from the center of buoyancy of the tower top node to the nacelle reference position. Since the tower center of buoyancy is assumed along the centerline and we are now defining the nacelle reference position as `(0,0,TwrHt)`, these points should be coincident, so I could remove this calculation. Does this sound correct, @jjonkman? I could also leave since it is more robust to future changes this way.

jjonkman: @hkross -- Yes, that sounds correct, but may as well leave that code in there in case the nacelle reference position is moved again.

hkross: I came across another potential issue with this. The AeroDyn driver does not pass the nacelle position to AeroDyn. Since the mesh creation in AeroDyn used to call `InitInp%HubPosition`, the nacelle position was not needed. This is fairly easy to fix and pass in the nacelle position. However, the nacelle position in the driver is set to `(0, 0, HubHt+overhang*sin(shftTilt))` rather than `(0,0,TwrHt)`, meaning that the nacelle position is defined differently when the AeroDyn driver vs OpenFAST driver is used. This will affect the buoyancy calculations, since the user inputs offsets for the nacelle center of buoyancy relative to the tower top. I'm not sure there is even enough info in the AeroDyn driver input file to be able to calculate `TwrHt`. @andrew-platt and @ebranlard, do either of you have any input on this?

Issue #248, opened 2019-02-21
Possible Bug in BeamDyn -- Root Pitching Moment Doesn't Match Pitching Moment from First Node?


jjonkman:
In a topic on the forum related to BeamDyn, Ryan Walker has raised concerns that the pitching moment (Mz) at the root doesn't match the pitching moment at the first node, when the first node is located at the root.  I have not confirmed this result myself.  Perhaps the issue has been fixed in the yet-to-be merged pull request #163?

Ryan has also reported that Figures 6.7, 6.8, and 6.9 do not show up in the readthedocs user documentation for BeamDyn.

See the following topic on the forum for more information: https://wind.nrel.gov/forum/wind/viewtopic.php?f=4&t=2215.

rafmudaf:
@jjonkman a somewhat similar problem was fixed here: https://github.com/OpenFAST/openfast/pull/114. However, that was merged into OpenFAST prior to the commit where these problems are reported: v1.0.0-198-gb31cfbad-dirty.

jjonkman:
@rafmudaf -- Yes, I agree, which is why I was surprised Ryan Walker raised the issue again (you'll notice in the forum that my first question to Ryan was which OpenFAST version was being used).

rafmudaf:
It was also mentioned that he started with `v2.0.0-25-g50c88c2f-dirty` then at one point modified some BeamDyn files from FAST 8 to OpenFAST (no version given) while the attached output files are at `v1.0.0-198-gb31cfbad-dirty`. I'm not implying that any of this is the cause of the problem, but just documenting for future reference. Both of the OpenFAST commits here are after the pull request which fixed a similar issue.

vparmar58:
Hi @jjonkman @rafmudaf ,
I work with Ryan and conducted a test with NREL input files. I ran simulation using Test18 from the input files provided with FAST 8 package. I had to make some changes in the input files to make them compatible with the openfast. I have attached the input and out files that I ran using "openfast v1.0.0-198-gb31cfbad-dirty"
The simulation used an 8 m/s steady wind, no controllers used, and no grid connection. All the blades had same beamdyn input file.  From the output file, I noticed two issues:
1. The Mz loads are different for B1RootMzr and B1N1Mzl
2. The Mz and My loads are also different for blade 1 compared to blade 2 and 3 at node 1. Please see the files attached. 
It would be really helpful to understand disparity between My and Mz loads at node 1 between the blades.
Thanks.

Best,
Vismay.

[...]

PR #996, opened 2022-02-10
[BugFix] BeamDyn nodal outputs occasionally segfaulted

andrew-platt:
**Ready for merging**

BeamDyn would occasionally incur a segmentation fault when nodal outputs were requested. This appeared to happen only with certain Intel compiler version on Windows and was difficult to reproduce reliably.  The root cause was the first entry in `NdIndxInverse` index lookup table not getting set (this index table is used to map the quadrature point node to the correct nodal output channel).  Thus the first index value would be whatever was resident in that memory location and might be a value outside the range of the output nodes which would result in a segmentation fault.

For whatever reason, I could only reproduce this error with some Intel compiles on Windows.  It did not show up on Linux or with the gcc compilers.  Maybe this is due to different memory layout or previous stored values in the particular memory space (hard to tell without spending more time on the issue).

This issue only affected the indexing for the first node, so may in some instances have caused incorrect outputs for the first node nodal outputs to not match the root outputs.

Maybe related: #248 
Noted in forum post: https://forums.nrel.gov/t/sending-rtaeropwr-via-servodyn-to-bladed-style-dll-controller/2043/14

BeamDyn node based outputs are the only bit of code affected.

### DIFF
--- a/modules/beamdyn/src/BeamDyn.f90
+++ b/modules/beamdyn/src/BeamDyn.f90
@@ -1062,6 +1062,7 @@ subroutine SetParameters(InitInp, InputFileData, p, ErrStat, ErrMsg)
          if (ErrStat >= AbortErrLev) return
 
       p%NdIndx(1) = 1
+      p%NdIndxInverse(1) = 1
       p%OutNd2NdElem(:,1) = 1 ! note this is an array
       indx = 2
       DO i=1,p%elem_total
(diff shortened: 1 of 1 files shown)


jjonkman (approved):
I'm not familiar with the details of the code here, but uninitialized variables can certainly cause problems.

PR #999, opened 2022-02-11
Fix HD added mass on member end (Close #992)

### LINKED ISSUE #992: HydroDyn added mass at end of member may not computed properly
**Bug description**
The variable `v2D` in the line below is (3x1): 
https://github.com/OpenFAST/openfast/blob/dev/modules/hydrodyn/src/Morison.f90#L2125
Therefore,  the expression `matmul(transpose(v2D), v2D)`  returns a scalar. 

I believe the equation should be replaced by `matmul( (v2d, transpose(v2d)  )` to return a 3x3 matrix. 

In my windows gfortran value, the statement is interpreted in a weird way, and the scalar value is placed in the element (3,3) of `AM_End`, which is quite surprising. 

I'll submit a pull request. 

**OpenFAST Version**
Openfast dev as of 2/7/2022

ebranlard:
This pull request is ready to be merged (after r-test update). 

#992

HydroDyn

One HD test now fails (hd_oc3_tripod)
Main differences is in HydroFzi, but also visible in HydroFxi and HydroMyi. 
image

- [ ] Update r-test


jjonkman (commented):
I agree with this change.  The implementation plan shows a V*V^T equation anyway.

I'm wondering, though, if there is another term missing here?  The equation for Vn appear to be missing the 2/3*Pi coefficient out front that is called out in the implementation plan.

ebranlard:
It seems we agreed that the 2/3 pi coefficient is indeed missing, I can add this to the pull request #999

jjonkman (approved):
Looks good.  Thanks for fixing this!

rafmudaf:
@ebranlard Are you able to update the HydroDyn module test baselines here?

bjonkman:
Could we get this bug fix merged into dev soon?

Issue #926, opened 2021-11-24
Help to install OpenFAST on macbook


Londu:
Hi,
I have just started my Master thesis and I need to install the OpenFAST. My supervisor suggested me to compile it with MATLAB as he is using the same. 
I am using macbook with MacOS Big-Sur. From "https://openfast.readthedocs.io" I used the command to install the Homebrew , install OpenFAST and later test it ( till 2.1.2 on openfast.readthedocs.io). 
At next step- to compile OpenFAST  I see instructions only to compile with Visual Studio. But I need to do it with MATLAB. 

I would appreciate any help on this. Thank you in advance. 

Best regards

whisperingkahuna:
If you dont want to make changes in OpenFast, you can simply use the precompiled binary files provided in the Assets section


![image](https://user-images.githubusercontent.com/28399678/149652260-ae02157e-3a46-48b2-b3b2-f0758effcbc3.png)


Or refer to this pdf shared by someone on this forum a while ago for compiling Openfast. 

[OpenFAST_Simulink.pdf](https://github.com/OpenFAST/openfast/files/7876573/OpenFAST_Simulink.pdf)

andrew-platt:
Closing as stale.

Issue #924, opened 2021-11-19
Bug report


isablanca:
Dear all,

I have encountered some error when running openfast with simulink. My system os is Ubuntu18.04, matlab version is R2021b (It can't work in R2020b either). I have tried two ways to install the openfast.

Method1:

I follow the document, compile openfast use

```
cmake .. -DBLAS_LIBRARIES=/usr/lib/x86_64-linux-gnu/openblas/libblas.so -DLAPACK_LIBRARIES=/usr/lib/x86_64-linux-gnu/openblas/liblapack.so -DBUILD_SHARED_LIBS=ON -DBUILD_OPENFAST_SIMULINK_API=ON
make -j
make install
```

Then, create_FAST_SFunc.m runs successfully.

But, when I  run Run_OpenLoop.m, the matlab crashed, with following message,

```
>> Run_OpenLoop

 **************************************************************************************************
 OpenFAST

 Copyright (C) 2021 National Renewable Energy Laboratory
 Copyright (C) 2021 Envision Energy USA LTD

 This program is licensed under Apache License Version 2.0 and comes with ABSOLUTELY NO WARRANTY.
 See the "LICENSE" file distributed with this software for details.
 **************************************************************************************************

 OpenFAST-v3.0.0-1-gebff1b46-dirty
 Compile Info:
  - Compiler: GCC version 7.5.0
  - Architecture: 64 bit
  - Precision: double
  - OpenMP: No
  - Date: Nov 19 2021
  - Time: 12:21:21
 Execution Info:
  - Date: 11/19/2021
  - Time: 12:56:10+0100

 OpenFAST input file heading:
     FAST Certification Test #06: AOC 15/50 with many DOFs with gen start, loss of grid, and
     tip-brake shutdown. Many parameters are pure fiction.

 Running ElastoDyn.
 Nodal outputs section of ElastoDyn input file not found or improperly formatted.
 Running AeroDyn14.
   WARNING:   AeroDyn 14 is deprecated and will be removed in a future release.
 Running InflowWind.
 Running ServoDyn.

 FAST_InitializeAll: IfW_Init:IfW_UniformWind_Init: Could not read upflow column in uniform wind
 files. Assuming upflow is 0.
 FAST_InitializeAll:SrvD_Init:ValidatePrimaryData:Yaw angle and rate are not commanded from
 Simulink model.
 ValidatePrimaryData:Pitch angles are not commanded from Simulink model.
[...]

Issue #556, opened 2020-10-05
Compiling with Simulink API: Matlab *.so undefined reference to  '***** @GLIBCXX_3.4.21''


Seager1989:
Hi everyone,

I am trying to compile the OpenFAST with the Simulink API using the dev branch of the latest version on an HPC cluster with the system information as follows. The MATLAB 2020a is installed. I use the manually compiled  GCC-6.3.0 since this is the version supported by the MATLAB 2020a.

Operating System: CentOS Linux 7 (Core)
CPE OS Name: cpe:/o:centos:centos:7
Kernel: Linux 3.10.0-1062.9.1.el7.x86_64
Architecture: x86-64

After cloning the OpenFAST, I did run the following comman.
```
mkdir build
cd build
cmake .. -DBUILD_OPENFAST_SIMULINK_API=ON -DBUILD_SHARED_LIBS=ON -DCMAKE_INSTALL_PREFIX="/home/xd101/local/openfast_sim" -DBUILD_TESTING=ON
make -j 4
```

However, error appeared with key information listed below:
```
/cache/sw/packages/MATLAB/R2020a/bin/glnxa64/libmex.so: undefined reference to `VTT for std::__cxx11::basic_ostringstream<char, std::char_traits<char>, std::allocator<char> >@GLIBCXX_3.4.21'
/cache/sw/packages/MATLAB/R2020a/bin/glnxa64/libmwmlutil.so: undefined reference to `std::__cxx11::basic_stringstream<char, std::char_traits<char>, std::allocator<char> >::basic_stringstream(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::_Ios_Openmode)@GLIBCXX_3.4.21'
/cache/sw/packages/MATLAB/R2020a/bin/glnxa64/libmwboost_log.so.1.70.0: undefined reference to `typeinfo for std::codecvt<char16_t, char, __mbstate_t>@GLIBCXX_3.4.21'
/cache/sw/packages/MATLAB/R2020a/bin/glnxa64/libmwfl.so: undefined reference to `std::ios_base::failure[abi:cxx11]::~failure()@GLIBCXX_3.4.21'
[...]

Issue #482, opened 2020-06-30
Undefined references when creating S-Function


brendansndrs:
I am trying to use OpenFAST with Simulink, however, I get errors when running `create_FAST_SFunc.m`. Some of the error message is below (the rest is similar, but for different functions).

    >> create_FAST_SFunc
    
    ----------------------------
    Creating ./FAST_SFunc.mexa64
    
    Building with 'gcc'.
    Error using mex
    /pathToOpenFAST/openfast/install/lib/libopenfastlib.a(FAST_Library.f90.o): In function `FAST_AllocateTurbines':
    FAST_Library.f90:(.text+0x48): undefined reference to `nwtc_io_mp_wrscr1_'
    FAST_Library.f90:(.text+0x59): undefined reference to `nwtc_io_mp_wrscr1_'
    FAST_Library.f90:(.text+0xf2): undefined reference to `for_check_mult_overflow64'
    FAST_Library.f90:(.text+0x166): undefined reference to `for_alloc_allocatable'
    FAST_Library.f90:(.text+0x197): undefined reference to `_intel_fast_memcpy'
    FAST_Library.f90:(.text+0x1d3): undefined reference to `_intel_fast_memset'
    /pathToOpenFAST/openfast/install/lib/libopenfastlib.a(FAST_Library.f90.o): In function `FAST_DeallocateTurbines':
    FAST_Library.f90:(.text+0xf172): undefined reference to `for_deallocate_all'
    FAST_Library.f90:(.text+0xf19d): undefined reference to `_intel_fast_memset'
    /pathToOpenFAST/openfast/install/lib/libopenfastlib.a(FAST_Library.f90.o): In function `FAST_Sizes':
    FAST_Library.f90:(.text+0xf1ef): undefined reference to `_intel_fast_memcpy'
    FAST_Library.f90:(.text+0xf21f): undefined reference to `_intel_fast_memmove'
    FAST_Library.f90:(.text+0xf233): undefined reference to `_intel_fast_memmove'
    FAST_Library.f90:(.text+0xf25d): undefined reference to `for_f90_index'
    FAST_Library.f90:(.text+0xf286): undefined reference to `for_cpystr'
    FAST_Library.f90:(.text+0xf2a1): undefined reference to `__dtoq'
    FAST_Library.f90:(.text+0xf3b5): undefined reference to `fast_subs_mp_fast_initializeall_t_'
[...]

PR #1010, opened 2022-02-22
Use Cmake OBJECT libraries to create openfastlib and add option to use downloaded reference lapack and blas sources

reos-rcrozier:

Updates to the build system to facilitate building on Linux, particularly with Simulink support. OpenFAST cmake files have been modified to properly make use of the cmake OBJECT library type to create the openfast library. In addition, the option to build a local version of the official LAPACK and BLAS sources has been added through the use of the Cmake command [ExternalProject_Add](https://cmake.org/cmake/help/latest/module/ExternalProject.html#command:externalproject_add). A new boolean variable has been added, `USE_LOCAL_STATIC_LAPACK`. If set to `ON`, the latest LAPACK and BLAS soures are downloaded and build locally. The install command always installs these locally build libraries in `${CMAKE_SOURCE_DIR}/install/lib`, i.e. regardless of the value of `CMAKE_INSTALL_PREFIX` to avoid user's accidentally overwriting system LAPACK and BLAS libraries. The purpose of this option is primarily to provide static LAPACK libraries to link to the FAST_Sfunc mex file, to avoid load time issues with shared libraries, but it will also allow testing with the latest LAPACK.

#924, #556, #482, #926, probably others

Only cmake files and `create_FAST_SFunc.m`

When building on Linux with Simulink support, there is an issue at load time with libraries being shipped with Matlab incorrectly being loaded when loading openfast as a shared library. These include libraries such as lapack and blas. A solution might be to build the openfast library as a static library, however, the current method of doing that on Linux is broken because the openfast library is created by linking together a bunch of other libraries created during the build. This is ok for shared libraries, but is incorrect on linux for static libraries. On Linux, the linker when creating a static library is `ar`. `ar` is just an archiver and doesn't really do a lot more than take a bunch of object files and stuff them in the static library (at least by default with Cmake). `ar` cannot link an existing library to the static library, one would have to extract the object files from the static library and then use `ar` to add them. With the current build system, when you try to link to the created static library on Linux, you get undefined references, because `ar` hasn't actually linked any of the sub-libraries like `aerodyn` etc.

Cmake have provided a mechanism to deal with this, which is [OBJECT libraries](https://gitlab.kitware.com/cmake/community/-/wikis/doc/tutorials/Object-Library). OBJECT libraries allow you to break up libraries into smaller modules as in OpenFAST, but link them correctly at the higher level. With this PR we have modified OpenFAST to use object libraries throughout.

This solves the issue with undefined references due to the missing OpenFAST modules, but, it also means that when later linking to `libopenfastlib.a` one must remember to link to the `gfortran`, `quadmath`, `dl`,  `lapack` and `blas` libraries to avoid undefined references from these libraries (which are linked by default if building a shared library). This presents a further problem when linking a mex file on Linux as the Matlab supplied LAPACK and BLAS shared libraries are not the same version as the versions which are linked by default by the `mex` command. For shared libraries the problem is not apparent at link time, but appears at load time when differences in the interfaces cause errors when running the `FAST_SFunc` command. The solution is to link to static LAPACK and BLAS libraries when creating the mex file. For this reason the `USE_LOCAL_STATIC_LAPACK` options described above was created. This builds the LAPACK and BLAS libraries locally, and installs them in a place which is easy for the mex linker to find, in the OpenFAST source tree. 

Some basic modifications to `create_FAST_SFunc.m` to make use of these modification have been made. On non-windows platforms it links to `gfortran`, `lapack`, `blas`, `quadmath` and `dt` by default to avoid undefined references. The default install directory is also modified to `../../install` as on windows, mainly to allow easy linking to static LAPACK installed in these directories. Therefore on linux, an appropriate cmake command is something like:

`
cmake -LH /home/username/source/path/openfast -DBUILD_FASTFARM=ON -DBUILD_OPENFAST_CPP_API=ON -DBUILD_OPENFAST_SIMULINK_API=ON -DBUILD_SHARED_LIBS=OFF -DCMAKE_INSTALL_PREFIX=/home/username/source/path/openfast/install -DDOUBLE_PRECISION=ON -DOPENMP=OFF -DUSE_LOCAL_STATIC_LAPACK=ON
`


reos-rcrozier:
FYI a backport of these modification to OpenFAST v3.0.0 is available on our branch [here]( https://github.com/REOptimize-Systems/openfast/blob/backport-build-changes_to_v3.0.0/glue-codes/simulink/src/create_FAST_SFunc.m)

reos-rcrozier:
looking at the output of the runs, it seems we might have missed specifying  a target dependency for versioninfolib_obj. Probably being picked up because the order of compilation is slightly different than on our test machine.

CMakeLists.txt
```diff
@@ -19,9 +19,12 @@ project(OpenFAST CXX C Fortran)
 
 include(${CMAKE_SOURCE_DIR}/cmake/OpenfastCmakeUtils.cmake)
 include(${CMAKE_SOURCE_DIR}/cmake/OpenfastFortranOptions.cmake)
+include (FindPackageHandleStandardArgs)
 
 set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
 
+set(CMAKE_Fortran_MODULE_DIRECTORY ${CMAKE_SOURCE_DIR}/build/ftnmods)
```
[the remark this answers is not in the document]
reos-rcrozier: I also realised my comment was wrong when I went to examine this. I'm not yet sure what the issue really is, but I think it might be something to do with the fact that the cmake files now separately specify the target libraries and object libraries. Possibly it need to be changed so the target 'real' libraries all use the object libraries, so cmake doesn't attmept to build the same object twice.

modules/aerodyn/CMakeLists.txt
```diff
@@ -111,7 +115,7 @@ set(UA_DRIVER_SOURCES
 add_executable(unsteadyaero_driver ${UA_DRIVER_SOURCES})
 target_link_libraries(unsteadyaero_driver aerodynlib fvwlib uaaerolib afinfolib nwtclibs versioninfolib ${CMAKE_DL_LIBS})
 
-install(TARGETS unsteadyaero_driver aerodyn_driver aerodynlib fvwlib uaaerolib afinfolib aeroacoustics
+install(TARGETS unsteadyaero_driver aerodyn_driver aerodynlib fvwlib uaaerolib afinfolib aeroacoustics aerodynlib_obj
```
rafmudaf, line 118: Do we need to install the new objects? As I understand it, they're only used at compile-time to create the new `openfast_prelib_obj`.

rafmudaf:
@reos-rcrozier Thanks for this pull request and my apologies that it has sat for a few months. I think I understand the motivation and how the code changes solve the problem.
Have you worked through the issue with the version lib?

modules/aerodyn/CMakeLists.txt
reos-rcrozier: Yes, intuitively they have to be in the install target to be used in other CMAKE libraries, but they are not actually installed like real libraries are.

reos-rcrozier:
> @reos-rcrozier Thanks for this pull request and my apologies that it has sat for a few months. I think I understand the motivation and how the code changes solve the problem. Have you worked through the issue with the version lib?

No I haven't, mainly because I can't replicate it, and am very squeezed for time. It could be weeks before I have a chance to look into it properly, unless the same issue crops up when we go to build on a new workstation we have.

reos-rcrozier:
@rafmudaf , I was able to replicate the issue on a parallel build. It seems that there can be an issue with cmake if a fortran source  file is built more than once. I have modified the cmake files to ensure this was the case. 

I also made a modification to reduce problems which occur when building with cmake option `BUILD_OPENFAST_SIMULINK_API=ON`. The mere act of linking to the mex libraries can cause the standard libraries shipped with matlab to be linked to throughout the project. Since the nwtcs library is linked to the mex libraries, and is used throughout the project the whole build can be polluted in this way, resulting in failures due to mismated library versions. I threrfore made changes to minimize as much as possible the places where the mex libraries are linked. In fat now two openfast libraries are built, `openfastlib` and `openfastlib_mex`, the latter being intended to be linked to the simulink interface.

Everything compiles now, including the simulink S-Function, however, I've not actually tested running anything yet.

reos-rcrozier:
@rafmudaf Could you now approve the testing workflow please?

I have now tested that everything seems to be working, by doing some basic simulations, I've not run the test suite. I also fixed an issue building the openfastcpp lib due to missing OpenMP compiler flags.

rafmudaf:
> @rafmudaf Could you now approve the testing workflow please?

@reos-rcrozier done and its running

reos-rcrozier:
ok, 6 out of 7 test now passing, just the interface-tests. What's different about this test? The error seems to be about multiply defined functions in the sys file, but why is it only showing up in this test?

AaronTUB:
While using OpenFAST 3.1.0, this pull request was a lifesaver for us. But we now upgraded to OpenFAST 3.3.0, where we recreated the changes from this pull request. In case anybody is interested you can find it [here](https://github.com/AaronTUB/openfast/tree/fixed_3.3.0).

reos-rcrozier:
It seems this is useful, can we trigger a workflow please, I haven't tested locally yet.

rafmudaf:
@reos-rcrozier I've enabled the automated tests. Please @ me directly if you need a maintainer to do anything via GitHub again.

reos-rcrozier:
Bah, I'll need to debug offline

reos-rcrozier:
@rafmudaf I have enabled actions in my fork and with recent changes to the branch all tests are passing except for `rtest-interfaces`. You can see the test results [here](https://github.com/REOptimize-Systems/openfast/actions/runs/3960280079/jobs/6784641583).

The failed tests seem to be timeouts in some of the tests, I don't really know how severe this is, will it prevent merging?

reos-rcrozier:
two checks are failing but one is after 6 seconds due to:

`Error: No file in /home/runner/work/openfast/openfast matched to [**/requirements.txt], make sure you have checked out the target repository`

other is due to timeouts.

andrew-platt:
@reos-rcrozier, that is a strange error.  We've had some issues with GitHub actions occasionally failing.  Rerunning the failed jobs often will fix it, so I'm doing that now.

andrew-platt:
After rerunning tests multiple times, there are some consistent failures in the python OF test cases. Example from recent run: 

`32: Traceback (most recent call last):
32:   File "/home/runner/work/openfast/openfast/reg_tests/executePythonRegressionCase.py", line 137, in <module>
32:     openfastlib = openfast_library.FastLibAPI(openfastlib_path, caseInputFile)
32:   File "/home/runner/work/openfast/openfast/reg_tests/../glue-codes/python/openfast_library.py", line 25, in __init__
32:     super().__init__(library_path)
32:   File "/opt/hostedtoolcache/Python/3.9.16/x64/lib/python3.9/ctypes/__init__.py", line 374, in __init__
32:     self._handle = _dlopen(self._name, mode)
32: OSError: /home/runner/work/openfast/openfast/build/reg_tests/glue-codes/python/../../../modules/openfast-library/libopenfastlib.so: cannot open shared object file: No such file or directory`

I can't tell yet if this is due to something in GH, or due to something in the changes (I haven't looked in detail yet).

There are also some consistent failures in the module regression tests that are not making much sense to me. I'm rerunning the entire set to see if this clears.

reos-rcrozier:
could be a real issue, I will check if `libopenfastlib.so` is still actually created

reos-rcrozier:
The failing test was due to my omitting to make openfastlib always be shared, as the python interface requires this (it uses dlopen). Now that is corrected all tests are passing. I think the PR is now ready for merge. It's worth noting it might have an impact on any branches under development which are for new modules not present in these sources. They might also need to be modified to use object libraries.

reos-rcrozier:
@rafmudaf @andrew-platt any chance to consider the merge? Sorry to push, but it would be good to complete this while it's fresh in my mind.

reos-rcrozier:
@rafmudaf @andrew-platt sorry, but pinging again to see if this can now be considered for merge?

andrew-platt:
@reos-rcrozier, my apologies for the delay in getting to this.  I have a few concerns about what impacts this might have on some of the systems we deploy on, but have not had a chance to test your branch or do a thorough review yet.

I would like to get this into our v3.5.0 release that we plan to release before the end of the month, so we will be looking at this within the next week or two.

@deslaughter, is there anything you wanted to add to the discussion at present?

reos-rcrozier:
@andrew-platt we have experience cross-compiling to many platforms, so if you need testing perhaps we can help.

For example I have previously cross-compiled FAST for PowerPC and (i think, it was some time ago) arm, although some tweaks were needed to FORTRAN code for powerPC.

henyau:
Hi all. I'm not sure if this will be helpful to anyone, but to get a working MEX file on my system (Ubuntu 22.04) with USE_LOCAL_STATIC_LAPACK ON, I had to explicitly add a lot of sources to the simulink CMakeLists.txt (NWTC, PDAS, NetLib SLATEC, Supercontroller, RanLux, Openfoam). Otherwise I would get errors like: "undefined symbol: __openfoam_types_MOD_opfm_destroyoutput" when calling FAST_SFunc in Matlab.

reos-rcrozier:
@henyau Did you do a static or shared build of OpenFAST. Doing a static build (-DBUILD_SHARED_LIBS=OFF as in the example at the top of this thread) should mean you don't have any undefined references.

deslaughter:
@henyau I'm also having a lot of issues with this combination of flags on Linux. It seems to work fine on my Mac. If you have the chance, could you try `-DBLA_STATIC=ON` instead of `-DUSE_LOCAL_STATIC_LAPACK=ON`? This option will find and use a static BLAS/LAPACK from your system instead of building from source. `BLA_STATIC` works with the Intel MKL library and has been tested on Windows and Linux.

henyau:
@reos-rcrozier: Yes, was doing a static build (-DBUILD_SHARED_LIBS=OFF).
@deslaughter:  I just tried it with -DBLA_STATIC=ON (Cmake finds and uses OpenBLAS). It compiles and links fine, but I get the same undefined symbol error as before when trying to run the Simulink model. Using Intel MKL would have been my next step if I hadn't gotten it to work.

reos-rcrozier:
@henyau ok, can I first confirm you using `create_FAST_SFunc.m` to create the mex file? If so, you could also try adding some more `-l` options to explicitly link to the libraries there which might also solve the problem. Would also be useful to confirm the compiler suite and version you were are using please.

reos-rcrozier:
Actually, I didn't realize until now that @deslaughter removed most of the changes introduced in this PR. So I probably can't help much.

reos-rcrozier:
@deslaughter A bit frustrating that all this work we did, specifically to make simulink mex functions work properly on lInux was simply removed without tagging me to see if we could resolve the mac issue. I don't think removing the object libs was the right way to go.

deslaughter:
> I just tried it with -DBLA_STATIC=ON (Cmake finds and uses OpenBLAS). It compiles and links fine, but I get the same undefined symbol error as before when trying to run the Simulink model. Using Intel MKL would have been my next step if I hadn't gotten it to work.

@henyau I was able to reproduce the issue on Linux and tracked it down to the order that the libraries are linked in `glue-codes/simulink/CMakeLists.txt`. Could you try the following order and let me know if it resolves the issue? If not, please include the full error message from Matlab.

```
  $<TARGET_FILE:openfast_prelib>
  $<TARGET_FILE:aerodyn14lib>
  $<TARGET_FILE:aerodynlib>
  $<TARGET_FILE:beamdynlib>
  $<TARGET_FILE:elastodynlib>
  $<TARGET_FILE:extptfm_mckflib>
  $<TARGET_FILE:feamlib>
  $<TARGET_FILE:hydrodynlib>
  $<TARGET_FILE:icedynlib>
  $<TARGET_FILE:icefloelib>
  $<TARGET_FILE:maplib>
  $<TARGET_FILE:moordynlib>
  $<TARGET_FILE:orcaflexlib>
  $<TARGET_FILE:sctypeslib>
  $<TARGET_FILE:servodynlib-matlab>
  $<TARGET_FILE:subdynlib>
  $<TARGET_FILE:foamfastlib>
  $<TARGET_FILE:ifwlib>
  $<TARGET_FILE:scfastlib>
  $<TARGET_FILE:foamtypeslib>
  $<TARGET_FILE:versioninfolib>
  $<TARGET_FILE:nwtclibs-matlab>
```

deslaughter:
@henyau I really appreciate you trying out the changes. Since you've resolved the issue and have a working MEX file, I don't want to take up any more of your time trying to troubleshoot the build system.

deslaughter:
> @deslaughter A bit frustrating that all this work we did, specifically to make simulink mex functions work properly on lInux was simply removed without tagging me to see if we could resolve the mac issue. I don't think removing the object libs was the right way to go.

@reos-rcrozier I apologize for the way this PR was handled, I certainly don't want to make you feel like your contribution wasn't appreciated. After it was merged, several issues were submitted regarding building on Windows and Mac. It was determined that the Object Libraries weren't compatible with Visual Studio on Windows and that their added complexity would be difficult for users to understand. I tried very hard to fix these issues while keeping the object libraries and they were only removed as a last resort. I should have communicated this information better and will strive to do so in the future.

I like the concepts that were introduced, many of which are part of the current build system, and would like to revisit the use of Object Libraries, once https://gitlab.kitware.com/cmake/cmake/-/issues/18090 is resolved. I agree that they solved several issues with managing the OpenFAST libraries.

PR #1019, opened 2022-03-01
docs: update rtd to use sphinxcontrib-bibtex v2.0.0 and later

andrew-platt:
This PR is ready to merge.

**Summary**
I had locked the `sphinxcontrib-bibtex` version to < 2.0.0 when I ran into issues with multiple bibliographies that we use (one per module).  This was increasingly becoming a problem as python 3.8 and later would like to use a newer version.

**Solution**
The `conf.py` was updated to include a `bibtex_bibfiles` variable.  However, the new version of `sphinxcontrib-bibtex` is more thorough in checking for repeated bibtex entries, and complained about a few duplicated entries (checked across all bib files).  This required hardcoding of the `keyprefix` values into each modules .bib file instead of using sphinx to insert them.  The rendering is therefore identical to the previous behavior, but will now compile correctly with the newer package.

See https://ap-openfast.readthedocs.io/en/f-docs-bibtex2.0.0/source/user/fast.farm/zrefs.html (duplicated entries first appeared in the fast.farm docs).

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -67,6 +67,17 @@ extensions = [
     'sphinxcontrib.bibtex',
     'sphinxcontrib.mermaid',
 ]
+bibtex_bibfiles = [
+    'source/user/aerodyn-aeroacoustics/references.bib',
+    'source/user/aerodyn-olaf/bibliography.bib',
+    'source/user/aerodyn/bibliography.bib',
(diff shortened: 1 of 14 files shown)


andrew-platt:
@bjonkman, this should fix the python version issue you ran into with the documentation builds.

PR #1020, opened 2022-03-01
Document AD outputs 

bjonkman:
Previously, AeroDyn had some undocumented outputs: `B{1:4}Aero{F|M}{x|y|z}` , `B{1:4}AeroPwr`, `B{1:3}Aero{F|M}{x|y|z}g`, `RtAero{F|M}{x|y|z}g`. This pull request adds those outputs to the `OutListParameters.xlsx` file and updates the Matlab-generated parameters in AeroDyn_IO.f90.

In addition, there are some minor changes to comments and example input files in the documentation.

AD, documentation

This should not change any test results.


andrew-platt:
I'm going to double check the failing AD15 driver cases to make sure we don't have an issue with the channel name changes.  If we do, it should be a minor update to those r-test cases.

bjonkman:
I think the problem is that intermittent issue (likely some memory issue related to the AD input file reading), since some of those files do pass on my fork: https://github.com/bjonkman/openfast/actions/runs/1918669057; also the gfortran11 debug test does pass here, too.

andrew-platt:
Good to know.  I was hoping it was our friendly intermittent failures and not an output channel name change.

andrew-platt (approved):
Looks good to me!  Thanks for putting this together!

The test case that is failing are the AD driver cases that occasionally fails -- it passed in Bonnie's branch so it is not due to output channel renaming.

PR #971, opened 2022-01-19
Update GH actions to update packages for ubuntu debug regression test

bjonkman:
For the past week, the OpenFAST GH actions for the ubuntu debug regression tests were failing with this error:
```
E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/l/lxml/python3-lxml_4.5.0-1ubuntu0.3_amd64.deb 404 Not Found [IP: 52.250.76.244 80]

E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?


Error: Process completed with exit code 100.
```
The requested package version wasn't available, but a newer one was. This PR adds the `apt-get update` command to the workflow to fix the problem.

PR #951, opened 2021-12-28
Corrected the description of SkewModFactor in Documentation

tatesnow:
Corrected the description of SkewModFactor

yes

**Feature or improvement description**

**Related issue, if one exists**

**Impacted areas of the software**

**Additional supporting information**

**Test results, if applicable**

PR #917, opened 2021-11-11
AD15: add nodal outputs for VUnd{xyz}i in global coords

andrew-platt:
This PR is  **NOT** ready for merging (more channel requests)

**Feature or improvement description**
Three new nodal outputs are added to AD15 for the undisturbed wind velocity at each blade node:

- `VUndxi`  Undisturbed wind velocity at blade node in global X
- `VUndyi`  Undisturbed wind velocity at blade node in global Y
- `VUndzi`  Undisturbed wind velocity at blade node in global Z

These are particularly useful in checking the inflow into a blade on an OpenFAST turbine in a CFD coupled or FAST.Farm simulation.  This feature was requested by @kshaler.

Example of data for the VUndxi channel for blade 1 (from the 5MW_Land_DLL_WTurb regression test case)
Screen Shot 2021-11-11 at 3 43 59 PM

**TODO**

- [ ] convert channel doc pdf to .rst and update
- [ ] add blade nodal channels  (auto generated sets):
   - [ ] induced velocity in global
   - [ ] disturbed wind in global
- [ ] regular outputs for (9 node ones):
   - [ ] all the above
   - [ ] VUndxi etc.

PR #911, opened 2021-11-06
Update this_doc.rst

abksr:
Fixed a typo from "PFD" to "PDF".

**Feature or improvement description**

**Related issue, if one exists**

**Impacted areas of the software**

**Additional supporting information**

**Test results, if applicable**

PR #1023, opened 2022-03-02
OpenFAST v3.1.0

rafmudaf:
Pull request to merge `main` into `dev` and create a tagged release for `v3.1.0`.

See GitHub Actions

# Release checklist:
- [x] Update the documentation version in docs/conf.py
- [x] Update the versions in docs/source/user/api_change.rst
- [x] Verify readthedocs builds correctly
- [ ] Create a tag in OpenFAST
- [ ] Create a merge commit in r-test and add a corresponding tag
- [ ] Compile executables for Windows builds
    - [ ] FAST_SFunc.mexw64
    - [ ] MAP_X64.dll
    - [ ] OpenFAST-Simulink_x64.dll
    - [ ] openfast_x64.exe
    - [ ] DISCON.dll

# Proposed Change Log

### General
#707 Add environmental variables to driver input files
#734 OpenFAST Registry algorithm change: simplify USE statements
Others (#826, #835, #836, #911, #918, #971, #1019)

### AeroDyn
#688 AeroDyn driver update for multiple wind turbines, with arbitrary motions and geometries
#729 New features for unsteady aerodynamics modeling
#834 Fix AD Driver unallocated variable error with GCC 11
#863 AeroDyn cleanup
#917 AD15: add nodal outputs for VUnd{xyz}i in global coords
#919 Segment treecode
#920 Update in Computing Default Unsteady Airfoil Coefficients
#922 [BugFix] Minor bugfix in AirfoilInfo
#982 [BugFix] AD15 nacelle reference position was set to hub position
#1001 Remove conditional statement for initialization of BEMT variable
#1009 [BugFix] Nacelle position set inconsistently by glue code and AeroDyn driver

### BeamDyn
#996 [BugFix] BeamDyn nodal outputs occasionally segfaulted

### FAST Farm
#839 Fix Bug in FAST.Farm Causing Wake Bounce-Back
#860 Fix some memory leaks in FAST.Farm
#895 [BugFix] incorrect init of aggregated output index arrays in FAST.Farm
#923 [BugFix] error handling in AWAE module

### HydroDyn
#756 HydroDyn primary input file passing and parsing
#831 HydroDyn Input/Output meshes: change from SWL to MSL for consistency with OF glue code
#838 [BugFix] segmentation fault in HD linearization 
#915 [BugFix] Incorrect reference frame used in HD for WAMIT/WAMIT2 
#998 Fix HydroDyn summary file nodal data is incorrect when Member is flipped

### InflowWind
#720 inflowWind C-bound interface and python wrapper
#929 Fix issue with uninitialized variables in InflowWind's Direct Scaling method

### NWTC Library
#1002 Increase line length in FileInfo parsing methods

### OpenFAST Library
#716 [BugFix] Fix C++ API for restart, Error handling in FAST Library, and AeroDyn echo file lock
#958 Lin: CalcSteady, forcing linearization at end of simulation

### ServoDyn
#902 Fixes for Intel in debug mode
#930 Stop OpenFAST for Simulink simulation when trim solution has been found

### SubDyn
#859 Various improvements to SubDyn

### Documentation
#740 Guidelines for performance considerations with Fortran
#753 Migrate the HydroDyn Manual to readthedocs
#805 Include legacy documentation in pdf and MS Word format - General & ElastoDyn
#828 Add instructions for adding new regression test cases
#858 Documentation for ExtPtfm
#951 Corrected the description of SkewModFactor in Documentation
#1020 Document AD outputs


icnosvbovwb:
how to use simulink moedel?

icnosvbovwb:
There are 53 number interfaces for simulink. It has 8 number with openfast 2.4.0.

jjonkman:
Dear @icnosvbovwb,

Regarding the switch from 8 to more channels in the OpenFAST v3.1 S-Function for Simulink, see: https://github.com/OpenFAST/openfast/discussions/1163 and https://github.com/OpenFAST/openfast/pull/1176.

Best regards,

icnosvbovwb:
Thanks，by the way，FAST_Library.f90 in the OpenFAST v3.1 with NumOuts_c = min(MAXOUTPUTS, SUM(Turbine(iTurb)%y_FAST%numOuts )),but it is "NumOuts_c = min(MAXOUTPUTS, 1+SUM(Turbine(iTurb)%y_FAST%numOuts ))" in V2.4.I don't sure change where.

> Dear @icnosvbovwb,
> 
> Regarding the switch from 8 to more channels in the OpenFAST v3.1 S-Function for Simulink, see: #1163 and #1176.
> 
> Best regards,

icnosvbovwb:
Dear Jonkman
&nbsp; &nbsp; I meet a problem.I run 5MW_ITIBarge_DLL_WTurb_WavesIrr.fst with V3.1 and it gives error like this and abortsAs follows：
&nbsp; &nbsp;I'm not sure if I want to add TMD or TLCD ?


Running ServoDyn.
&nbsp; Time: 0 of 600 seconds.
+&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;&nbsp;
+ Time: 10 of 600 seconds.&nbsp; Estimated final completion at 22:20:12.
+&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;&nbsp;
+ Time: 20 of 600 seconds.&nbsp; Estimated final completion at 22:20:23.
+&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;&nbsp;
+ Time: 30 of 600 seconds.&nbsp; Estimated final completion at 22:20:02.
+&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;&nbsp;
+ Time: 40 of 600 seconds.&nbsp; Estimated final completion at 22:19:54.


&nbsp;FAST_Solution:FAST_AdvanceStates:ED_ABM4:ED_CalcContStateDeriv:SetCoordSy:Small angle assumption
&nbsp;violated in SUBROUTINE SmllRotTrans() due to a large blade deflection (ElastoDyn SetCoordSy). The
&nbsp;solution may be inaccurate. Simulation continuing, but future warnings from SmllRotTrans() will
&nbsp;be suppressed.
&nbsp; Additional debugging message from SUBROUTINE SmllRotTrans(): 43.6 s


+&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;&nbsp;
+ Time: 50 of 600 seconds.&nbsp; Estimated final completion at 22:19:55.
+&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;&nbsp;
+ Time: 60 of 600 seconds.&nbsp; Estimated final completion at 22:19:56.


&nbsp;FAST_Solution:FAST_AdvanceStates:SolveOption2c_Inp2AD_SrvD:InflowWind_CalcOutput:CalcOutput:IfW_FF
&nbsp;Wind_CalcOutput [position=(-5, 0, 90)&nbsp; in wind-file coordinates]: Error: FF wind array was
&nbsp;exhausted at 65.65 seconds (trying to access data at 72.108 seconds).


&nbsp; OpenFAST encountered an error at simulation time 65.638 of 600 seconds.
&nbsp; Simulation error level: FATAL ERROR


&nbsp; Aborting OpenFAST.




------------------&nbsp;原始邮件&nbsp;------------------
发件人:                                                                                                                        "OpenFAST/openfast"                                                                                    ***@***.***&gt;;
发送时间:&nbsp;2022年8月25日(星期四) 晚上10:58
***@***.***&gt;;
***@***.******@***.***&gt;;
主题:&nbsp;Re: [OpenFAST/openfast] OpenFAST v3.1.0 (PR #1023)





 
The change in NumOuts_c is due to the fact that OpenFAST now counts the time channel in its y_FAST%numOuts variable, but it didn't in v2.4. You shouldn't have to worry about that in your Simulink model.
 
For OpenFAST 3.1, you'll need to add 43 zeros to the array of inputs to the S Function. For example:
  
see Airfoil Controller (3 zeros), Cable deltaL controller (20 zeros), and Cable deltaLdot controller (20 zeros) on this example OpenLoop model
 
 
see 43 zeros in #1205 (comment)
  
Those additional 43 inputs must exist, but they won't be used in any models that you are converting to this new version.

jjonkman:
Dear @icnosvbovwb,

A similar question was asked and answered in the following issue: https://github.com/OpenFAST/openfast/issues/832.

Best regards,

icnosvbovwb:
Dear Jonkman
Thank you for your answer. I used the TurbSim_v150.exe you mentioned to generate 630 seconds of turbulent wind, and I can run the ITIBarge model. On this basis, I added TMD at 1.5 meters of the platform, but the effect did not change. The parameters of TMD are as follows:

STRUCTURAL CONTROL (StC) INPUT FILE ----------------------------
Input file for tuned mass damper, module by Matt Lackner, Meghan Glade, and Semyung Park (UMass)
---------------------- SIMULATION CONTROL --------------------------------------
True          Echo         - Echo input data to <RootName>.ech (flag)
---------------------- StC DEGREES OF FREEDOM ----------------------------------
          2   StC_DOF_MODE - DOF mode (switch) {0: No StC or TLCD DOF; 1: StC_X_DOF, StC_Y_DOF, and/or StC_Z_DOF (three independent StC DOFs); 2: StC_XY_DOF (Omni-Directional StC); 3: TLCD; 4: Prescribed force/moment time series}
true          StC_X_DOF    - DOF on or off for StC X (flag) [Used only when StC_DOF_MODE=1]
true          StC_Y_DOF    - DOF on or off for StC Y (flag) [Used only when StC_DOF_MODE=1]
FALSE         StC_Z_DOF    - DOF on or off for StC Z (flag) [Used only when StC_DOF_MODE=1]
---------------------- StC LOCATION ------------------------------------------- [relative to the reference origin of component attached to]
          0   StC_P_X      - At rest X position of StC (m)
          0   StC_P_Y      - At rest Y position of StC (m)
         1.5   StC_P_Z      - At rest Z position of StC (m)
---------------------- StC INITIAL CONDITIONS --------------------------------- [used only when StC_DOF_MODE=1 or 2]
          1   StC_X_DSP    - StC X initial displacement (m) [relative to at rest position]
          1   StC_Y_DSP    - StC Y initial displacement (m) [relative to at rest position]
          0   StC_Z_DSP    - StC Z initial displacement (m) [relative to at rest position; used only when StC_DOF_MODE=1 and StC_Z_DOF=TRUE]
"none"        StC_Z_PreLd  - StC Z pre-load (N) {"gravity" to offset for gravity load; "none" or 0 to turn off} [used only when StC_DOF_MODE=1 and StC_Z_DOF=TRUE]
---------------------- StC CONFIGURATION -------------------------------------- [used only when StC_DOF_MODE=1 or 2]
         20   StC_X_PSP    - Positive stop position (maximum X mass displacement) (m)
        -20   StC_X_NSP    - Negative stop position (minimum X mass displacement) (m)
         20   StC_Y_PSP    - Positive stop position (maximum Y mass displacement) (m)
        -20   StC_Y_NSP    - Negative stop position (minimum Y mass displacement) (m)
         10   StC_Z_PSP    - Positive stop position (maximum Z mass displacement) (m) [used only when StC_DOF_MODE=1 and StC_Z_DOF=TRUE]
        -10   StC_Z_NSP    - Negative stop position (minimum Z mass displacement) (m) [used only when StC_DOF_MODE=1 and StC_Z_DOF=TRUE]
---------------------- StC MASS, STIFFNESS, & DAMPING ------------------------- [used only when StC_DOF_MODE=1 or 2]
      1.0904E+05   StC_X_M      - StC X mass (kg) [must equal StC_Y_M for StC_DOF_MODE = 2]
      1.0904E+05   StC_Y_M      - StC Y mass (kg) [must equal StC_X_M for StC_DOF_MODE = 2]
          0   StC_Z_M      - StC Z mass (kg) [used only when StC_DOF_MODE=1 and StC_Z_DOF=TRUE]
      1.0904E+05   StC_XY_M     - StC XY mass (kg) [used only when StC_DOF_MODE=2]
      3.2550E+04   StC_X_K      - StC X stiffness (N/m)
      3.2550E+04   StC_Y_K      - StC Y stiffness (N/m)
          0   StC_Z_K      - StC Z stiffness (N/m) [used only when StC_DOF_MODE=1 and StC_Z_DOF=TRUE]
      1.0306E+04   StC_X_C      - StC X damping (N/(m/s))
      1.0306E+04   StC_Y_C      - StC Y damping (N/(m/s))
          0   StC_Z_C      - StC Z damping (N/(m/s)) [used only when StC_DOF_MODE=1 and StC_Z_DOF=TRUE]
      1500000   StC_X_KS     - Stop spring X stiffness (N/m)
      1500000   StC_Y_KS     - Stop spring Y stiffness (N/m)
          0   StC_Z_KS     - Stop spring Z stiffness (N/m) [used only when StC_DOF_MODE=1 and StC_Z_DOF=TRUE]
      100000   StC_X_CS     - Stop spring X damping (N/(m/s))
      100000   StC_Y_CS     - Stop spring Y damping (N/(m/s))
          0   StC_Z_CS     - Stop spring Z damping (N/(m/s)) [used only when StC_DOF_MODE=1 and StC_Z_DOF=TRUE]
---------------------- StC USER-DEFINED SPRING FORCES ------------------------- [used only when StC_DOF_MODE=1 or 2]
False         Use_F_TBL    - Use spring force from user-defined table (flag)
         17   NKInpSt      - Number of spring force input stations
---------------------- StC SPRING FORCES TABLE -------------------------------- [used only when StC_DOF_MODE=1 or 2]
    X                F_X               Y              F_Y              Z              F_Z
   (m)               (N)              (m)             (N)             (m)             (N)
-6.0000000E+00  -4.8000000E+06  -6.0000000E+00  -4.8000000E+06  -6.0000000E+00  -4.8000000E+06
-5.0000000E+00  -2.4000000E+06  -5.0000000E+00  -2.4000000E+06  -5.0000000E+00  -2.4000000E+06
-4.5000000E+00  -1.2000000E+06  -4.5000000E+00  -1.2000000E+06  -4.5000000E+00  -1.2000000E+06
-4.0000000E+00  -6.0000000E+05  -4.0000000E+00  -6.0000000E+05  -4.0000000E+00  -6.0000000E+05
-3.5000000E+00  -3.0000000E+05  -3.5000000E+00  -3.0000000E+05  -3.5000000E+00  -3.0000000E+05
-3.0000000E+00  -1.5000000E+05  -3.0000000E+00  -1.5000000E+05  -3.0000000E+00  -1.5000000E+05
-2.5000000E+00  -1.0000000E+05  -2.5000000E+00  -1.0000000E+05  -2.5000000E+00  -1.0000000E+05
-2.0000000E+00  -6.5000000E+04  -2.0000000E+00  -6.5000000E+04  -2.0000000E+00  -6.5000000E+04
 0.0000000E+00   0.0000000E+00   0.0000000E+00   0.0000000E+00   0.0000000E+00   0.0000000E+00
 2.0000000E+00   6.5000000E+04   2.0000000E+00   6.5000000E+04   2.0000000E+00   6.5000000E+04
 2.5000000E+00   1.0000000E+05   2.5000000E+00   1.0000000E+05   2.5000000E+00   1.0000000E+05
 3.0000000E+00   1.5000000E+05   3.0000000E+00   1.5000000E+05   3.0000000E+00   1.5000000E+05
 3.5000000E+00   3.0000000E+05   3.5000000E+00   3.0000000E+05   3.5000000E+00   3.0000000E+05
 4.0000000E+00   6.0000000E+05   4.0000000E+00   6.0000000E+05   4.0000000E+00   6.0000000E+05
 4.5000000E+00   1.2000000E+06   4.5000000E+00   1.2000000E+06   4.5000000E+00   1.2000000E+06
 5.0000000E+00   2.4000000E+06   5.0000000E+00   2.4000000E+06   5.0000000E+00   2.4000000E+06
 6.0000000E+00   4.8000000E+06   6.0000000E+00   4.8000000E+06   6.0000000E+00   4.8000000E+06
---------------------- StructCtrl CONTROL -------------------------------------------- [used only when StC_DOF_MODE=1 or 2]
          0   StC_CMODE     - Control mode (switch) {0:none; 1: Semi-Active Control Mode; 4: Active Control Mode through Simulink (not available); 5: Active Control Mode through Bladed interface}
          0   StC_CChan     - Control channel group (1:10) for stiffness and damping (StC_[XYZ]_K, StC_[XYZ]_C, and StC_[XYZ]_Brake) (specify additional channels for blade instances of StC active control -- one channel per blade) [used only when StC_DOF_MODE=1 or 2, and StC_CMODE=4 or 5]
          1   StC_SA_MODE   - Semi-Active control mode {1: velocity-based ground hook control; 2: Inverse velocity-based ground hook control; 3: displacement-based ground hook control 4: Phase difference Algorithm with Friction Force 5: Phase difference Algorithm with Damping Force} (-)
          0   StC_X_C_HIGH  - StC X high damping for ground hook control
          0   StC_X_C_LOW   - StC X low damping for ground hook control
          0   StC_Y_C_HIGH  - StC Y high damping for ground hook control
          0   StC_Y_C_LOW   - StC Y low damping for ground hook control
          0   StC_Z_C_HIGH  - StC Z high damping for ground hook control [used only when StC_DOF_MODE=1 and StC_Z_DOF=TRUE]
          0   StC_Z_C_LOW   - StC Z low damping for ground hook control  [used only when StC_DOF_MODE=1 and StC_Z_DOF=TRUE]
          0   StC_X_C_BRAKE - StC X high damping for braking the StC (Don't use it now. should be zero)
          0   StC_Y_C_BRAKE - StC Y high damping for braking the StC (Don't use it now. should be zero)
          0   StC_Z_C_BRAKE - StC Z high damping for braking the StC (Don't use it now. should be zero) [used only when StC_DOF_MODE=1 and StC_Z_DOF=TRUE]
---------------------- TLCD --------------------------------------------------- [used only when StC_DOF_MODE=3]
     7.9325   L_X             - X TLCD total length (m)
     6.5929   B_X             - X TLCD horizontal length (m)
     2.0217   area_X          - X TLCD cross-sectional area of vertical column (m^2)
      0.913   area_ratio_X    - X TLCD cross-sectional area ratio (vertical column area divided by horizontal column area) (-)
     2.5265   headLossCoeff_X - X TLCD head loss coeff (-)
       1000   rho_X           - X TLCD liquid density (kg/m^3)
     3.5767   L_Y             - Y TLCD total length (m)
     2.1788   B_Y             - Y TLCD horizontal length (m)
     1.2252   area_Y          - Y TLCD cross-sectional area of vertical column (m^2)
     2.7232   area_ratio_Y    - Y TLCD cross-sectional area ratio (vertical column area divided by horizontal column area) (-)
     0.6433   headLossCoeff_Y - Y TLCD head loss coeff (-)
       1000   rho_Y           - Y TLCD liquid density (kg/m^3)
---------------------- PRESCRIBED TIME SERIES --------------------------------- [used only when StC_DOF_MODE=4]
          0   PrescribedForcesCoord- Prescribed forces are in global or local coordinates (switch) {1: global; 2: local}
"TimeForceSeries.dat"  PrescribedForcesFile   - Time series force and moment (7 columns of time, FX, FY, FZ, MX, MY, MZ)
-------------------------------------------------------------------------------

jjonkman:
Dear @icnosvbovwb,

Do you see absolutely no change in response or just less change than you are expecting?

Have you set `NumSStC` to 1 and specified your StC input file via `SStCfiles` in ServoDyn?

Have you added StC-related outputs to the ServoDyn `OutList` (documented here: https://openfast.readthedocs.io/en/main/_downloads/3f19498a5dc774461e022b671ff01ec6/OutListParameters.xlsx)?  What do the time series of `SStC1_XQ` and `SStC1_YQ` for example look like?

Best regards,

icnosvbovwb:
Dear Jonkman：
For the first question, from the comparison of the pictures, there should be basically no change. The pictures are as follows.
For the second question I added the number 1 and the TMD file name in the servo control file.
For the third question, the outlist outputs a total of 139 data names. I compared pitch and roll, so I'm not sure if TMD has a bad effect?






------------------ Original ------------------
From: jjonkman ***@***.***&gt;
Date: Tue,Aug 30,2022 8:50 PM
To: OpenFAST/openfast ***@***.***&gt;
Cc: icnosvbovwb ***@***.***&gt;, Mention ***@***.***&gt;
Subject: Re: [OpenFAST/openfast] OpenFAST v3.1.0 (PR #1023)





 
Dear @icnosvbovwb,
 
Do you see absolutely no change in response or just less change than you are expecting?
 
Have you set NumSStC to 1 and specified your StC input file via SStCfiles in ServoDyn?
 
Have you added StC-related outputs to the ServoDyn OutList (documented here: https://openfast.readthedocs.io/en/main/_downloads/3f19498a5dc774461e022b671ff01ec6/OutListParameters.xlsx)?  What do the time series of SStC1_XQ and SStC1_YQ for example look like?
 
Best regards,

icnosvbovwb:
piture
piture2

icnosvbovwb:
piture3
piture4

jjonkman:
Dear @icnosvbovwb,

I would suggest adding StC-related outputs to the ServoDyn `OutList` (documented here: https://openfast.readthedocs.io/en/main/_downloads/3f19498a5dc774461e022b671ff01ec6/OutListParameters.xlsx), including `SStC1_XQ` and `SStC1_YQ`. What do the time series of `SStC1_XQ` and `SStC1_YQ` look like?

Best regards,

icnosvbovwb:
Dear Jonkman
Your reply is very effective, and I found a problem, I used V2.4 and V3.1 to calculate the 1 model, their results are not the same, I am not sure if something has changed?
Thank you for your reply.

icnosvbovwb:
Dear Jonkman
Your reply is very effective, and I found a problem, I used V2.4 and V3.1 to calculate the 5MW_OC4Semi_WSt_WavesWN model, their results were not the same, I was not sure if something had changed?For example, the result of the platform pitch is as follows:
Thank you for your reply.

icnosvbovwb:
Dear Jonkman
Your reply is very effective, and I found a problem, I used V2.4 and V3.1 to calculate the 5MW_OC4Semi_WSt_WavesWN model, their results were not the same, I was not sure if something had changed?For example, the result of the platform pitch and platform roll are as follows:
Thank you for your reply.

icnosvbovwb:
PITCURE3
PITCURE4

jjonkman:
Dear @icnosvbovwb,

I'm not sure what (if anything) you changed in the r-test of the NREL 5-MW wind turbine atop the OC4-DeepCwind semisumbersible, but there have certainly been major updates to OpenFAST between v2.4 and v3.1, including a major change to HydroDyn in v2.6.  You could always rerun the simulation with each major release of OpenFAST to more specifically target which change influences the results for a particular case.  I'm not sure I can comment in general how the various changes made would impact a particular simulation, especially one where I'm not familiar with your model set up.

A similar question was discussed here: https://github.com/OpenFAST/openfast/issues/1022.

So, did you resolve the issue with StC?

Best regards,

icnosvbovwb:
Dear Jonkman：
The previous problem has been solved. Now the new problem is that the calculation results of the two versions between V2.4 and V3.1 are different. The code of the example I calculated has not changed. They are all compiled from source code, which meas that I have not changed anything in it.

jjonkman:
Dear @icnosvbovwb,

While it sounds like you have not changed the source code, the source code was changed by NREL between this different releases of OpenFAST, and some of the changes could impact the results.  Again, I recommend identifying which version led to the major change in results and refer to #1022.

Best regards,

icnosvbovwb:
Thank you for your reply. There are other questions about how to use the animation function for the ITI barge r-test. I read the discussion of #1090 and understood that the switch V14 was changed to V15, but what about other corresponding specific module changes? Because I replaced NRELOffshrBsline5MW_AeroDyn_Equil_noTwr.dat with NRELOffshrBsline5MW_OC3Hywind_AeroDyn15 alone, an error will be reported.

icnosvbovwb:
Dear @jjonkman 
Thank you for your reply. There are other questions about how to use the animation function for the ITI barge r-test. I read the discussion of #1090 and understood that the switch V14 was changed to V15, but what about other corresponding specific module changes? Because I replaced NRELOffshrBsline5MW_AeroDyn_Equil_noTwr.dat with NRELOffshrBsline5MW_OC3Hywind_AeroDyn15 alone, an error will be reported.

jjonkman:
Dear @icnosvbovwb,

To use the visualization functionality of OpenFAST, I agree that you must use AeroDyn v15 instead of v14.  When you switch AeroDyn versions with the input files you describe, what error are you getting?  I would think this switch would generally be OK, although, you'd likely need to change the tower aerodynamic properties because the tower design is different between the ITI Energy barge and OC3-Hywind spar.  It is probably better to start with _NRELOffshrBsline5MW_OC3Monopile_AeroDyn15.dat_ from the OC3-monopile model.

Best regards,

icnosvbovwb:
Dear @jjonkman 
I replace ”NRELOffshrBsline5MW_AeroDyn_Equil_noTwr“ with  “NRELOffshrBsline5MW_OC3Monopile_AeroDyn15” , click simulink to apply, and an error window pops up.
pitcure1

jjonkman:
Dear @icnosvbovwb,

Assuming you didn't change the AeroDyn input file, my guess is the input file you are using is formatted for a different version of OpenFAST than you are running within MATLAB/Simulink.  As with any input file processing error, I recommend enabling the `Echo` option in AeroDyn to debug.

Best regards,

icnosvbovwb:
Dear @jjonkman 
Thank you for your answer. I am learning how to add interfaces from simulink recently. My purpose is to input external force from simulink in V3.1, so that I can try the module function without TMD and cancel the function of TMD. This force is equivalent to I add the force of the control effect of TMD, but how to add this force to the source code for calculation? Do you have any suggestions?

jjonkman:
Dear @icnosvbovwb,

Where do you want to apply the external force from Simulink (blade, nacelle, tower, substructure)?

You can always review the Simulink interface source code to see how to add additional input arguments to the Simulink S-Function.  And you can review the source code of the OpenFAST glue code to see how the load output from the Structural Control (StC) submodel of ServoDyn is currently mapped to the structural solvers (BeamDyn, ElastoDyn, or SubDyn).

Given that StC already supports user-defined load time histories, perhaps StC already supports the feature you need (although, StC is not currently set up so that the user-defined load time history is input from Simulink)?

Best regards,

icnosvbovwb:
Dear @jjonkman 
I want to add a force to the top of the tower, originally I can use the StC function, but now I don't turn on the Structural Control (StC) switch, I want to add a form of external force at the simulink interface, for example, Gen. Torque (Nm) and Power ( W) these two ports, to replace the STC function, but how to add this force to the tower top calculation?

jjonkman:
Dear @icnosvbovwb,

You can see in `SUBROUTINE ED_InputSolve()` of _FAST_Solver.f90_ how the the loads from a nacelle-based StC are mapped to ElastoDyn via a call to `Transfer_Point_to_Point()`.  You'd have to implement something similar for your new forces from Simulink.  To do that, you'd have to store the forces from Simulink on a mesh, similar to what is implemented in StC.

Best regards,

icnosvbovwb:
Dear @jjonkman
Thank you for your reply, I would like to ask, if the stc switch is turned off, are the variables related to TMD all 0 at this time? If there are 0, then at this time I can add the variable of external force on the basis of the original variable. For example, the force of TMD was originally transmitted through variable a. Now my understanding is that after the STC switch is turned off, a=0, so there is no need to delete the source code, then I will add a new external force variable b to become a+b, which is equivalent to the force value of variable a that turns on the stc switch.
There is another question, I add two force interfaces for transmission outside simulink. At this time, the s-function.c file does not need to be changed, right? Should I just define fixed interfaces in MAXInitINPUTS in FAST_Library.f90 and FAST_Library.h?

jjonkman:
Dear @icnosvbovwb,

I would say that if the StC model is not enabled, then the StC variables are never allocated, which is not the same as the variables equaling zero.  You could always enable StC, but apply zero loads (by selecting `StC_DOF_MODE` = 4 and setting the prescribed loads to zero), and then do as you propose.

I'm not sure I fully understand your question regarding Simulink.  Are you trying to add additional inputs from Simulink to the S-Function or convert one of the existing inputs to loads as you suggested in a prior post?

Best regards,

icnosvbovwb:
Dear @jjonkman 
I am thinking about your suggestion, I think when the TMD switch is turned off, the structure control function will not be calculated and called, I think of a method, that is, the incoming wind acts on the tower structure, at this time I add an external force , as the TMD control force on the top of the tower, can I add this TMD control force to the corresponding incoming wind position on the top of the tower before the incoming wind acts on the top of the tower?

icnosvbovwb:
I am thinking about your suggestion, I think when the TMD switch is turned off, the structure control function will not be calculated and called, I think of a method, that is, the inflow wind acts on the tower structure, at this time I add an external force , as the TMD control force on the top of the tower, can I add this TMD control force to the corresponding inflow wind position on the top of the tower before the inflow wind acts on the top of the tower?This is equivalent to reducing the wind at the top of the tower, which means that the control effect of TMD is effective.

icnosvbovwb:
Dear @jjonkman 
I am thinking about your suggestion, I think when the TMD switch is turned off, the structure control function will not be calculated and called, I think of a method, that is, the inflow wind acts on the tower structure, at this time I add an external force , as the TMD control force on the top of the tower, can I add this TMD control force to the corresponding inflow wind position on the top of the tower before the inflow wind acts on the top of the tower?This is equivalent to reducing the wind at the top of the tower, which means that the control effect of TMD is effective. Of course, I didn't turn on the TMD switch, just imagined that there is a TMD control force acting on the top of the tower. I'm not sure if this idea is correct

jjonkman:
Dear @icnosvbovwb,

I'm not sure I really understand your question.  Please clarify. 

Please note that the aerodynamic loads in OpenFAST are distributed across the blades and tower.

Best regards,

icnosvbovwb:
Dear @jjonkman 
Thank you for taking time out of your busy schedule to read my emails. Do aerodynamic loads distribute the force at each node according to the structural mass? I'm thinking that every particle of the tower is subject to turbulent wind, which can be reduced to every force on the particle. So I think the TMD module can directly transfer the force of TMD to the tower calculation from the simulink interface without turning on the switch, in this way to replace the force of TMD. If so, how can I subtract the external force I input from simulink from the wind force corresponding to the top of the tower?

icnosvbovwb:
Dear @jjonkman
Thank you for taking time out of your busy schedule to read my emails. Do aerodynamic loads distribute the force at each node according to the structural mass? I'm thinking that every particle of the tower is subject to turbulent wind, which can be reduced to every force on the particle. So I think the TMD module can directly transfer the force of TMD to the tower calculation from the simulink interface without turning on the switch, in this way to replace the force of TMD. If so, how can I subtract the external force I input from simulink from the wind force corresponding to the top of the tower?
Is it necessary to subtract the external force from the aerodynamic  module? Or is it necessary to subtract the external force in the elastic dynamic module?

jjonkman:
Dear @icnosvbovwb,

I'm still not really understanding.  E.g., why are trying to "subtract the external force"?

In the modular framework of OpenFAST, the aerodynamic applied loads in AeroDyn are computed distributed across an aerodynamics mesh, which need not be aligned with the structural mesh in ElastoDyn (or BeamDyn).  Internal mesh-mapping is used to transform the aerodynamic applied loads from the aerodynamic mesh to the structural mesh.

The StC module adds additional applied loads to the structure, which get summed with the aerodynamic applied loads.  Again, internal mesh-mapping is used to transform the StC loads from the StC mesh to the structural mesh.

Best regards,

jjonkman:
Dear @icnosvbovwb,

My understanding is that you want to specify loads in Simulink that gets applied to ElastoDyn and cancel out some of the aerodynamic applied loads.  Again, I would suggest enabling StC, but applying zero loads (by selecting `StC_DOF_MODE` = 4 and setting the prescribed loads to zero), and then modifying the source code of the OpenFAST glue code and Simulink interface as discussed in correspondence above so that the applied loads are calculated within Simulink and summed with the prescribed loads from StC that are zero before these loads are mapped to ElastoDyn.

Best regards,

jjonkman:
Dear @icnosvbovwb,

I'm not understanding again.  The aerodynamic loads are already mapped as applied loads to ElastoDyn, but they are applied as loads distributed across the rotor.  Do you want to change how the aerodynamic loads are mapped to ElastoDyn?  If so, how and why?

And when you say that you want your external force added to the aerodynamic part, how is this different than adding a load that gets summed with other loads applied to ElastoDyn?

Best regards,

icnosvbovwb:
Dear @jjonkman 
Ok, I try to add a variable in ElastoDyn, this variable can receive the external force data transmitted by SIMULINK interface. This force acts on the top of the tower. As you said If there are multiple applied loads at the same ElastoDyn node, then the summation happens in the OpenFAST glue code. Where do I add this variable to the code at this point?
At the same time, I Set StC_DOF_MODE = 4 and setting the prescribed loads to zero), which means that the seven columns of data in "TimeForceSeries.dat" in the STC file need to be set to 0 (columns of time, FX, FY, FZ, MX, MY, MZ), right?

jjonkman:
Dear @icnosvbovwb,

> Ok, I try to add a variable in ElastoDyn, this variable can receive the external force data transmitted by SIMULINK interface. This force acts on the top of the tower. As you said If there are multiple applied loads at the same ElastoDyn node, then the summation happens in the OpenFAST glue code. Where do I add this variable to the code at this point?

You don't need to add a variable to ElastoDyn.  You should only have to change the OpenFAST glue code and the Simulink interface.  I would review how existing data is passed from Simulink to OpenFAST and mimic for your additional loads.  My post linked above explained where in the source code you can add these loads from Simulink to the StC load (which should otherwise be zero) that gets mapped to ElastoDyn: https://github.com/OpenFAST/openfast/pull/1023#issuecomment-1250987785.

> At the same time, I Set StC_DOF_MODE = 4 and setting the prescribed loads to zero), which means that the seven columns of data in "TimeForceSeries.dat" in the STC file need to be set to 0 (columns of time, FX, FY, FZ, MX, MY, MZ), right?

Correct.

Best regards,

icnosvbovwb:
Dear @jjonkman 
image
As you said, you can add these loads from Simulink to the StC load (which should otherwise be zero) and you'd have to implement something similar for your new forces from Simulink. To do that, you'd have to store the forces from Simulink on a mesh, similar to what is implemented in StC. How to add this force to the mesh, although the principle is generally understood, facing these codes, I don't know how to start? Now the key point is what you said , store this force in the corresponding tower top grid.

icnosvbovwb:
Dear @jjonkman 
I still have a question that the applied loads are calculated within Simulink and summed with the prescribed loads from StC that are zero before these loads are mapped to ElastoDyn. At this point there should be a variable to receive the external force from simulink, I named it ForceTLD. So is it need to define something like m_FAST%ExternInput%ForceTLD = InputAry(55)? I define REAL(ReKi) , ::ForceTLD in FAST_ExternInputType and define typedef ^FAST_ExternInputType ReKi ForceTLD in FAST_Registry.txt. Right？

icnosvbovwb:
Dear @jjonkman 
I have a question, why the force transmitted from simulink needs to be added to the zero specified load from StC, can I directly add this external load without opening StC? Or can it be understood that the external force transmitted from simulink is to replace the seven columns of 0 of StC with its own data calculation?As you said, you can add these loads from Simulink to the StC load (which should otherwise be zero) that gets mapped to ElastoDyn. I guess this external load from Simulink is added to the seven columns of StC data, which is equivalent to 7 column 0 data is updated to the data transmitted from simulink.

The external loads transmitted from simulink are calculated by adding to the highest node of the tower. So I need to modify the source code in TOWER. When it is assumed that StC_DOF_MODE=1 is turned on, does TMD calculate the highest node of the tower based on the global coordinate calculation of external force? In the same node of the ElastoDyn module, I am not sure which coordinate system each load is based on? I guess it is the global coordinate, because the wind load, and the StC load need to be in the same coordinate system before they can be added or subtracted.

jjonkman:
Dear @icnosvbovwb,

My suggestion to introduce an StC with zero load was simply my suggested approach to minimize the amount of code changes that would need to be made and to avoid violating aspects of the modular framework.

The ElastoDyn module is set up so that tower loads are applied through a module-level input that makes use of a mesh data structure.  To work with meshes, it is easiest to use existing internal mesh functionality, such as the `Transfer_Point_to_Point()` SUBROUTINE.  Because the StC submodel of ServoDyn is already set up to apply loads to tower nodes in ElastoDyn, I felt that what I suggested above is the easiest approach you could take.

StC does not make use of the ElastoDyn tower nodes directly, nor does it even know where the ElastoDyn tower nodes are located.  Rather, the location where the StC is applied is specified within StC (via the at-rest position); you can locate this point at the tower top, but it can be located anywhere.  StC will create a point mesh with a node at this location.  Within OpenFAST, the internal mesh mapping functionality is used to transfer the loads from the StC point mesh to the ElastoDyn tower mesh, but it would be easiest if you don't change any of this code.

Best regards,

icnosvbovwb:
Dear @jjonkman
Ok, thanks for your answer, you didn't answer the second question, it is about the control force transmitted from simulink, do I need to do coordinate transformation? Because I used the experiment to simulate the vibration of the node at the top of the tower, and the TMD was placed at the top of the tower. At this time, I used the mass and acceleration of the TMD to calculate the collected TMD control force. The control force generated by it needs to be added to seven columns of data files. Is it based on tower coordinates or global coordinates?

icnosvbovwb:
Dear @jjonkman 
I have some doubts whether the global coordinates and the tower coordinates are consistent. For example, the X direction of the global coordinates corresponds to the X direction of the tower coordinates, and there is no angle transformation, that is to say, do the three directions overlap?

jjonkman:
Dear @icnosvbovwb,

The global inertial frame coordinate and tower base coordinate system will have the same orientation in OpenFAST if the three platform rotations (roll, pitch, and yaw) are zero, e.g., for a land-based wind turbine with rigid foundation.

Best regards,

icnosvbovwb:
Dear @jjonkman 
image
I modified the code in the Tower part according to what you said. The two strings of codes y_SrvD%TStCLoadMesh(j)%Force and y_SrvD%TStCLoadMesh(j)%Moment on the right side of the picture are the initial data for transmitting the TimeForceSeries.dat file, both of which are 0 . The m_FAST%ExternInput%ForceTLD(1:3) and m_FAST%ExternInput%ForceTLD(4:6) transmitted from simulink add this 0 to update the data. The code is shown in the picture. I'm not sure if such modification is correct?

And I still have a question, there are 7 columns of data in the TimeForceSeries.dat file, the first column is time, and I define ForceTLD as an array type of (1:6), I am not sure if the addition is correct? Because I guess the code internal calculation will take the time out. So is it also 6 columns add 6 columns?

jjonkman:
Dear @icnosvbovwb,

Yes, that looks correct to me, assuming your loads from Simulink are specified in global coordinates and you only have one tower-based StC (so, `j` is always 1).

Time does not need to be explicitly defined in these equations because `SUBROUTINE ED_InputSolve()` is called at distinct time steps.

Best regards,

icnosvbovwb:
Dear @jjonkman 
image
When I compile, there are some problems with the modified two lines of code, but I don't know how to modify it

And I want to obtain the FASTv7 case about Seismic. The FASTv7 I downloaded from NREL does not contain the example of Seismic file(Examples\1_NRELOffshrBsline5MW_Onshore_udm_wo_baseline_wo_target…). Please help me(mailto: wjl473008299@qq.com)

icnosvbovwb:
Dear @jjonkman 
image
image
image
I modified the type INTENT (INOUT) of y_SrvD according to your suggestion, and there is still one error that has not been solved, that is, Force and Moment are displayed. Do they need to be changed to Force (1:3) and Moment (1:3), and isn't m_FAST%  ExternInput%ForceTLD just a 6-column array?

icnosvbovwb:
Dear @jjonkman
I confirm the size of the array I added, m_FAST%ExternInput%ForceTLD is 6-column array

icnosvbovwb:
Dear @jjonkman
m_FAST%ExternalInput%ForceTLD is 6 columns of data input by simulink, which are FX, FY, FZ, MX, MY, MZ.
u_SrvD%ExternalForceTLD is passed to ServDyn by FAST, they are the same data, the picture shows that I define u_SrvD%ExternalForceTLD=m_FAST%ExternalInput%ForceTLD

jjonkman:
Please share the ALLOCATE statements you used for variables `u_SrvD%ExternalForceTLD` and `m_FAST%ExternalInput%ForceTLD`.

Best regards,

icnosvbovwb:
image
image

icnosvbovwb:
image

jjonkman:
You have not actually shared an ALLOCATE statement.  Did you ALLOCATE these arrays?  If not, that is likely the problem.

Best regards,

icnosvbovwb:
image
image
image

icnosvbovwb:
Dear @jjonkman 
image
image
image
image
image
image
I modified the registration file {:} to {3} according to your suggestions. Because I think Force is three arrays and Moment is three arrays, I divided six arrays into two and transferred three arrays at the same time. But there are more errors than before, so I don't know what to do. The modified code picture is shown in the figure.

icnosvbovwb:
Dear @jjonkman 
Thank you for your suggestions. I successfully solved the problem according to your suggestions. Then I recently started to try to add the seismic function in SoilDyn.f90. Do you have any specific suggestions? I guess UserPtfmLd_ Seismic.f90 should be modified, but I don't know how to start.

jjonkman:
Dear @icnosvbovwb,

NREL has not yet coupled the Seismic functionality of FAST v7 into OpenFAST, and so, cannot provide detailed guidance on doing that.  That said, similar questions have been asked before, e.g. see: https://github.com/OpenFAST/openfast/issues/617 and on our forum, e.g. see: https://forums.nrel.gov/t/running-examples-with-the-seismic-module/1846.

Best regards,

icnosvbovwb:
Dear @jjonkman 
image
image
I modified the code according to your method and compiled it successfully, but when I used it for simulink calculation, this error will occur. The first figure shows that StrucCtrl was run and terminated.

jjonkman:
Dear @icnosvbovwb,

I'm not following; I don't see that you actually shared an error.

Best regards,

icnosvbovwb:
Dear @jjonkman 
I output these lines of code in the picture. The result is in the OPENFAST91. txt file. As shown in the figure below: here75 and write (91, *) m_ FAST% ExternInput% ForceTLD (1)=0000000 E+00. I found that u_ SrvD%ExternalForceTLD(1)   = m_ FAST% ExternalInput% ForceTLD (1) is not executed, which indicates that data is transferred from outside, but there is no space to assign values. But in InputType ReKi ExternalForceTLD {:} -- and InputType ReKi ExternalMomentTLD {:} -- ，I have been defined the space in ServoDyn_ Registry.txt, so I don't konw  where there is an error. I request your help and support.
image
image

icnosvbovwb:
Dear @jjonkman 
image
image
This is the source code I finally modified.I have made two changes in this file. The first pitcure is to add two lines of code according to your suggestions, and the second pitcure is SUBROUTINE SrvD_ If (ALLOCATED (u_SrvD% ExternalForceTLD)) then in SetExternalInputs (p_FAST, m_FAST, u_SrvD).

icnosvbovwb:
Dear @jjonkman 
> As I mentioned in a prior post:
> 
> > loads on a mesh are stored in the global inertial frame coordinate system. You'll have to apply a rotation matrix if the loads you are calculating are specified in some other local coordinate system.
I'm sorry some records were deleted incorrectly.
As you said，The global inertial frame coordinate and tower base coordinate system will have the same orientation in OpenFAST if the three platform rotations (roll, pitch, and yaw) are zero, e.g., for a land-based wind turbine with rigid foundation.（5MW_Onshore or 5MW_OC3Monopile)

How about an offshore wind turbine? The three platform rotations (roll, pitch, and yaw) are not zero. How do I deal with the rotation matrix? I don't know where I can find this rotation matrix? I don't know where to add this rotation matrix. Can you provide a method?

jjonkman:
Dear @icnosvbovwb,

I'm not sure I understand enough about what you are doing to provide clear guidance.  But the direction cosine matrix (the transpose of the rotation matrix--i.e., the transformation matrix such that premultiplication of a vector in global coordinates will convert the vector to local coordinates--is stored in the `Orientation` field of the platform motion point mesh output of ElastoDyn (`PlatformPtMesh`).  This matrix--dependent on the platform roll, pitch, and yaw angles--is documented as Eq. (2-2) in my PhD-thesis turned NREL report: http://www.nrel.gov/docs/fy08osti/41958.pdf.

Best regards,

icnosvbovwb:
Dear @jjonkman 
image
image
Thank you for your instructions. The above figure is the method you suggested before. I will describe the experimental work I am doing in detail. Suppose the experimental object is ITI barge offshore wind turbine, and a TMD is set on the tower top. I turn on StC_ DOF_ MODE=4, and set 6 columns of data as 0 (FX, FY, FZ, MX, MY, MZ), time is not 0, it is 0, 0.0125, 0.025... I added two interfaces outside the simulink to transfer TMD Force and Moment, and modified the source code to transfer the two variables of the simulink to y_ SrvD% TStCLoadMesh (j). 

Ｑ１：At this time, I'm a little confused about the coordinates of forces and moments transmitted through simulink. I don't know whether it is global or local. Because I output the tower top speed to TMD through Ｏｕｔｄａｔａ to make ＴＭＤ vibrate, collect TMD forces and moments, and transmit TMD forces and moments to y_ SrvD% TStCLoadMesh (j) through simulink for summation calculation．　So I'm not sure whether to apply global or local coordinates．

Ｑ２：If the tower top speed is applied to local coordinates, how will I perform rotation coordinate conversion? Do you have a specific method?　Will it be necessary to modify the source code？

Ｑ３：If I place TMD respectively in the platform substructure, tower and engine room for vibration reduction, will the position coordinates of TMD correspond to their respective local coordinate systems? For example, does the platform take the sea level as the zero point（０，０，０）, the tower base take(10m above the sea level) as the zero point（０，０，１０）, and the cabin take the tower top as the zero point（０，０，８７．６）?　If I set NumTStC=1, turn on the tower control switch and set StC_ P_ Z=2, where TMD is located in the platform? But actually suspended from the tower?

jjonkman:
Dear @icnosvbovwb,

Here are my responses:
Q1. The answer depends on which ElastoDyn outputs you are using to output the tower-top velocity.  Which ones are you using?  This documentation lists the coordinate system of each ElastoDyn output: https://openfast.readthedocs.io/en/main/_downloads/3f19498a5dc774461e022b671ff01ec6/OutListParameters.xlsx.
Q2. I explained in my post above how to convert a vector from global to local coordinates.  Using the transpose of the direction cosine matrix (DCM) provides the rotation matrix, through which premultiplication of a vector in local coordinates will convert the vector to global coordinates.  The DCM for the tower top will depend on the tower-top rotations, and if you have enabled the platform DOFs as well, will depend on those as well.
Q3. Yes.  For a tower-based StC, the StC location is positioned relative to the tower-base.  If the tower-base is located at (0,0,10) m, then an StC located at (StC_P_X,StC_P_Y,StC_P_Z) = (0,0,2)m would be located at (0,0,12) m in global coordinates.

Best regards,

icnosvbovwb:
Dear @jjonkman 
Thanks for your reply.
Q1: If I output this variable YawBrTVxp, this should be the velocity in the fore-aft directions of the tower top, and it applies absolute coordinates, then I don't need to change the coordinate system at this time, Just transfer the external force and moment into  y_ SrvD% TStCLoadMesh (j) ，right?

Q2:  If I set NumSSttC=1, this is platform structure control and set StC_ P_ Z=20, the TMD is actually located at the tower position. If I set NumTStC=1, this is the tower structure control, set StC_ P_ Z=10, this is also the tower position, so I have a question, should the responses under these two control modes be the same? But the results I get are different. I don't know what the problem is?
Although there are two different control modes (NumSStC,NumTStC), I think the TMD is already in the same position and the control effect should be the same in theory, but I get different TTDspFA results in practice.

jjonkman:
Dear @icnosvbovwb,

Here are my responses:
Q1: ElastoDyn output YawBrTVxp is expressed in the tower-top/base-plate (p) coordinate system, which is a local coordinate system.  You'd need to convert this velocity to global coordinates if that is what you want.

Q2: The substructure-based StC will apply loads to the substructure while the tower-based StC will applied loads to the tower.  Even if location of both StCs is the same, the responses may still differ as a result of the point of application of the StC load.

Best regards,

icnosvbovwb:
Dear @jjonkman 
668155913e62d51ed3c0bb2aa3ba16c
This is the schematic diagram of my experiment. According to what you mentioned “You'd need to convert this velocity to global coordinates if that is what you want.” My understanding is like this：I need to convert this variable（YawBrTVxp）to the value of the global coordinate and provide it to the TLD as speed to generate vibration. At this time, I measure the external force of the TLD and transmit it to the two interfaces of the simulink. These two interfaces transmit Force and Moment to y_SrvD%TStCLoadMesh(j)%Force and y_SrvD%TStCLoadMesh(j)%Moment, right?  But how do I convert this output variable（YawBrTVxp） to a value in global coordinates?

jjonkman:
Dear @icnosvbovwb,

I've already summarized how to convert a vector from local to global coordinates.

Best regards,

icnosvbovwb:
Dear @jjonkman 
I am a new Fortran learner. I found y%PlatformPtMesh%Orientation(:,:,1) in ElastoDyn. This should be the DCM. How do I perform pre multiplication in m%AllOuts(YawBrTDzp) =  DOT_PRODUCT( rOPO, m%CoordSys%b2 )? Please tell me how to modify the source code?

Another question is that suppose I don't want to output the speed at the top of the tower. What about the speed at any position of the tower? For example, the speed at 20m and 30m of the tower.

jjonkman:
Dear @icnosvbovwb,

You can use the Fortran intrinsic function `MATMUL()` to premultiply a matrix with a vector and you can use the intrinsic function `TRANSPOSE()` to take the transpose of the DCM.  But `m%AllOuts(YawBrTDzp)` is a scalar, not a vector.

While the velocity of the tower nodes is certainly calculated within the source code, ElastoDyn does not have write outputs for the velocity of tower nodes, so, accessing these velocities would certainly require a source code change.

Best regards,

icnosvbovwb:
Dear @jjonkman 
image
I have modified the following code and am not sure whether it is correct.
 m%AllOuts(YawBrTVxp) =  MATMUL( TRANSPOSE( y%PlatformPtMesh%Orientation(:,:,1) ), m%AllOuts(YawBrTVxp) )

jjonkman:
Dear @icnosvbovwb,

I see two problems with your change: (1) `m%AllOuts(YawBrTVxp)` is a scalar, not a vector and (2) `y%PlatformPtMesh%Orientation(:,:,1)` represents the coordinate system of the platform (floating substructure), not the tower-top / base-plate coordinate system (p).  Please note that velocity vector `m%RtHS%LinVelE0` is already in global coordinates, except that the second element is along global Z and the third element is along negative global Y (the global coordinate system used within ElastoDyn is different than the global coordinate system used by OpenFAST).  So, if you want write outputs `YawBrTVxp`, `YawBrTVyp`, and `YawBrTVzp` expressed in global coordinates instead of local p coordinates, you can use:
```
m$AllOuts(YawBrTVxp) = m%RtHS%LinVelE0(1)  ! Global X
m$AllOuts(YawBrTVyp) = -m%RtHS%LinVelE0(3) ! Global Y
m$AllOuts(YawBrTVzp) = m%RtHS%LinVelE0(2)  ! Global Z
```

Best regards,

icnosvbovwb:
Dear @jjonkman 
Thank you for your detailed reply, which is very helpful to me. I will replace the original deletion with your suggested source code, that is, replace m%AllOuts(YawBrTDxp) = DOT_PRODUCT( rOPO, m%CoordSys%b1) with m%AllOuts(YawBrTVxp) = m%RtHS%LinVelE0(1).

icnosvbovwb:
> Dear @icnosvbovwb,
> 
> You can use the Fortran intrinsic function `MATMUL()` to premultiply a matrix with a vector and you can use the intrinsic function `TRANSPOSE()` to take the transpose of the DCM. But `m%AllOuts(YawBrTDzp)` is a scalar, not a vector.
> 
> While the velocity of the tower nodes is certainly calculated within the source code, ElastoDyn does not have write outputs for the velocity of tower nodes, so, accessing these velocities would certainly require a source code change.
> 
> Best regards,

Dear @jjonkman 
I set NumSSttC=1, StC_ DOF_ MODE=3，StC_  P_  Z=87.6m, this is platform structure control, is TLCD placed at the top of the tower at this time? If NumTStC=1, StC_ DOF_ MODE=3，StC_  P_  Z=77.6m This is tower control, is TLCD placed at the top of the tower at this time? This problem is that TLCD uses these two types of controls (NumSSttC and NumTStC). Although they are at the same location, the applied load positions are different. For example, platform control (NumSSttC=1) adds the load of TLCD to the platform calculation, and tower control (NumTStC=1) adds the load of TLCD to the tower top calculation, right?
Assuming my understanding is correct, then the position of TLCD is at the top of the tower. Is the speed of TLCD at the top of the tower under these two controls the same? Does this mean that these two types of controls (NumSSttC and NumTStC) drive TLCD motion at the same speed?

icnosvbovwb:
Dear @jjonkman 
I set NumSSttC=1, StC_ DOF_ MODE=3，StC_  P_  Z=87.6m, this is platform structure control, is TLCD placed at the top of the tower at this time? If NumTStC=1, StC_ DOF_ MODE=3，StC_  P_  Z=77.6m This is tower control, is TLCD placed at the top of the tower at this time? This problem is that TLCD uses these two types of controls (NumSSttC and NumTStC). Although they are at the same location, the applied load positions are different. For example, platform control (NumSSttC=1) adds the load of TLCD to the platform calculation, and tower control (NumTStC=1) adds the load of TLCD to the tower top calculation, right?
Assuming my understanding is correct, then the position of TLCD is at the top of the tower. Is the speed of TLCD at the top of the tower under these two controls the same? Does this mean that these two types of controls (NumSSttC and NumTStC) drive TLCD motion at the same speed?

jjonkman:
Dear @icnosvbovwb,

I agree with your response to Q3, which answers Q1 and Q2.  Regarding Q4, the input to the TLCD would not be the same in these two cases because for a substructure-based TLCD (NumSStC = 1), the input to the TLCD is coming from the substructure (so, a large, rigid moment arm between the substructure and TLCD would be assumed) and for the tower-based TLCD (NumTStC = 1), the input to the TLCD is coming from the tower (and so, includes tower flexibility, when enabled).

Best regards,

icnosvbovwb:
Dear @jjonkman 
If it is platform control  (NumSStC = 1), TLCD is driven by this rigid arm to move, and this speed essentially should be the motion speed of the platform, transmitted to TLCD through the rigid arm to make TLCD move. If it is tower control  (NumTStC = 1), TLCD is driven by the speed of the tower top to move.
Is it correct? Because I want to know what speed drives TLCD movement in these two control modes, respectively.

jjonkman:
Yes, that is correct.

jjonkman:
Dear @icnosvbovwb,

I'm sorry, but I'm not familiar enough with the design of TLDC systems to comment if your response is expected given the inputs that you've set.

Best regards,

icnosvbovwb:
Dear @jjonkman 
I want to output the speed of the TLCD drive in the cabin, which means this speed is the tower top speed. I placed the TLCD at the top of the tower. So should I output YawBrTVxp or QD_TFA1? Because the speed displayed by the YawBrTVxp parameter is  abnormal, I have calculated the YawBrTVxp of the ITI Barge wind turbine with a turbulent wind effect of 11.4m/s, and it can reach a maximum speed of 7.5m/s approximately. What is your opinion on this?
I guess the average wind speed is 11.4m/s, which shouldn't cause such a large speed at the top of the tower. So my question is, are these two parameters(YawBrTVxp, QD_TFA1) the speed at the top of the tower? If so, is this 7.5m/s speed normal?
image
image
image
image

jjonkman:
Dear @icnosvbovwb,

ElastoDyn output `YawBrTVxp` is the absolute velocity of the yaw bearing expressed in the x-axis of the tower-top coordinate system, which is what I expect you want.  `QD_TFA1` is the first-time derivative of the degree of freedom (DOF) associated with the first bending mode of the tower.   `QD_TFA1` could be less than `YawBrTVxp` because (1) other DOFs play a role in the tower-top velocity (such as the second bending mode and the platform DOFs) and (2) the coordinate system is a bit different between the two outputs.

Best regards,

icnosvbovwb:
Dear @jjonkman 
image
image
I have added two interfaces (Force and moment）and modified two rows of data based on your suggestion.（The second pitcure）One interface transmit three numbers in the XYZ direction. Although these two interfaces have been added, the calculation results of the data transmission from both interfaces are the same as the data without control, which means that the two interface data I have added have no effect. The data from these two interfaces cannot participate in the calculation.

My experimental idea is to collect some shear force from the testing equipment, which is TLCD, and place it on the top of the tower. I use the tower top speed of the wind turbine as the speed of the external specimen driven by the motor, and then use sensors to collect the control force of the TLCD, forming a closed cycle calculation.

I could have set up StC_ DOF_ MODE=3, which allows for calculations in the simulation environment. Now I have chosen to turn off StC_ DOF_ MODE=3. Because I want to achieve this calculation through experimentation. I have modified two interfaces and allowed them to transmit shear and bending moment, but you previously suggested opening StC_ DOF_ MODE=4. What is the principle of doing this with ? Is open StC _ DOF_ Mode=4, will the 6 transmitted data (FX FY FZ MX MY MZ)be automatically assigned to the TimeForceSeries.dat file? I thought the data transmitted from these two interfaces would be automatically assigned values, but now it seems that it is not involved in the calculation with the force at the top of the tower, and I am not quite sure where the source code needs to be modified.

icnosvbovwb:
> Dear @icnosvbovwb,
> 
> I see two problems with your change: (1) `m%AllOuts(YawBrTVxp)` is a scalar, not a vector and (2) `y%PlatformPtMesh%Orientation(:,:,1)` represents the coordinate system of the platform (floating substructure), not the tower-top / base-plate coordinate system (p). Please note that velocity vector `m%RtHS%LinVelE0` is already in global coordinates, except that the second element is along global Z and the third element is along negative global Y (the global coordinate system used within ElastoDyn is different than the global coordinate system used by OpenFAST). So, if you want write outputs `YawBrTVxp`, `YawBrTVyp`, and `YawBrTVzp` expressed in global coordinates instead of local p coordinates, you can use:
> 
> ```
> m$AllOuts(YawBrTVxp) = m%RtHS%LinVelE0(1)  ! Global X
> m$AllOuts(YawBrTVyp) = -m%RtHS%LinVelE0(3) ! Global Y
> m$AllOuts(YawBrTVzp) = m%RtHS%LinVelE0(2)  ! Global Z
> ```
> 
> Best regards,

Dear @jjonkman 
If I turn on StC_ DOF_ MODE=4, set NumTStC=1, and the predefined external force is acting on the tower control. You suggest that I modify the source code to add the external force of SIMULINK to the predetermined 6 columns of data (TimeForceSeries. dat). 
Q1: Is this predefined external force coordinate a local coordinate or a global coordinate?
I collect the forces and moments of external experimental objects. 
Q2：Is this force and moment in local coordinates or global coordinates? 
Q3: Assuming it is a local coordinate, does it require coordinate conversion before adding this force and moment? 
Q4：How to convert it specifically?

jjonkman:
Dear @icnosvbovwb,

Your first four questions seem to be focused on your own experiment, so, they would be better answered by you than me.

Regarding (5), the tower base coordinate system in ElastoDyn is aligned with global coordinates when the platform motion is zero.

Regarding (6), the write outputs available in the `OutList` of ElastoDyn are used only for writing to the output file and are not used for other purposes.

Best regards,

icnosvbovwb:
> Dear @icnosvbovwb,
> 
> Your first four questions seem to be focused on your own experiment, so, they would be better answered by you than me.
> 
> Regarding (5), the tower base coordinate system in ElastoDyn is aligned with global coordinates when the platform motion is zero.
> 
> Regarding (6), the write outputs available in the `OutList` of ElastoDyn are used only for writing to the output file and are not used for other purposes.
> 
> Best regards,

Dear @jjonkman 
Thanks for your reply.

icnosvbovwb:
Dear @jjonkman 
image

I would like to ask if these external forces (ForceTLCD and MomentTLCD) are added to the openfast calculation, are their symbols positive or negative? Because I am not sure what symbols are used to participate in the open fast calculation of the added forces and moments.

jjonkman:
Dear @icnosvbovwb,

To answer your question, please clarify how you've modified the source code of OpenFAST to apply the force and moment signals from Simulink to the structural module of OpenFAST.

Best regards,

jjonkman:
Dear @icnosvbovwb,

It looks like you are directly summing your commands from Simulink to the `y_SrvD%TStCLoadMesh` output of the StC module.  This output is expressed in the global coordinate system and can be thought of as the load applied to the tower from the StC.

Best regards,

icnosvbovwb:
Dear @jjonkman 
image
Another question is:  Is the TStC1_Myi of TLCD calculated by TStC1_Fxi ? What is the computational relationship between them specifically?

jjonkman:
Dear @icnosvbovwb,

I'm not an expert on the TLCD functionality of the StC submodel of ServoDyn.  I would suggest reviewing the TLCD theory basis documented here: https://openfast.readthedocs.io/en/main/source/user/servodyn-stc/StC_TLCD_Theory.html.  I don't see that this theory explicitly states how the TLCD equations of motion are related to the resultant StC force and moment outputs, but that is spelled out in the implementation that you can find in routine `StC_CalcOutput()` of source file _StrucCntrl.f90_.

Best regards,

jjonkman:
Dear @icnosvbovwb,

The `PrescribedForcesCoord` switch in the StC input file determines whether the applied loads are in the global or local coordinate system.  But if I understand your changes correctly (that you are adding `ForceTLD` and `MomentTLD` to the tower loads output from StC and applied as input to the tower in ElastoDyn within the OpenFAST glue code), the mapping within the OpenFAST glue code is always in global coordinates (as if `PrescribedForcesCoord` = 1, regardless of its setting).  So, `ForceTLD` and `MomentTLD` must also be in global coordinates.

Moreover, if you want to apply loads at two different locations along the tower (tower top and tower base), you must have two tower-based StCs (`NumTStC` = 2) and two sets of `ForceTLD` and `MomentTLD` inputs from Simulink, each added to the distinct instances of the tower-based StCs.

Best regards,

icnosvbovwb:
Dear @jjonkman 
image
image
image
image
image


Yes, your understanding is correct. The "specified force coordinate" switch in the StC input file determines that the applied load is in the global coordinate system. If I add ForceTLD and MomentTLD to the tower load of StC output and apply them as input to the tower in ElastoDyn in OpenFAST glue code, the mapping in OpenFAST glue code will always be in global coordinates (as if ProvidedForcesCourt=1).
If external loads are to be added at two different positions of the tower (top and base), two sets of ForceTLD and MomentTLD inputs from Simulink are required, each input being added to different instances of tower based Stc.

Q1: I don't quite understand this paragraph. Do you mean to establish 2 sets of external forces like Figure 5?
Q2: I have already set NumTStC=2, as shown in Figure 1, and there are two NRELOffshrBsline5MW_ServoDyn_StC.dat files, as shown in Figures 2 and 3. Does this mean that the openfast glue core needs to be modified again? 
Q3: What I previously modified is shown in Figure 4, which represents only the TLCD acting on the tower top and not the TMD acting on the tower base. So my question is: assuming my understanding is correct, how can I modify the glue core (y_SrvD%TStCLoadMesh(j)%Force(1:3,1)  = u_SrvD%ExternalForceTLD(1:3) + y_SrvD%TStCLoadMesh(j)%Force(1:3,1)  and y_SrvD%TStCLoadMesh(j)%Moment(1:3,1)  = u_SrvD%ExternalMomentTLD(1:3) + y_SrvD%TStCLoadMesh(j)%Moment(1:3,1)) of Figure 4 to add the TMD force acting on the tower base to the glue core?

jjonkman:
Dear @icnosvbovwb,

Overall, your approach looks OK,  But I agree that an extra set of inputs must be provided, as in your Figure 5; the source code in your Figure 4 needs to be changed.  You can see in your Figure 4 a loop through index `j`, i.e,.

`do j=1,size(y_SrvD%TStCLoadMesh)`

This is the loop through each tower-based StC, one for the tower-top and one of the tower base.  In your current code, you assign the same loads from Simulink to both of these StCs, so, both the tower top and tower base are receiving the same loads.  You'll need to change this source code so that each StC receives unique inputs from Simulink.

Best regards,

icnosvbovwb:
Dear @jjonkman 
Thank you for replying to the message during your busy time. Thank you very much for your help.

icnosvbovwb:
Dear @jjonkman 
image
image
I tried to compile and encountered an error with the following image. Can you clarify how I will make specific modifications?
（IF j=1 THEN） is line 320

jjonkman:
Add parentheses around each test condition and using double equals, i.e., `IF (J==1) THEN` and `ELSEIF (J==2) THEN`

icnosvbovwb:
Dear @jjonkman 
image
Another new error has occurred after modifying the source code according to your suggestion. 
lIne 347 is  CALL Transfer_Point_to_Point( y_SrvD%TStCLoadMesh(j), MeshMapData%u_ED_TowerPtLoads, MeshMapData%TStC_P_2_ED_P_T(j), ErrStat2, ErrMsg2, u_SrvD%TStCMotionMesh(j), y_ED%TowerLn2Mesh )

jjonkman:
Can you share your your complete set of updates to the ED_InputSolve() routine?

icnosvbovwb:
Dear @jjonkman 
image

icnosvbovwb:
Dear @jjonkman 
Thank you for your help. Under your guidance, I have resolved the compilation issue.
image

PR #1032, opened 2022-03-07
AD: added more nodal outputs for debugging UA

bjonkman:
The 5 UA states for UAMod 4-6 can be output, primarily to debug UA in BEMT (UA in FVW will need a little more work to get these outputs). This PR also adds 3 nodal "Debug" outputs that can be used by developers to output quantities not otherwise available as output.

It also adds a `UA_Flag` parameter to the main AD data structure, to make logic easier (otherwise UA_Flag is in both FVW and BEMT submodules).

none

AD nodal outputs

These updates are primarily to help us align our code bases better.

This should not affect any tests.


andrew-platt (approved):
Looks good.  Thanks!

PR #1037, opened 2022-03-09
Bug fix: BEMT was disabled for negative inflow

bjonkman:
The check to turn off BEM for low TSRs should have been using `ABS(TSR)` instead of `TSR` so that BEM is not disabled for all negative TSR values.

https://github.com/OpenFAST/openfast/discussions/940

AeroDyn

Tests where TSR < -1.5 will have differences due to BEM now turning back on at that point.

**AD Timeseries Shutdown case (AD driver):**
bokeh_plot(4)
bokeh_plot(3)

ad_timeseries_shutdown

**AD BAR RNA motion case (AD driver):**
bokeh_plot(2)
bokeh_plot(1)
ad_BAR_RNAmotion

### DIFF
--- a/modules/aerodyn/src/BEMT.f90
+++ b/modules/aerodyn/src/BEMT.f90
@@ -1139,9 +1139,9 @@ subroutine check_turnOffBEMT(p, u, Weight, axInduction, tanInduction, FirstWarn)
    integer(IntKi)                                 :: i                  !< blade node counter
    integer(IntKi)                                 :: j                  !< blade counter
 
-   if( u%TSR < BEMT_upperBoundTSR ) then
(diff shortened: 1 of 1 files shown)


andrew-platt:
Merge this around the same time we do PR #932

bjonkman:
> Actually, it looks like you also updated the r-test commit previously. 
That "update" was just to get the submodule to point to the most recent dev commit, which it _should_ have done when I merged the branch (on the next commit). I just did an extra commit because the submodule initialization was being a pain on my computer today. 

Anyway, it should be pointing to your updated r-test commit now. Thanks for updating it!

PR #1043, opened 2022-03-14
AeroDyn: combine some FVW and BEMT output calculations

bjonkman:
FVW and BEMT duplicated some WriteOutput calculations, primarily ones that applied to the AeroDyn meshes. I removed those duplicated parts of the code and created a new subroutine that is called instead. This also ensures that both versions of the routine compute the same output parameters for the meshes (FVW computed a few extra outputs that weren't in the BEMT version of the code).

None

AeroDyn

This should not affect any test results (except perhaps some small numerical differences). 
- The only failing test on [my branch](https://github.com/bjonkman/openfast/actions/runs/1964182614) is a result of the intermittent issue with the AeroDyn.
- Local comparisons for the openfast tests fail only with small numerical differences on the cases that normally fail when build settings change.

### DIFF
--- a/modules/aerodyn/src/AeroDyn.f90
+++ b/modules/aerodyn/src/AeroDyn.f90
@@ -276,7 +276,7 @@ subroutine AD_Init( InitInp, u, p, x, xd, z, OtherState, y, m, Interval, InitOut
       if (Failed()) return;
       NumBlades(iR)          = InitInp%rotors(iR)%NumBlades
       p%rotors(iR)%NumBlades = InitInp%rotors(iR)%NumBlades
-      if (size(InitInp%rotors)>1) then
+      if (nRotors > 1) then
          p%rotors(iR)%RootName  = TRIM(InitInp%RootName)//'.AD.R'//trim(num2lstr(iR))
       else
          p%rotors(iR)%RootName  = TRIM(InitInp%RootName)//'.AD'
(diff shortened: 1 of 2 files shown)


ebranlard:
This is great, I've been wanting this for a while, but didn't dare to change the code to avoid conflicts with your code base.

I know @hkross has worked on unifying FVW and DMST (double multiple stream tube), for nodal outputs and maybe regular outputs as well. I don't know how much this would conflict, but it's definitely going in the right direction. 

Ultimately, if BEM was returning induced velocity vectors in global, we could unify all lifting-line codes I believe.

ebranlard (approved):

andrew-platt:
Interesting that a large number of AD driver tests fail in the `debug-ubuntu-20.04-gfotran-10` test suite:
```
	ad_EllipticalWingInf_OLAF
	ad_VerticalAxis_OLAF
	ad_BAR_CombinedCases
	ad_BAR_SineMotion
	ad_BAR_RNAMotion
	ad_Kite_OLAF
	ad_HelicalWakeInf_OLAF
	ad_timeseries_shutdown
	ad_BAR_OLAF
	ad_QuadRotor_OLAF
```

but only one test on the

andrew-platt:
Ignore my previous comment about failed cases -- I misread the results.  Only the usual failing AD driver case failed (one case on each gfortran test suite).  Merging!

PR #1045, opened 2022-03-15
OLAF: check for division by zero to avoid invalid calculations

bjonkman:
This PR adds some checks for potential division by zero in induction output calculations in OLAF and outputs zero in that case.

AeroDyn/OLAF

This should not affect any of the current reg_tests.


ebranlard (approved):
This looks good to me.

PR #1160, opened 2022-06-29
Correction to documentation on Extended Bladed DLL interface

andrew-platt:
This PR is ready to merge.

The figure in the documentation on the Extended Bladed interface on read the docs incorrectly showed the channel allocations for `TMD control` at channel 3001, but should have shown `Structural controls` starting at channel 2801.

none

This is a documentation only PR.


ebranlard:
I would have probably pushed to use svg instead of pptx for the figure. Tracking binary files increases the size of the repo with time.

andrew-platt:
Good point.  I happened to have a pptx laying around that the figure came from.  Sometime it should be converted to svg.

I expect some modifications to that file in the next few months, so I'll redo it then as SVG.

PR #1180, opened 2022-07-13
S-Function build errors

bjonkman:
The Simulink S-Function in OpenFAST v3.2.0 does not build. These are the changes I made to get it to build.

I have no idea if this will cause issues with other code that also uses `FAST_Library.h`

Here are the error message from the build:
.\..\..\modules\openfast-library\src\FAST_Library.h(29): error C2143: syntax error: missing ')' before '='
..\..\..\modules\openfast-library\src\FAST_Library.h(29): error C2072: 'FAST_Sizes': initialization of a function
..\..\..\modules\openfast-library\src\FAST_Library.h(29): error C2059: syntax error: 'type'
C:\<omitted path>\openfast\glue-codes\simulink\src\FAST_SFunc.c(206): error C2198:
'FAST_Sizes': too few arguments for call

### DIFF
--- a/glue-codes/simulink/src/FAST_SFunc.c
+++ b/glue-codes/simulink/src/FAST_SFunc.c
@@ -203,7 +203,7 @@ static void mdlInitializeSizes(SimStruct *S)
        FAST_AllocateTurbines(&nTurbines, &ErrStat, ErrMsg);
        if (checkError(S)) return;
 
-       FAST_Sizes(&iTurb, InputFileName, &AbortErrLev, &NumOutputs, &dt, &ErrStat, ErrMsg, ChannelNames, &TMax, InitInputAry);
+       FAST_Sizes(&iTurb, InputFileName, &AbortErrLev, &NumOutputs, &TMax, &dt, &ErrStat, ErrMsg, ChannelNames, &TMax, InitInputAry);
        n_t_global = -1;
        if (checkError(S)) return;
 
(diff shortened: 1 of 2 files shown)


andrew-platt:
I conformed that this solution works with Simulink on MacOS.

glue-codes/simulink/src/FAST_SFunc.c
```diff
@@ -204,6 +204,7 @@ static void mdlInitializeSizes(SimStruct *S)
        if (checkError(S)) return;
 
        FAST_Sizes(&iTurb, InputFileName, &AbortErrLev, &NumOutputs, &TMax, &dt, &ErrStat, ErrMsg, ChannelNames, &TMax, InitInputAry);
```
andrew-platt, line 206: Should this line be dropped?

bjonkman: yeah, it looks like it got committed strangely. Let me fix that.

bjonkman:
@rafmudaf , the optional arguments are the FAST Library aren't interoperable with C (Fortran has C bindings, but not C++):
```
2><path>\openfast\modules\openfast-library\src\FAST_Library.f90(86): warning #8809: An OPTIONAL or EXTERNAL dummy argument to a BIND(C) procedure is not interoperable.   [TMAX]
2><path>\openfast\modules\openfast-library\src\FAST_Library.f90(86): warning #8809: An OPTIONAL or EXTERNAL dummy argument to a BIND(C) procedure is not interoperable.   [INITINPARY]
```
I would really prefer to not have OPTIONAL arguments on those interface routines. I made an attempt to modify this in another branch: https://github.com/bjonkman/openfast/tree/b/SFunc_NoOptionalArgs, though I didn't update the cpp API for OpenFAST.

With conditional compilation flags, we can probably get away with using them for now, but it's not ideal.

bjonkman:
I thought this PR was going to also be merged in a v3.2.1.... is that still happening? @rafmudaf @andrew-platt

PR #1199, opened 2022-07-28
Small improvements for -VTKLin visualization outputs

ebranlard:

This addresses the VTK output features when OpenFAST is being restarted using -VTKLin, to generate VTK files for mode shape visualization. 
The changes include:
- Screen outputs when modes are being processed
- Creating VTK directory on restart (if user deleted)
- Allowing default surface outputs when AeroDyn is not used. 

I have been using our mode shape visualization features lately and felt like some small changes could help the process.

### DIFF
--- a/modules/openfast-library/src/FAST_Subs.f90
+++ b/modules/openfast-library/src/FAST_Subs.f90
@@ -3308,6 +3308,7 @@ SUBROUTINE SetVTKParameters(p_FAST, InitOutData_ED, InitOutData_AD, InitInData_H
    INTEGER(IntKi)                          :: ErrStat2
    CHARACTER(ErrMsgLen)                    :: ErrMsg2
    CHARACTER(*), PARAMETER                 :: RoutineName = 'SetVTKParameters'
+   INTEGER(IntKi)                          :: rootNode, cylNode, tipNode
 
 
    ErrStat = ErrID_None
@@ -3417,16 +3418,8 @@ SUBROUTINE SetVTKParameters(p_FAST, InitOutData_ED, InitOutData_AD, InitInData_H
             call move_alloc( InitOutData_AD%rotors(1)%BladeShape(k)%AirfoilCoords, p_FAST%VTK_Surface%BladeShape(k)%AirfoilCoords )
          end do
       ELSE
-#ifndef USE_DEFAULT_BLADE_SURFACE
-         call setErrStat(ErrID_Fatal,'Cannot do surface visualization without airfoil coordinates defined in AeroDyn.',ErrStat,ErrMsg,RoutineName)
-         return
-      END IF
-   ELSE
-      call setErrStat(ErrID_Fatal,'Cannot do surface visualization without using AeroDyn.',ErrStat,ErrMsg,RoutineName)
-      return
-   END IF
-#else
       ! AD used without airfoil coordinates specified
+      call WrScr('Using generic blade surfaces for AeroDyn (S809 airfoil, assumed chord, twist, AC). ')
 
          rootNode = 1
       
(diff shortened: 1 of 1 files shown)


jjonkman:
Thanks, @ebranlard.  I agree that these upgrades would be useful.  These improvements will pair nicely with the automation of Campbell diagram generation that we have plans to develop next year.

Regarding your questions:

- I agree that it would be good to have default blade visualization when AeroDyn v15 is not enabled.  Perhaps a simple rectangular cross section with the chordlength scaled from the blade length.
- I agree that adding an option to overwrite `VTK_fps` in the OpenFAST mode-shape visualization input file would be a good addition.  Having the same number of steps per mode shape when `VTKLinTim` = 2 makes sense (which is equivalent to using a higher `VTK_fps` for high-frequency modes).

ebranlard:
I've changed the code such that:
-  the number of outputs are the same for all modes when `VTKLinTim=2` (so far I've used VTK_FPS as a parameter for the number of outputs, but I will add that as an input to the VIZ file). 
 - ElastoDyn and BeamDyn use a rectangular cross section when AD is not present. 

I think we might need to work on the scaling of the modes on the matlab side. The beamdyn modes at high frequencies have low deflections compared to the ones at low frequencies. Interestingly, when I increase `VTKLinScale` too much, I received the following error: 
```
FAST_RestoreForVTKModeShape_Tary:FAST_RestoreForVTKModeShape_T:CalcOutputs_And_SolveForInputs:SolveOption1:FullOpt1_InputOutputSolve:LAPACK_SGETRF: U(          11,          11)=0. Factor U is exactly singular.
```

andrew-platt:
I'm a little puzzled why two of the python regression tests now fail.  This PR should have no effect on them.

```
33 - 5MW_ITIBarge_DLL_WTurb_WavesIrr_py (Failed)
34 - 5MW_TLP_DLL_WTurb_WavesIrr_WavesMulti_py (Failed)
```

bjonkman:
> I'm a little puzzled why two of the python regression tests now fail. This PR should have no effect on them.
> 
> ```
> 33 - 5MW_ITIBarge_DLL_WTurb_WavesIrr_py (Failed)
> 34 - 5MW_TLP_DLL_WTurb_WavesIrr_WavesMulti_py (Failed)
> ```

From the error message, it looks like they fail due to trying to copy the same AWT27 directories in parallel. They didn't make it to the part of the script that actually runs the cases:
```
-- Using gold standard files with machine-compiler type linux-gnu
33: Traceback (most recent call last):
33:   File "/home/runner/work/openfast/openfast/reg_tests/executePythonRegressionCase.py", line 127, in <module>
33:     shutil.copytree(os.path.join(moduleDirectory, data), dataDir)
33:   File "/opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/shutil.py", line 568, in copytree
33:     return _copytree(entries=entries, src=src, dst=dst, symlinks=symlinks,
33:   File "/opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/shutil.py", line 467, in _copytree
33:     os.makedirs(dst, exist_ok=dirs_exist_ok)
33:   File "/opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/os.py", line 225, in makedirs
33:     mkdir(name, mode)
33: FileExistsError: [Errno 17] File exists: '/home/runner/work/openfast/openfast/build/reg_tests/glue-codes/python/AWT27'

34: -- Using gold standard files with machine-compiler type linux-gnu
34: Traceback (most recent call last):
34:   File "/home/runner/work/openfast/openfast/reg_tests/executePythonRegressionCase.py", line 127, in <module>
34:     shutil.copytree(os.path.join(moduleDirectory, data), dataDir)
34:   File "/opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/shutil.py", line 568, in copytree
34:     return _copytree(entries=entries, src=src, dst=dst, symlinks=symlinks,
34:   File "/opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/shutil.py", line 467, in _copytree
34:     os.makedirs(dst, exist_ok=dirs_exist_ok)
34:   File "/opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/os.py", line 225, in makedirs
34:     mkdir(name, mode)
34: FileExistsError: [Errno 17] File exists: '/home/runner/work/openfast/openfast/build/reg_tests/glue-codes/python/AWT27'
```

ebranlard:
Maybe there is a `vtk` folder already present somewhere. 

Instead of using `shutil.copyTree`, I would use the function `rtl.copyTree` that I added to the rtestlib.py, I think it should be more robust if file already exists.

andrew-platt:
I noticed the `rtl.copyTree` some time ago -- it's a good addition.  A few of the `execute*` scripts were updated to use it, but not all.   We can either update that here, or perhaps it would be better to do a quick PR to update these and a few other minor issues (a couple of `_py` cases are failing, but their regular OpenFAST versions were temporarily disabled for stability reasons).

deslaughter:
I think you’ll always have a race condition when copying these files from
running tests concurrently. Using rtl.copyTree makes it less likely to
occur because it checks the existence of each file instead of the entire
directory. However, another test can still begin writing a file in between
the existence check and the call to copy. I think it would be better to
copy these shared directories with the build system instead of the python
scripts.

bjonkman:
I agree. I've run into errors due to this race condition several times already. We could also copy all the files in a separate script that runs serially, prior to running any of the test cases. Would be a nice option for those of us that use the manual python scripts.

deslaughter:
A manual python script could be incorporated into CMake. That should work
for both cases.

andrew-platt:
The change to use `rtl.copyTree` is in PR #1244.  That provides a partial solution to the race condition.  I'll rerun checks after that is merged.

andrew-platt:
Minor error handling update with suggestion from @deslaughter.

deslaughter (approved):

PR #1203, opened 2022-07-29
Add parallel branches to GitHub Actions

rafmudaf:
This pull request reworks the GitHub Actions configuration so that jobs build on each other and parallelization between jobs is supported.

Screen Shot 2022-07-29 at 2 58 48 PM

Initially, three jobs launch to compile various parts of the project. Then, subsequent jobs use the first ones to either run tests or build other more time consuming parts of OpenFAST. Specifically, this allows for compiling the OpenFAST and FAST.Farm drivers concurrently without shared resources. Finally, the more time consuming regression tests also get their own job and run concurrently. This structure results in a 20-30 minute improvement in time to completion from [v3.2.0](https://github.com/OpenFAST/openfast/actions/runs/2664415181). The performance of the GitHub-hosted runners are variable, but this generally completes in 45-50 minutes.

See the results live [here](https://github.com/rafmudaf/openfast/actions/runs/2762693528).

GitHub Actions and some CMake configuration to support the changes to Actions

### DIFF
--- a/.github/workflows/automated-dev-tests.yml
+++ b/.github/workflows/automated-dev-tests.yml
@@ -192,6 +192,7 @@ jobs:
           cmake --build . --target openfast_cpp -- -j ${{env.NUM_PROCS}}
           cmake --build . --target openfastcpp -- -j ${{env.NUM_PROCS}}
           cmake --build . --target ifw_c_binding -- -j ${{env.NUM_PROCS}}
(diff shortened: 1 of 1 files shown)


andrew-platt (approved):
The diagram is super useful in understanding this.  I assume that the long `OF` regression tests are the ones that were split out to gain speed (they were a bottleneck when run in series).

It is interesting to see that the linearization results changed very slightly (1E-18 etc become zero in the `.lin` files)

PR #1211, opened 2022-08-12
Use dt_out when storing OpenFAST outputs in Python interface

deslaughter:
This pull request is ready for merging.

All regression tests are passing and development is complete. No additional documentation is required.

This pull request resolves an issue where the openfast_library.py file was allocating too much memory on large simulations. 

None

- glue-codes/openfast/src/FastLibAPI.cpp
- glue-codes/openfast/src/FastLibAPI.h
- glue-codes/python/openfast_library.py
- glue-codes/simulink/src/FAST_SFunc.c
- modules/openfast-library/src/FAST_Library.f90
- modules/openfast-library/src/FAST_Library.h

The existing versions of openfast_library.py and FastLibAPI allocate an output array with a size NUM_OUTPUTS*(T_MAX/DT). This size is inconsistent with the output files which contain data written every DT_OUT seconds. The code changes in this PR cause the output data to be collected on the DT_OUT interval which reduces memory usage and matches current output files. DT_OUT was exposed to these libraries by adding it to the FAST_Sizes arguments which required changes to the Fortran, C, C++, Python, and Simulink libraries.

These changes also resolve a potential memory leak in FastLibAPI where the output data array was not freed by the parent class destructor. This was fixed by changing the output array to a vector of vectors which is automatically destroyed.

Developer testing showed a reduction is memory use due to allocation of smaller output arrays.

### DIFF
--- a/glue-codes/openfast/src/FastLibAPI.cpp
+++ b/glue-codes/openfast/src/FastLibAPI.cpp
@@ -56,7 +56,7 @@ void FastLibAPI::fast_init() {
         &t_max,
         &_error_status,
         _error_message,
-        channel_names
+        output_channel_names
     );
     if (fatal_error(_error_status)) {
         throw std::runtime_error( "Error " + std::to_string(_error_status) + ": " + _error_message );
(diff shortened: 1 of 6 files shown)


andrew-platt:
Will the addition of the `dt_out` variable passing require updates to SOWFA / AMR-Wind or the Simulink interfaces?

PR #1227, opened 2022-08-23
Seg Fault due to hub model and external inflow

psakievich:

The hub index is 1 in most of the code (first entry in the array), but here it is the last one.  I've had segfaults in nalu-wind and amr-wind that both trace back to this line of code when running actuators with the hub turned on.

Interestingly we only see this issue with GCC, and in testing a `RelWithDebugInfo` build of OpenFAST the failures were intermittent as many memory leaks are.

https://github.com/Exawind/nalu-wind/issues/979

Any turbine run using the external inflow and hub model.

Pinging @andrew-platt @rafmudaf @gantech 

I am not an openfast developer and am not aware of the tests. I assume this is just not getting covered in the test suite.  We've had a segfault in the nalu-wind test suite for months now.

https://my.cdash.org/test/60805600

### DIFF
--- a/modules/openfast-library/src/FAST_Solver.f90
+++ b/modules/openfast-library/src/FAST_Solver.f90
@@ -607,6 +607,7 @@ SUBROUTINE AD_InputSolve_IfW( p_FAST, u_AD, y_IfW, y_OpFM, ErrStat, ErrMsg )
       end if
       
       if (u_AD%rotors(1)%NacelleMotion%NNodes > 0) then
+!        for cfd we will lump the hub and nacelle together
          u_AD%rotors(1)%InflowOnNacelle(1) = y_OpFM%u(1)
          u_AD%rotors(1)%InflowOnNacelle(2) = y_OpFM%v(1)
          u_AD%rotors(1)%InflowOnNacelle(3) = y_OpFM%w(1)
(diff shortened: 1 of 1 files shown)


andrew-platt (approved):
This change agrees with `OpenFOAM.f90` line 80 (https://github.com/OpenFAST/openfast/blob/v3.2.1/modules/openfoam/src/OpenFOAM.f90#L80).

andrew-platt:
I'm going to merge this into the `dev` branch.  It will be released with v3.3.0

bjonkman:
I think this is a fine temporary fix, but technically this last point is the nacelle and not the hub. The hub point (index 1) gets sent to ServoDyn (see `SrvD_InputSolve`), so it is omitted in subroutine `AD_InputSolve_IfW` and is the reason the node index starts at 2 there instead of 1. 

We added a nacelle point in AeroDyn for some other work (hydropower? + some syncing with Envision), but the OpenFOAM interface portion of that in OpenFAST apparently didn't get updated. I was actually working on a fix for this interface last week, but it will also require that everything interfacing with this (nalu-wind, amr-wind, and similar) will need to supply another point for wind at the nacelle.

psakievich:
Hmm this is interesting. I would request that you add an option to make the addition of the new point for the nacelle optional. From an actuator line implementation standpoint this will require us to add an additional actuator point, code infrastructure etc. This is inconsistent in terms of restarts and can potentially add larger impacts on the codes coupling to OpenFAST. @tonyinme what are your thoughts?

tonyinme:
hey all, @psakievich had a conversation offline and agreed that it is best to bundle the hub and nacelle contribution into one for the CFD coupling.

andrew-platt:
@psakievich, could you add a comment in the code about using the hub point as the nacelle point for the CFD coupling?  This could help prevent confusion when we are looking at that code later.  I'll merge after that.

psakievich:
sorry. I forgot to turn off the auto-formatting on my neovim config.  I don't ever work in fortran. will fix

bjonkman (approved):

PR #1228, opened 2022-08-24
r-test: Remove -m64 in CMAKE_Fortran_FLAG from r-test

andrew-platt:
This PR is an update to the r-test/dev head.  The only changes this introduces are in the CMake setup for r-test.


deslaughter (approved):
Looks good to me, all of the test cases passed.

PR #1086, opened 2022-04-11
MoorDyn v2 + shared moorings + wave propagation in FAST.Farm

mattEhall:
This pull request can be merged once tests are finalized.

**Feature or improvement description**
This pull request introduces MoorDyn v2 (a massive increase in mooring modeling capabilities) and updates to FAST.Farm to better model floating wind farms including with shared mooring systems.

MoorDyn improvements:
- New rigid-body Rod and Body object types (Connections renamed Points)
- Completed MoorDyn Driver
- Standardized input file format
- Active tensioning
- Linearization functions
- Nonlinear tension-strain relations for lines
- Viscoelastic capability for lines (for synthetic ropes)
- Line bending stiffness for power cables
- Built-in grid-based wave kinematics based on input wave elevation time series
- Built-in table-based steady current profile
- Water kinematics applied on Lines, Points, and Rods
- Rod free-surface hydrodynamics in all orientations (with wave stretching, fully nonlinear)
- Seabed bathymetry grid accounting for arbitrary seabed contours and effect of slope and change in depth
- Line-seabed friction modeling including static and kinetic friction coefficients
- Support for multiple input and output meshes for use with FAST.Farm

FAST.Farm improvements:
- "Mod_SharedMooring" - Alternate time stepping arrangement that allows array-level communication between FAST instances and an array-level MoorDyn instance that can be used for shared moorings and shared anchors as well as fully dynamic intra-array cables
- "Mod_WaveField" - Can account for individual turbine positions when computing incident waves and result loads so that phasing from wave propagation throughout the array is accurately modeled

**Impacted areas of the software**
- MoorDyn and how it is called within FAST (I/O meshes are now arrays)
- FAST.Farm (new alternative time stepping option) and FAST wrapper through to HydroDyn and MoorDyn (turbine reference positions in the array are passed down to the modules from FAST.Farm)

**Additional supporting information**
This is an updated PR of postponed PRs https://github.com/OpenFAST/openfast/pull/750 and https://github.com/OpenFAST/openfast/pull/747 along with many improvements done for various NOWRDC and ATLANTIS projects since then.

**Dependencies**
This depends on PR #999

**Test results, if applicable**
TBD

**ToDo**

- [ ] update `api_change.rst`
- [x] update baseline results
- [ ] link to MoorDyn C++ version documentation
- [x] merge #999

PR #1244, opened 2022-09-06
Reg-test scripts modification to help avoid race condition and cleanup of caselist

andrew-platt:
This PR is ready to be merged

As discussed in PR #1199, there is a frequent issue with race conditions when copying test case inputs over to the regression testing location.  This is caused by the `shutil.copytree` command that errors out if there are pre-existing files.  To get around this issue, the `rtl.copyTree` command is now used.  

Additionally, three test cases are dropped from the python interface testing and HydroDyn module testing.  These cases are equivalent to ones that had been dropped from main OpenFAST test case list as they do not agree sufficiently between platforms for the new testing methodology (likely due to numerical issues in HydroDyn).  These cases include:
- Python interface test:
    - `5MW_ITIBarge_DLL_WTurb_WavesIrr_py`
- HydroDyn module tests:
    - `hd_5MW_ITIBarge_DLL_WTurb_WavesIrr`
    - `hd_5MW_OC4Jckt_DLL_WTurb_WavesIrr_MGrowth`

See comment https://github.com/OpenFAST/openfast/pull/1199#issuecomment-1233158938 and following discussion in PR #1199 for context on the `rtl.copyTree`.
The dropped cases should have been dropped with PR #1217.

This only affects the running of test cases.

See above.

### DIFF
--- a/reg_tests/CTestList.cmake
+++ b/reg_tests/CTestList.cmake
@@ -254,7 +254,7 @@ of_fastlib_regression("AWT_YFree_WSt"                    "fastlib;elastodyn;aero
 
 # OpenFAST Python API test
 of_regression_py("5MW_Land_DLL_WTurb_py"                     "openfast;fastlib;python;elastodyn;aerodyn15;servodyn")
-of_regression_py("5MW_ITIBarge_DLL_WTurb_WavesIrr_py"        "openfast;fastlib;python;elastodyn;aerodyn14;servodyn;hydrodyn;map;offshore")
+#of_regression_py("5MW_ITIBarge_DLL_WTurb_WavesIrr_py"        "openfast;fastlib;python;elastodyn;aerodyn14;servodyn;hydrodyn;map;offshore")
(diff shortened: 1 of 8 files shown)


andrew-platt:
This may partially fix an issue in testing for PR #1086.

deslaughter:
The `rtest-interfaces` action failed because the `5MW_Land_DLL_WTurb_cpp` test couldn't find its `yaml` input file. I think that `.yaml` was one of the file types in the `excludeExt` list which was added to the Python files. Changing `.yaml` to `.sum.yaml` in the `excludeExt` list should fix the issue by excluding the regression test reference files, but still allowing the input file. However, this change would require modifying `rtestlib.py` to something other than `os.path.splitext` to get the extension. It may be better to remove `.yaml` from `excludeExt` in `executeOpenfastCppRegressionCase.py`.

While these changes will reduce the chances of the script failing due to simultaneous tests writing to the same file, they won't prevent it, as discussed in PR #1199.

andrew-platt:
Thanks for looking at this!  I agree this is not going to completely solve the race condition.

deslaughter (approved):
Changes look good and all tests are passing.

PR #1248, opened 2022-09-12
OpenFAST Registry: allow pointers

bjonkman:

The Registry was modified in the SeaState code development to allow pointers (instead of allocatable arrays) in some of the data types. This PR pulls out those code changes so it can be merged into dev before https://github.com/OpenFAST/openfast/pull/1008 is merged. As part of the registry change, error handling was also added to some auto-generated routines that had ignored errors before. After this was modified, we discovered that the error status wasn't set in the NWTC Library `FreeDynamicLib` routine, so that has also been fixed here.

#1008 

This affects all of the types files (including NWTC Library). Code will change, but results should not. 

These should not change.

### DIFF
--- a/modules/aerodyn/src/AeroAcoustics_Types.f90
+++ b/modules/aerodyn/src/AeroAcoustics_Types.f90
@@ -300,15 +300,27 @@ CONTAINS
     DstBladePropsTypeData%TEAngle = SrcBladePropsTypeData%TEAngle
  END SUBROUTINE AA_CopyBladePropsType
 
- SUBROUTINE AA_DestroyBladePropsType( BladePropsTypeData, ErrStat, ErrMsg )
+ SUBROUTINE AA_DestroyBladePropsType( BladePropsTypeData, ErrStat, ErrMsg, DEALLOCATEpointers )
   TYPE(AA_BladePropsType), INTENT(INOUT) :: BladePropsTypeData
   INTEGER(IntKi),  INTENT(  OUT) :: ErrStat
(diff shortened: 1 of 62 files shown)


andrew-platt:
@bjonkman, thanks for putting this in!

@deslaughter, this is related to what we are doing with InflowWind.

I'll try to take a look at it in the next week.

deslaughter (approved):
Small typo on line 496 of reg_parse.c, but everything else looks good.

bjonkman:
Good catch, @deslaughter. That typo should be fixed now. Thanks!

bjonkman:
@rafmudaf : We've addressed some of the issues that were in the original pointer code in #1008 in this PR. There are still some concerns to be aware of, but it doesn't necessarily "break" restart capability. Those comments in #1008 should probably be updated.

For existing code, this modified code will behave as before. The problem is that if you use two pointers that point to the same data, on restart those pointers will be pointing to _separate_ copies of the data. As long as the data it was pointing to is static (in the sense that modules are using its values, but not changing it), the only issue is that the restart will consume a larger memory footprint. If the developers expect one module to modify data and the other to use it, that would be a problem on restart. This isn't any different than how the mesh data structures behave with restart (and have always behaved this way). It's just important for module developers to keep in mind that the pointers have this limitation. 

Adding pointers to the Registry also puts more responsibility on developers to make sure that deallocate/nullify statements are used appropriately and consistently to prevent memory leaks and/or dangling pointers. But, that's _always_ something to watch out for when using pointers.

PR #1263, opened 2022-09-27
FAST.Farm WriteOutput: fix for Windows Intel OMP build

bjonkman:

FAST.Farm's OMP build on Windows with Visual Studio was failing on the `Farm_SetOutParam` routine (all other builds seem to be working okay). I updated the variables in there to use Fortran parameters (using the algorithm from https://github.com/OpenFAST/openfast/pull/1254), and that seems to have fixed the issue on Windows.

In the process, I also fixed the incorrect leading zeros on the parameter names for FAST.Farm in OutListParameters.xlsx.

[NREL Forum discussion reporting the issue](https://forums.nrel.gov/t/fast-farm/1887/391)

FAST.Farm, particularly Windows OMP

With this PR, I also am now building the Visual Studio OMP version of FAST.Farm with static libraries. Unfortunately, the executable still depends on libiomp5md.dll, so the binary will still need an Intel redistributable library, but the list of dependencies is significantly smaller and won't require the Visual Studio redistributable.

This should not change any results.


jjonkman (approved):
I discussed these changes with Bonnie and agree with the solution.

PR #1050, opened 2022-03-16
Fix aeroelastic stability analysis with BeamDyn

ptrbortolotti:
This PR is ready to merge once PR #962 is merged.

Many thanks to @michaelasprague, @ebranlard, @jjonkman, and @rafmudaf for their contributions in isolating the issue.  The majority of the testing was performed by @ebranlard  and @ptrbortolotti.

**Description**
This PR fixes two issues with BeamDyn.

1. revises the method used for small angles in the linearization process.  Previously, a logarithmic mapping had been applied for small angle changes.  This resulted in discontinuities in linearization involving BeamDyn, which caused large negative damping for any blade close to a +/-180 degree azimuth (178 to 182 degree window) - see #823.
To fix this issue, the logarithmic mapping call has been removed from the `PackMotionMesh` routine in the ModMesh module, and replaced with a small angle mapping for small angles, and DCM differencing method for large angles.

2. A limit is placed on the `QFit` parameter within BeamDyn.  This now limits the maximum order of the spline fit to the keypoints to 7 instead of `order_elem`.  Without this limit, the solver within BeamDyn would go unstable in some configurations or loadings resulting in incorrect solutions for the geometry, particularly in twist.  When BeamDyn would stay stable, this could appear as differences between blade tip motions for a simplified rotor (see #816 and #818).

**Impacted areas of code**
BeamDyn primarily

A few other modules were also modified for consistency with the documentation: some modules should use small angle difference and others a large angle DCM difference.

Small angle:
- HydroDyn
- SubDyn
- ElastoDyn platform and tower
- mooring modules

Large angle DCM:
 - Blades
 - AD
 - BeamDyn

**Test Results**
No existing test cases should be affected.

**Items to complete**

- [x] Merge PR #962

**Related issues**
closes #816 #818 #823

This may partially improve issue #366

PR #1275, opened 2022-10-07
Linear Trim Solution Improvements

bjonkman:

The packed operating point array that is used to see if the turbine has reached steady-state in a trim solution was modified in #1050 to assume that all orientation angles in the code are small (<25 deg). Since this is not a valid assumption, I have reverted that code to use log maps.  

Also, I have removed the rotational accelerations from the packed arrays for the trim solution. These can be noisy outputs that don't want to converge in this algorithm; since we are already making sure that the position and velocity values are converging, these accelerations should not also have to be in the convergence calculation.

For code readability, I changed the names of some arguments in the `GetOP` and `PackMeshMotion` routines.

Also, I fixed an issue where the SD operating point for inputs would not have been displayed correctly in linearization output files.

This is a follow-up to https://github.com/OpenFAST/openfast/pull/1158.

Linear trim solution (BD, ED, SD structural codes)

This should not change any r-test results. 

For other cases that use the trim solution, it is possible that it will now converge sooner than in previous versions of the code, possibly leading to insignificant differences in the steady-state solution.


jjonkman (approved):
I have discussed these changes with Bonnie and generally agree with the solution.

PR #1283, opened 2022-10-11
AD: add new projection method and BEM methods

ebranlard:
This pull request is ready to be merged. It has been reviewed by @bjonkman. 

To help the development of new BEM algorithms, with different projection methods and implementations, some flags are introduced to the AeroDyn code. The flags are currently "hidden" to the user, as we need more time to decide on the implementation, document it, and add tests. In the long term, this version should have limited code differences with Envision's version. 
 
AeroDyn

Two main toggles are introduced 
-	“AeroProjMod” to distinguish between the different coordinate system used: 
o	APM_BEM_NoSweepPitchTwist=1  "Original AeroDyn model where momentum balance is done in the WithoutSweepPitchTwist system" 
o	APM_BEM_Polar=2 "Use staggered polar grid for momentum balance in each annulus" 
o	APM_LiftingLine=3   "Use the blade lifting line (i.e. the structural) orientation (currently for OLAF with VAWT)" 
-	“BEM_Mod” to switch between different BEM implementations:
o	BEMMod_2D=0 "2D BEM assuming Cx, Cy, phi, L, D are in the same plane" 
o	BEMMod_3D=2 "3D BEM assuming a momentum balance system, and an airfoil system" 

The BEM time integration now uses the standard "utimes" vector instead of indices 1 and 2 which were hardcoded. 

The coordinate system "withoutSweepPitchTwist" is now renamed orientationAnnulus, and of type R8Ki. 

Test should match.

**Checklist**
- [x]  potentially wait for #957


modules/aerodyn/src/AeroDyn_AllBldNdOuts_IO.f90
```diff
@@ -766,7 +765,7 @@ SUBROUTINE Calc_WriteAllBldNdOutput( p, p_AD, u, m, m_AD, x, y, OtherState, Indx
          CASE ( BldNd_Mm )
             DO IdxBlade=1,p%BldNd_BladesOut
                DO IdxNode=1,p%NumBlNds   
-                  y%WriteOutput( OutIdx )  = m%M(IdxNode,IdxBlade)
+                  y%WriteOutput( OutIdx )  = m%Mz(IdxNode,IdxBlade)
```
bjonkman, line 768: @ebranlard , I think I had an error in the original code I sent you where `m%M` wasn't always set, but it looks like you fixed that in `SetOutputsFromBEMT`. I think this should still be m%M.

modules/aerodyn/src/AeroDyn_IO.f90
```diff
@@ -1928,7 +1928,7 @@ subroutine Calc_WriteOutput_BEMT()
 
             m%AllOuts( BNFl(   beta,k) ) =  m%X(j,k)*cp - m%Y(j,k)*sp
             m%AllOuts( BNFd(   beta,k) ) =  m%X(j,k)*sp + m%Y(j,k)*cp
-            m%AllOuts( BNMm(   beta,k) ) =  m%M(j,k)
+            m%AllOuts( BNMm(   beta,k) ) =  m%Mz(j,k)
```
bjonkman, line 1931: As above, I think this should still be `m%M`.

ebranlard:
Hi @bjonkman 

These outputs are indeed a bit confusing with the different coordinate systems. Thank you for this careful review!

PR #1317, opened 2022-11-08
Fix for Visual Studio builds with ADI

bjonkman:

This PR updates the build process so that ADI's registry runs in the VS solution files. It also 
- removes a few extrap/interp routines that aren't used (and don't need to be generated).
- removes non-standard tabs from the Fortran source code
- removes some unused variables from AD.

Visual Studio builds with code that contains the ADI module.

This does not change any results


andrew-platt (approved):

ebranlard:
I suspected the parallelization of FWrap Init to be buggy. I've removed it and the tests seems to pass now...

bjonkman:
Thanks, @ebranlard!

bjonkman:
FWIW: when I pulled up the GitHub Action artifacts of the failing FAST.Farm case (before Emmanuel reverted that change in dev), the plots are unreadable:

image

I haven't spent any time trying to figure out what's going wrong there, but thought I'd draw it to someone's attention. 

The other PRs that I submitted about the same time also failed on the same case as this one, but they didn't produce any .html file. So, something strange is going on with that.

ebranlard:
Yes, I've noticed this as well, I'm not quite sure what's going on. I ended up comparing the files in pyDatView instead.

andrew-platt:
I'm able to reproduce this locally as well.  
Screen Shot 2022-11-09 at 5 32 52 PM

PR #1319, opened 2022-11-09
Cleanup OpenFAST VTK output for HydroDyn

bjonkman:

I have cleaned up the HydroDyn meshes that are written as VTK files when all meshes are output. This update makes sure all 3 HD input meshes can be output, and their sibling meshes are also output. The names of these VTK files have been updated to match names in the code.

None.

VTK output of HD meshes

This does not change any test results.

### DIFF
--- a/modules/openfast-library/src/FAST_Subs.f90
+++ b/modules/openfast-library/src/FAST_Subs.f90
@@ -5604,22 +5604,9 @@ SUBROUTINE WrVTK_AllMeshes(p_FAST, y_FAST, MeshMapData, ED, BD, AD, IfW, OpFM, H
    
 ! HydroDyn            
    IF ( p_FAST%CompHydro == Module_HD .and. allocated(HD%Input)) THEN     
-      !TODO: Fix for Visualizaton GJH 4/23/20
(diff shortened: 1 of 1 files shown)


modules/openfast-library/src/FAST_Subs.f90
```diff
@@ -5723,12 +5708,12 @@ SUBROUTINE WrVTK_BasicMeshes(p_FAST, y_FAST, MeshMapData, ED, BD, AD, IfW, OpFM,
 
 
 ! Blades
-   IF ( p_FAST%CompAero == Module_AD ) THEN  ! These meshes may have airfoil data associated with nodes...
+   IF ( p_FAST%CompAero == Module_AD .and. ALLOCATED(AD%Input) ) THEN  ! These meshes may have airfoil data associated with nodes...
```
andrew-platt, line 5711: Could this potentially fail with AD14 or AeroDisk in use?  If I remember correctly, `AD%Input` will be allocated even if AD15 isn't used, but things inside it wouldn't be.

andrew-platt: Never mind.  This is not an issue (reread logic above).

andrew-platt (approved):

modules/openfast-library/src/FAST_Subs.f90
bjonkman: The biggest concern is that if the code is aborting due to error and is trying to output the `DebugError` versions of the VTK files, we need to make sure things are allocated before using them. We should probably also check that `AD%Input(1)%rotors` and `AD%y%rotors` are allocated. I'll add that to another PR.

PR #1355, opened 2022-11-29
OLAF: Adding free near wake panels

ebranlard:

This pull request includes:
- New input file option for OLAF `nNWPanelsFree` for number of "free near wake panel". This allows part of the near-wake to be frozen, which can greatly speed up the computation and "stabilize" the wake as the "rollup" is stopped after a while. The shorter the length of the free wake the least accurate the results are likely to be. This unfortunately adds another tuning parameter for the user. We recommend at least 1D for the free near wake. There is also some tuning regarding the convection velocity of the frozen wake (which is determined based on the convection velocity of the free wake before the frozen region). OLAF has some internal parameters to tune this convection velocity, for now, they are not available to the standard user. **It is possible that we will change this convection in the future**. Therefore, a warning message is provided when the user uses this feature. 
- Give the possibility (for advanced user) to use the segment tree for the lifting line-induced velocity calculation. This also greatly speeds up the simulation. Unfortunately, the segment tree (and particle tree) introduces some "noise" in the signals, most likely due to numerics and how things are summed (or a bug in the tree code, hard to debug). This is done by currently "doubling" the size of `VelocityMethod` and `TreeBanchFactor` in the code (first value for the wake, second value for the lifting line). For now, this is not exposed to the user.   
- Some of the default options of OLAF have been slightly changed (documented in API changes)
- Attempt to harmonize the variable names in the input file and documentation. 
- Adding simple timing routines (like the tic/toc routines of Matlab) to easily profile OLAF. 
- Remove unused variables in FVW* routines. (fake print statements are introduced if needed at end of routines). 
- The storage for "particles" (used when the particle tree is used for instance), is now allocated at initialization. (Note despite the allocation that was done at each time step previously, preallocation didn't show any computational time benefit). 

OLAF

A 120s simulation of the NM80 turbine with uniform inflow/constant RPM was performed using the new frozen wake feature and without (fully free). 
The fully-free simulation takes about 0.5hr on my machine, the one with a partially-frozen wake (800=nNWPanel,  180=nNWPanelFree) takes 0.2hr. 

Differences in rotor performance: (differences occurs when the wake starts to be frozen, but at the end of the simulation, the differences are reduced).  
image
Quantities along the blade at simulation end:
image
Comparison of wakes: (note, low vorticity is set to transparent, but the full wake lattice is present throughout the entire wake, no "far-wake" is used here) (the "first quarter" of the wake is free in the partially-frozen case)
image
Vorticity profiles (the convection velocity was tuned so that the location of the tip vortices at the beginning of the frozen region is approximately the same as the ones for the free wake. But as soon as vortex pairing occurs, the similarity stops):
image

Tests are unchanged. Input files are modified to accommodate the new input line. The input files also rely more on "default".


jjonkman (commented):
Sorry for the delay, but these changes all look good to me; thanks!

PR #1357, opened 2022-11-29
CI: exclude bokeh 3.0.[0-3] -- broken plots

andrew-platt:
Bokeh v 3.0.[0-3] had a bug that caused embedded bokeh plots to have a 0x0 size.  This showed up in error plots from ctest.

With bokeh 3.0.2:
Screen Shot 2022-11-29 at 1 51 52 PM

Correctly displays with bokeh 2.4.3:
Screen Shot 2022-11-29 at 3 50 02 PM

Above figures from local testing.

Related to bokeh bug https://github.com/bokeh/bokeh/issues/12614 which is fixed in the bokeh dev with https://github.com/bokeh/bokeh/pull/12653 (v3.0.4 to be released)

This was causing plots from automated testing (using `CTEST_PLOT_ERRORS=On` to not display for channels that did not match. 

This was tested locally.  Will run a sanity check on GH actions with a test commit that will throw away.

### DIFF
--- a/.github/workflows/automated-dev-tests.yml
+++ b/.github/workflows/automated-dev-tests.yml
@@ -46,7 +46,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install numpy "Bokeh>=2.4"
+          pip install numpy "Bokeh>=2.4,!=3.0.0,!=3.0.1,!=3.0.2,!=3.0.3"
       - name: Setup workspace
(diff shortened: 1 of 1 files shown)


andrew-platt:
To check the artifacts were working correctly, I made a temporary turned off the waves in the `hd_5MW_OC4Semi_WSt_WavesWN` HydroDyn module case.  This failed as expected and generated plots as expected.

Screen Shot 2022-11-29 at 11 20 20 PM

After testing, the temporary commit was removed from this PR.

deslaughter (approved):

PR #1361, opened 2022-12-01
TurbSim: User-defined time series updates

bjonkman:

When generating simulated data from a user-defined time-series input file, there was a mismatch between calculating the number of points in the time-series FFT and in the inverse FFT if either of those values were originally entered as not a product of small primes. Time was subtracted from the user input time series, but added to the `AnalysisTime`, possibly causing an error to be triggered. Now both subtract when user-defined time series are used. The code is unchanged when user-defined time series are NOT used.

Also, a possible array overflow issue was corrected when the user entered time series that were not a product of small primes.

Best practice is still to enter the time history and analysis time to give the number of steps as a products of small primes, in which case these modifications won't do anything.

None

TurbSim with user-defined time series (TurbModel = "TIMESR")

This should not change any test results.


jjonkman (approved):
I discussed this with Bonnie and approve of these changes.  Thanks for sharing!

PR #1369, opened 2022-12-06
UA: adding UA_Driver outputs, fix separation function for UAMod=6, and adding r-tests

ebranlard:

The following are included with this pull request:
- Bug Fix: When UAMod_6 (Oye), the separation functions needs to be computed using Cl and not Cn (see the variable `UA_f_cn`)
- Adding an "Output" section to the Driver to toggle outputs of the SummaryFile and of the Polar Coefficients
- Update of the way the polars are written back (adding more columns, and changing the column names depending on the UAMod and if the user provided Cm or not). (see function `WriteAFITables`)
- Additional error handling in unsteady aerodynamics based on the value of Clalpha and the separation function.  
- Regression tests for UA were added. I suspect that we will need to work on those in the future if they end up being hard to pass. In these tests, I use multiple driver files per folders, so I've done of few tweaks in the python script. In particular, I'm using the `inputfile` to come up with the logfile name. 

AeroDyn/UnsteadyAero

**Future work**
Based on the polar written back I can see that we likely have to work more on the separation function when computed based on Cl.

**Questions for reviewers**
And for @bjonkman :)
- I'd be tempted to write the UA summary file in YAML (maybe in a future pull request). I find it a bit hard to read currently. The current form is convenient if the polar contains many tables tough. But otherwise, I find it a bit hard to retrieve information from it. 
- I'd be tempted to "flush out" the error message (with a `WrScr`) after the validation of the polar. Any error message above "None" would be written to disk for each polar, this way we avoid filling up the stack of `ErrMsg`. (I've had occurences where the simulation aborts but the error message is not visible because it is taken up by the warnings and non fatal errors). 
- Do we want to keep the output of the polar in this "OpenFAST format"? I think it would be better to write it as a CSV file. 
- I've chosen to do a build & test for the UA driver test. It requires a special compilation, and there is little benefit to creating a cache, and then reloading it. But I'm open to change this.


bjonkman:
1. I think the UA summary file would be fine in YAML format, but I would still need to be able to read the information in a table-like form (we generate a separate airfoil file for each AD analysis node, and if there are multiple tables in each airfoil file, the number of lines in the current UA summary table can be quite large). 
2. That seems fine to me.
3. I was trying to use the same plotting scripts for the airfoils that I use with the OpenFAST output, but as long as you have the same number of header lines, csv works, too. Is the `cd` column always present? I thought that was optional like `cm`, too. It might be easier to use `WrFileNR()` to get all of the combinations of which columns are present (or always write all the coefficient columns and use `WrFileNR` on the headers names to determine which column represents which coefficient... just depends if you want to list a column of zeros when a coefficient isn't used).

Issue #1325, opened 2022-11-15
Update the directory of generated DLL files


liruiqi97:
**Bug description**
I installed OpenFAST on my MacBook with `cmake` commands. The installation was successful, but some issues happened with the Regression Test. 
When running the [regression tests](https://github.com/OpenFAST/r-test.git), almost all those `5MW` tests were failed on my computer. The reason of the failure was that the project could not find `DISCON.dll`, `DISCON_ITIBarge.dll`, `DISCON_OC3Hywind.dll` in correct folders. Those DLL files were generated somewhere in the `openfast/build` folder. After I moved those DLL files properly, those tests became passed.

The original location of those files were
`openfast/build/reg_tests/glue-codes/openfast/5MW_Baseline/ServoData/DISCON/build/DISCON.dll`
`openfast/build/reg_tests/glue-codes/openfast/5MW_Baseline/ServoData/DISCON_ITI/build/DISCON_ITIBarge.dll`
`openfast/build/reg_tests/glue-codes/openfast/5MW_Baseline/ServoData/DISCON_OC3/build/DISCON_OC3Hywind.dll`
`openfast/build/reg_tests/glue-codes/python/5MW_Baseline/ServoData/DISCON/build/DISCON.dll`
`openfast/build/reg_tests/glue-codes/python/5MW_Baseline/ServoData/DISCON_ITI/build/DISCON_ITIBarge.dll`
`openfast/build/reg_tests/glue-codes/python/5MW_Baseline/ServoData/DISCON_OC3/build/DISCON_OC3Hywind.dll`.
I moved those DLL files respectively into the following locations, and then the problem was solved.
`openfast/build/reg_tests/glue-codes/openfast/5MW_Baseline/ServoData/DISCON.dll`
`openfast/build/reg_tests/glue-codes/openfast/5MW_Baseline/ServoData/DISCON_ITIBarge.dll`
`openfast/build/reg_tests/glue-codes/openfast/5MW_Baseline/ServoData/DISCON_OC3Hywind.dll`
`openfast/build/reg_tests/glue-codes/python/5MW_Baseline/ServoData/DISCON.dll`
`openfast/build/reg_tests/glue-codes/python/5MW_Baseline/ServoData/DISCON_ITIBarge.dll`
`openfast/build/reg_tests/glue-codes/python/5MW_Baseline/ServoData/DISCON_OC3Hywind.dll`.

[...]

PR #1376, opened 2022-12-13
Add `regression_tests` to the ALL target.

andrew-platt:
Fixes issue #1325

When `BUILD_TESTING` is enabled, a user would generally expect that a `make install` command would include everything for the regression testing.  This PR adds the `regression_tests` target to the `ALL` target in the CMake build system.

#1325

This only affects the cmake build system when testing is enabled.

### DIFF
--- a/reg_tests/CMakeLists.txt
+++ b/reg_tests/CMakeLists.txt
@@ -188,7 +188,7 @@ add_custom_target(
 )
 
 add_custom_target(
-  regression_test_module_drivers
+  regression_test_module_drivers ALL
   DEPENDS
     aerodyn_driver
     beamdyn_driver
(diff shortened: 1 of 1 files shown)


reg_tests/CMakeLists.txt
```diff
@@ -181,7 +181,7 @@ add_custom_target(
 )
 
 add_custom_target(
-  regression_tests
+  regression_tests ALL
```
rafmudaf, line 184: It might be good to also add `ALL` to the `regression_test_module_drivers` target below to be consistent.

andrew-platt: done!

deslaughter (approved):
Adding ALL to regression_test_module_drivers isn't necessary as it's a dependency of regression_tests, but it shouldn't hurt either. I think this is ready to be merged.

PR #1418, opened 2023-01-27
SD: SumPrint line in input file had name of SSSum

andrew-platt:
If there was a file parsing error in JOINT ADDITIONAL CONCENTRATED MASSES section, the `SumPrint` line in the input file could have a read error.  This results in an error printed to the screen about `SSSum` not found -- but the input files all say `SumPrint`.

Another option instead of changing this is to update all the SD input files to have `SSSum`.  However, this would be nonstandard with the rest of our input files

This is ready for merging.

Issue reported by @shashankNREL

### DIFF
--- a/docs/source/user/subdyn/input_files.rst
+++ b/docs/source/user/subdyn/input_files.rst
@@ -602,7 +602,7 @@ Output: Summary and Outfile
 In this section of the input file, the user sets flags and switches for
 the desired output behavior.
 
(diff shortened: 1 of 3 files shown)


ebranlard (commented):
Thanks, that makes sense. 
I think the documentation needs also the have instances of SDSum (I've seen one) and SSSum (Not sure if present) replaced by SumPrint

andrew-platt:
Done.  Updated docs and input files in the regression tests.

PR #1427, opened 2023-02-01
Improve cmake's FindMKL to avoid runtime issues and preloading libraries

gbarter:
This was a fix that we identified in WEIS projects running on Eagle a [long time ago](https://github.com/WISDEM/WEIS/commit/c8b262e44504efc9d7f05309aa7b0ff187d01359) but of course forgot to submit a PR until we overwrote the fix with more recent OpenFAST versions.

When running on Eagle linux machines with the Intel compiler suite, we encounter an error that is well captured in various forums ([here](https://community.intel.com/t5/Intel-oneAPI-Math-Kernel-Library/mkl-fails-to-load/m-p/1155538), [here](https://github.com/ikinsella/trefide/issues/2), [here](https://bugs.launchpad.net/ubuntu/+source/intel-mkl/+bug/1947626)).

The suggested fix is to link to the `mkl_rt` library first or to preload the libraries in an environment variable using hard paths.  The first approach is much more appealing and I can confirm that this has resolved our issues on Eagle using WEIS.

Not that I know

Regression tests run with intel compilers

Let me know if I can augment the description.

I think this will be easier to do using the Github Actions workflow via this PR.


andrew-platt (approved):

andrew-platt:
@rafmudaf this worked ok with a CMake + OneAPI + VS just now.  I started to wonder as well, so I tested it.

`cmake .. -G "Visual Studio 16 2019"`

rafmudaf:
Thanks for checking @andrew-platt

PR #1329, opened 2022-11-16
FAST.Farm: Preliminary implementation of wake-added turbulence (WAT) …

ebranlard:
This pull request is ready to be merged (after release of v3.4.0)

**Feature or improvement description**
Sets the stage for the future implementation of Wake Added Turbulence in FAST.Farm, by introducing the calculation of `k_mt` in FAST.Farm WakeDynamics. 
AWAE and glue-code changes will come in a separate pull request. 

**Impacted areas of the software**
FAST.Farm, WakeDynamics

No API changes or test results changes are introduced in this pull request.

PR #1432, opened 2023-02-01
[BugFix] incorrect routine call in WakeDynamics.f90, FAST.Farm not compiling

andrew-platt:
This PR is ready for merging

PR #1329 broke the compilation of FAST.Farm due to an incorrect subroutine call in `WakeDynamics.f90`.  This fixes it.  Wanted to do this as a PR to force testing.

None yet, but I'm sure someone will complain soon.

FAST.Farm now compiles.


ebranlard (approved):
Yes, that right! VTK.f90 is now a very private module and relies on interfaces. 
Thanks for doing that

PR #1434, opened 2023-02-02
Specify StC Force from External DLL

dzalkind:
Add StC_DOF_Mode to specify force acting on StC from an external DLL. 

@andrew-platt, please let me know if I'm missing something normally part of the OpenFAST PR process.  

`StrucCtrl`, uses `BladedInterface_EX`


docs/source/user/servodyn-stc/StC_input.rst
```diff
@@ -425,7 +426,8 @@ when* **StC_DOF_MODE==4**.
 
 **PrescribedForcesCoord** [switch]
 
-   Prescribed forces are in global or local coordinates   {1: global; 2: local}
+   Prescribed forces are in global or local coordinates  {1: global; 2: local}.  
+   When using StC_DOF_MODE==5, PrescribedForcesCoord must be 1.
```
andrew-platt, line 430: I don't think we have to strictly enforce this, but I'm not opposed to it.

andrew-platt:
The only other thing would be to update the StC input files in the regression tests.  This might be easiest to clone `r-test` and do a PR from there to `r-test`.  Then commit the pointer to that r-test commit to this branch.

dzalkind:
I updated the r-tests here: https://github.com/OpenFAST/r-test/pull/88 

I think the pointer is to my fork/branch and needs to be fixed.

dzalkind:
I tested this by applying a -4e5 N force in the z-direction on the upwind column of the IEA-15MW RWT + VolturnUS platform at 50 seconds in the red plot.

image

andrew-platt:
I migrated the r-test branch over.  So it should be good now.

andrew-platt (approved):
Looks good!

PR #524, opened 2020-08-12
Workaround DOUBLE_PRECISION macro when building with HPE MPT

sayerhs:
OpenFAST currently uses `DOUBLE_PRECISION` compile-time definition to activate double precision as default during build. This conflicts with the define in the MPI header file (for HPE MPT) when building the C++ API interface. This commit works around this issue by using `OPENFAST_DOUBLE_PRECISION` variable when building with CMake. The code checks for the original `DOUBLE_PRECISION` macro for backward compatibility. The VS build pathway is unaffected.

Fixes #522

THIS PULL REQUEST IS READY TO MERGE

**Impacted areas of the software**
- `NWTC_Library`

PR #1453, opened 2023-02-14
Remove QuKi (quad-precision) from OpenFAST.

deslaughter:

The changes in this pull request remove the quad-precision (QuKi) from OpenFAST. When the `DOUBLE_PRECISION` CMake option was set to `ON`, it caused `ReKi` reals to be promoted to `real64` precision (double) and `DbKi` reals to be promotoed to `real128` (quadruple precision). Some architectures and compilers do not support quadruple precision (Flang) which caused issues with compiling. In addition, using quad-precision numbers is significantly slower because operations are typically implemented in software instead of in hardware. This change had negligible effects on the linearization regression test outputs (minor fluctuations in numbers near zero).

NWTC Library had the most modifications where all uses of QuKi and associated routines were removed.
- SingPrec.f90
- YAML.f90
- JSON.f90
- NWTC_Num.f90
- NWTC_IO.f90
- NWTC_RandomNumber.f90
- NWTC_SLATEC.f90

HydroDyn Driver: changed dcm from R8Ki to ReKi to match main HydroDyn code and resolve issue with generic interface.

UnsteadyAero: discovered that the reference output in the unit test had uninitialized data in the first time step because it wrote several members of the KC (UA_KelvinChainType) before they had been set. Modified the registry file to provide initial values for all members of the structure (0.0).

Updated unit test reference results for the UnsteadyAero_Driver. Branch created in r-test which will need to be merged into dev. All other tests pass.


andrew-platt (approved):

modules/nwtc-library/src/SingPrec.f90
```diff
@@ -60,18 +57,18 @@ MODULE Precision
 INTEGER, PARAMETER              :: IntKi          = B4Ki                        !< Default kind for integers
 INTEGER, PARAMETER              :: BYTES_IN_INT   = 4                           !< Number of bytes per IntKi number    - use SIZEOF()
 
-#if !defined (DOUBLE_PRECISION) && !defined (OPENFAST_DOUBLE_PRECISION)
```
bjonkman, line 63: Removing this `DOUBLE_PRECISION` preprocessor definition check made the vs-build project files not able to build in double precision any more. (Notice that the new code checks only for `OPENFAST_DOUBLE_PRECISION`). I am adding a fix for this in https://github.com/OpenFAST/openfast/pull/1008, but maybe we need to update documentation, too? See the original reason for the two preprocessor definitions here: https://github.com/OpenFAST/openfast/pull/524

andrew-platt: @deslaughter, can you take a look at this?  I think we need to fix it for 3.5.0

deslaughter: @bjonkman, could I change the vs-build project files to define `OPENFAST_DOUBLE_PRECISION` instead of `DOUBLE_PRECISION`?

bjonkman: @deslaughter You can cherrypick this commit with the changes to the vs-build files: https://github.com/OpenFAST/openfast/commit/9787e501d0fcf824b21f0b7c30d26057c4319804 Hopefully I got them all.

deslaughter: @bjonkman, thanks for creating the commit. I've cherry-picked it into a bugfix branch and created PR #1560. I looked through the VS project files and updated `DOUBLE_PRECISION` in `AeroDyn_Inflow_c_binding.vfproj`.

PR #1474, opened 2023-03-01
BeamDyn driver visualization, and bugfix

andrew-platt:
This PR is ready to merge.

- Added VTK outputs from the driver to simplify checking of BD driver results.
- Fixed bug in driver where the `TranslationDisp` was not set during initialization.  This caused the blade root to "jump" at the first timestep if the `GlbRotBladeT0` is set to `F` with a non-identity DCM and a non-zero`GlbPos` vector.  This required really tiny timesteps and led to huge transients.

None.

BeamDyn driver only.  This does not affect anything in the coupling to OF.

Visualization of blade set to 90 degrees azimuth.  Before the fix of the `TranslationDisp` bug, initialization started with the root at the reference location:
bd_5MW_dynamic_gravity_Az90--T0

After first timestep; note the huge shape change as a result of the sudden jump of the root from (0,0,10) to (0,-10,0):
bd_5MW_dynamic_gravity_Az90--T1

The GlbRotBladeT0 flag in the  bd_5MW_dynamic_gravity_Az90 test case is now set to `False` to test the `TranslationDisp` bug.  The test result is identical to before (as is expected).

### DIFF
--- a/modules/beamdyn/src/Driver_Beam.f90
+++ b/modules/beamdyn/src/Driver_Beam.f90
@@ -33,6 +33,7 @@ PROGRAM BeamDyn_Driver_Program
    REAL(DbKi)                       :: t_global         ! global-loop time marker
    INTEGER(IntKi)                   :: n_t_final        ! total number of time steps
    INTEGER(IntKi)                   :: n_t_global       ! global-loop time counter
+   INTEGER(IntKi)                   :: n_t_vtk          ! global vtk step counter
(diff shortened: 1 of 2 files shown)


andrew-platt:
Fun visualization from BD driver of the bug with the TransDisp.
https://user-images.githubusercontent.com/2745453/222256764-7281f6da-6e4b-40ee-b71e-6cce42a6da18.mov

deslaughter (approved):

michaelasprague (approved):

PR #1480, opened 2023-03-07
hd: fix NBody>1 & NBodyMod=1 WAMIT2 bug

andrew-platt:
This is ready for merging.

An outside user reported issues with a multibody potential flow HydroDyn model with a .12d input file.  We traced this to some arrays in WAMIT2 were not initialized to the correct dimensions for multiple WAMIT bodies in the second order files.

None

HydroDyn 2nd order potential flow only (WAMIT2)

This bug was introduced in v2.5.0 when flexible multibody hydrodynamics was introduced.

We have no publicly sharable models available that demonstrate this issue.


andrew-platt:
@luwang00, could you review this?

jjonkman (approved):
I can't comment on the specific code changes, but they look correct at a high level.  I'm glad to also hear that this change solves Alvaro's issue.

luwang00:
@andrew-platt 
I think I'm a bit late, but the changes make sense. Thanks for fixing this, Andy!

PR #1482, opened 2023-03-07
Bug/bd conv

deslaughter:

This pull request resolves an issue where the BeamDyn driver results were different from OpenFAST results. Specifically, OpenFAST failed to converge during initial correction steps whereas the driver succeeded because it didn't perform corrections. It was possible to make OpenFAST converge by specifying a smaller time step which is why the issue wasn't apparent. During the initial correction steps, the `MiscVars` variable `vvv` wasn't being reset between steps which caused velocity errors to accumulate and the solution to diverge. This was corrected by re-initializing `m%vvv` in `BD_InitAcc` which is only called during these initial correction steps.

- BeamDyn
- Regression Tests: `5MW_Land_BD_DLL_WTurb`

The `5MW_Land_BD_DLL_WTurb` regression test results were updated because they were slightly different at the beginning of the simulation. A new test case, `5MW_Land_BD_Init` was added to demonstrate that the model converges at a larger time step.


andrew-platt (approved):
Thanks for adding the initialization test case!  Good to see we have a large timestep there.

PR #1493, opened 2023-03-16
Updates to Allow Non-Uniform Force Point Distribution on Blades

mchurchf:
I am not sure if this is ready to be merged.  The new feature is fully implemented, so it is ready in that way, but I don't know if it breaks tests.  This should not affect tests unless there is a C++ interface test that now needs new input.  I am not sure if I should call this "draft" or not.  Andy/Raf, please get with me and we can determine.

Worked with Ganesh to add a feature that allows for non-uniform force point distribution.  The particular non-uniform spacing added here adds resolution based on decreasing chord size (i.e., more resolution near tips).  This has been coded so that other options can be easily added later.  This is an important feature because it reduces the actuator line point count requirements by a factor of 5-10 for high-resolution cases.

None

The C++ interface to CFD codes (i.e., the OpenFOAM interface).

N/A

N/A


modules/openfast-library/src/FAST_Library.f90
```diff
@@ -318,6 +318,10 @@ subroutine FAST_Update(iTurb, NumInputs_c, NumOutputs_c, InputAry, OutputAry, En
    OutputAry(1)              = Turbine(iTurb)%m_FAST%t_global 
    OutputAry(2:NumOutputs_c) = Outputs 
 
+   CALL FillOutputAry_T(Turbine(iTurb), Outputs)
```
rafmudaf, line 321: This is duplicated from the lines above. Does this need to be here or is it a mistake?

modules/openfoam/src/OpenFOAM.f90
```diff
@@ -1004,42 +1005,137 @@ END SUBROUTINE CalcForceActuatorPositionsTower
...
+   REAL(ReKi)                              :: dRforceNodes ! Uniform distance between two consecutive force nodes
+   INTEGER(IntKI)                          :: i            ! Loop variables
+   INTEGER(IntKi)                          :: ErrStat2     ! temporary Error status of the operation
+   CHARACTER(ErrMsgLen)                    :: ErrMsg2      ! temporary Error message if ErrStat /= ErrID_None
+   CHARACTER(*),   PARAMETER               :: RoutineName = 'OpFM_CreateActForceBladeTowerNodes'
 
    ErrStat = ErrID_None
    ErrMsg = ""
 
    ! Line2 to Line2 mapping expects the destination mesh to be smaller than the source mesh for deformation mapping and larger than the source mesh for load mapping. This forces me to create nodes at the very ends of the blade.
```
rafmudaf, line 1034: Is this comment still relevant?

modules/openfoam/src/OpenFOAM.f90
```diff
@@ -1004,42 +1005,137 @@ END SUBROUTINE CalcForceActuatorPositionsTower
...
+
+       ! Get the chord at the initial force points.
+       call OpFM_InterpolateForceNodesChord(initOut_AD, p_OpFM, u_OpFM, ErrStat2, ErrMsg2)
+       cNonUniform(1:p_OpFM%nNodesForceBlade) = u_OpFM%forceNodesChord(2:p_OpFM%nNodesForceBlade+1)
+
+       ! Iterate on a chord-based non-uniform spacing.
+       counter = 0
+       e(:) = 1.0e+6
+       eTol = 1.0e-6
+       eSum = sqrt(sum(e(:) * e(:)))
```
rafmudaf, line 1099: I would remove the `(:)` when intending to use a whole array. See https://stevelionel.com/drfortran/2008/03/31/doctor-it-hurts-when-i-do-this/. That's from 2008 so it may or may not be relevant today, but I'd say this should be profiled or just remove it to be safe.

modules/openfoam/src/OpenFOAM.f90
```diff
@@ -1004,42 +1005,137 @@ END SUBROUTINE CalcForceActuatorPositionsTower
...
+          ! convergence.
+          e(:) = cByS(2:p_OpFM%nNodesForceBlade) - cByS(1:p_OpFM%nNodesForceBlade-1)
+          eSum = sqrt(sum(e(:) * e(:)))
+
+          ! increment the iteration counter
+          counter = counter + 1
+
+       end do
+
+       print*, "Using chord-scaled blade force node clustering."
```
rafmudaf, line 1131: @andrew-platt is this appropriate or required to use `ErrMsg` and `ErrStat`?

rafmudaf (changes requested):

modules/openfoam/src/OpenFOAM.f90
mchurchf: Can you point me to an example of how to do that.  I just want some indication to the user of what method it used, and if it used chord-clustered, some statistics about the final result.

modules/openfoam/src/OpenFOAM.f90
mchurchf: I think so, but I don't really know.

modules/openfast-library/src/FAST_Library.f90
mchurchf: I am not sure.  I'll need to ask Ganesh--he did this part.

modules/openfoam/src/OpenFOAM.f90
andrew-platt: I think changing this to `WrScr("Using chord-scaled blade force node clustering"` is sufficient.

Alternatively, you could set `ErrStat = ErrID_Info` and set `ErrMsg` to the text, preferably with the `SetErrStat` subroutine.

We occasionally run into issues with `print` when coupling to other codes (I have vague memories of weirdness with SimuLink).

modules/openfoam/src/OpenFOAM.f90
andrew-platt: We can probably drop that comment.  

If I remember correctly, we are no using a Line2 mesh and a Point mesh.  There is some kind of tolerance there on length, but I don't think it is as stringent.

modules/openfast-library/src/FAST_Library.f90
rafmudaf: @gantech see these comments

mchurchf: I removed the repeated lines and reran a case and got identical results, so I have removed the duplicates.

modules/openfoam/src/OpenFOAM.f90
mchurchf: Okay, I removed the comment.

modules/openfoam/src/OpenFOAM.f90
mchurchf: How about this other print statement

WrScr("   -converged to ", eSum, " in ", counter," iterations.")

That won't compile.  How can I do something like that that mixes strings and numbers?

andrew-platt: The `WrScr` can only accept a single string.  So try the following:

`WrScr(" -converged to "//trim(Num2LStr(eSum))//" in "//trim(Num2LStr(counter))//" iterations.")`

mchurchf: That works!  Thanks, Andy.

modules/openfoam/src/OpenFOAM.f90
mchurchf: Good call.  I removed the (:) from line 1097 and 1099.  The behavior is identical to including the (:)

modules/openfoam/src/OpenFOAM.f90
```diff
@@ -1004,42 +1005,133 @@ END SUBROUTINE CalcForceActuatorPositionsTower
...
+          end do
+          pNonUniform(p_OpFM%nNodesForceBlade) = p_OpFM%BladeLength
+          p_OpFM%forceBldRnodes(:) = pNonUniform(:)
+
+          ! interpolate chord to the new points to get the updated chord values
+          call OpFM_InterpolateForceNodesChord(initOut_AD, p_OpFM, u_OpFM,ErrStat2, ErrMsg2)
+          cNonUniform(1:p_OpFM%nNodesForceBlade) = u_OpFM%forceNodesChord(2:p_OpFM%nNodesForceBlade+1)
+
+          ! compute a = c/ds
+          cByS(:) = cNonUniform(:)/sNonUniform(:)
```
rafmudaf, line 1118: @mchurchf here, too, on the `(:)`. There are quite a few places in the new code that has this syntax and should be removed

mchurchf: @rafmudaf Okay, I fixed those and tested.  All works identically when the (:) are removed.

andrew-platt (approved):

rafmudaf (approved):

PR #1504, opened 2023-03-23
Allow up to 99 output planes in FAST.Farm

andrew-platt:
This PR is ready for merging

Previously we only allowed 9 output planes from the low resolution domain in FAST.Farm.  Users typically want more planes than this, so they've had to rerun the simulation with a different output plane set to get more planes.  This was both time consuming and annoying.

This PR bumps that limit up to 99 planes.  If more are required, it would be relatively trivial to bump this limit up to 999 or more.

None.  Just user complaints directly to me.

FAST.Farm outputs only

See plot that @rthedin will add below.

No test results are changed.  @rthedin is testing with a large turbine array and will report below if it works as expected.

### DIFF
--- a/docs/source/user/fast.farm/InputFiles.rst
+++ b/docs/source/user/fast.farm/InputFiles.rst
@@ -649,15 +649,14 @@ excluding its file extension, where *<n*\ :sub:`t`\ *>* and
 *<n*\ :sub:`low`\ *>* are as specified in
 :numref:`FF:Input:VTK`, but include leading zeros.
 
-For visualization, FAST.Farm can also output low-resolution disturbed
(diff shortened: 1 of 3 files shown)


andrew-platt:
@pdoubraw, this PR may be of interest to you.

jjonkman (commented):
Looks good to me. Thanks for making this change.

Just one comment: Can you also change the FAST.Farm documentation (change 9 to 99 in the following section: https://openfast.readthedocs.io/en/main/source/user/fast.farm/InputFiles.html#visualize?

Thanks,

jjonkman (approved):
Thanks; approved!

PR #1516, opened 2023-04-06
InflowWind Update / restructure

deslaughter:

This pull request contains the following changes to InflowWind:

- Create `FlowFieldType` which contains and represents all wind types.
- Create `FlowFieldType` supporting functions to:
  - Initialize the structure
  - Read wind files
  - Convert between Grid3D and Uniform flow fields
  - Write converted flow field types
- Add ability to calculate flow field acceleration for Uniform and Grid3D flow fields (needed for MHK)
- Create a subroutine to calculate flow field velocities and accelerations only using the `FlowFieldType`
- Reworked Grid3D data layout and interpolation routines to improve locality and performance
- Add flag in the `InflowWind` input file (`VelCubicInterp`) to optionally calculate accelerations and use cubic spline interpolation for velocities.  
- Add new OutList variables for acceleration components and magnitudes (e.g. Wind#AccX),. These values will be zero if accelerations are not requested and the unit field will be set to `INVALID`.
- Updated `InflowWind_CalcOutput` to use new methods and structures
- Updated `FAST.Farm` to use `FlowFieldType`
- Added ability for OLAF and LidarSim to exceed the wind box with out crashing by adding extrapolation to an average flow velocity/acceleration
- Add functionality to InflowWind_Driver to output accelerations and exceed the wind box

- `InflowWind`
- `FAST.Farm` (renamed a variable)

All unit and regression tests are currently passing. This pull request adds three new regression tests for InflowWind:

- Calculation of flow field accelerations (added to ifw_turbsimff in the `Points.Velocity.dat` file)
- Exceeding the wind box without tower grids (ifw_BoxExceed test)
- Exceeding the wind box with tower grids (ifw_BoxExceedTwr test)


glue-codes/fast-farm/src/FASTWrapper.f90
```diff
@@ -184,7 +184,7 @@ SUBROUTINE FWrap_Init( InitInp, u, p, x, xd, z, OtherState, y, m, Interval, Init
          return
       end if
       
-      call move_alloc(m%Turbine%IfW%m%FDext%V, u%Vdist_High)
+      call move_alloc(m%Turbine%IfW%m%FlowField%Grid4D%Vel, u%Vdist_High)
```
andrew-platt, line 187: This will work for now.  We'll storage from `m%` to parameters when we switch to pointers.

modules/inflowwind/src/InflowWind.txt
```diff
@@ -28,34 +23,11 @@ param    ^                       -                 IntKi             HAWC_WindNu
 param    ^                       -                 IntKi             User_WindNumber      -     6     -     "User defined wind."                                                       -
 param    ^                       -                 IntKi             BladedFF_Shr_WindNumber  - 7     -     "Native Bladed binary full-field file."      -
 param    ^                       -                 IntKi             FDext_WindNumber     -     8     -     "4D wind from external souce (i.e., FAST.Farm)."                                                       -
-param    ^                       -                 IntKi             Highest_WindNumber   -     8     -     "Highest wind number supported."                                           -
+param    ^                       -                 IntKi             Point_WindNumber     -     9     -     "4D wind from external souce (i.e., FAST.Farm)."                                                       -
```
andrew-platt, line 26: Do we need the `Point_WindNumber`?  I think FAST.Farm might be able to use the `FDext_WindNumber`.

modules/inflowwind/src/InflowWind_Driver.f90
```diff
@@ -109,73 +108,16 @@ PROGRAM InflowWind_Driver
...
-   CLSettingsFlags%WrHAWC              =  .FALSE.        ! don't convert to HAWC format
-   CLSettingsFlags%WrBladed            =  .FALSE.        ! don't convert to Bladed format
-   CLSettingsFlags%WrVTK               =  .FALSE.        ! don't convert to VTK format
-   CLSettingsFlags%WrUniform           =  .FALSE.        ! don't convert to UniformWind format
-
-      ! Initialize the driver settings to their default values (same as the CL -- command line -- values)
-   Settings       =  CLSettings
-   SettingsFlags  =  CLSettingsFlags
-
+    CLSettings%ProgInfo = ProgInfo
```
andrew-platt, line 119: I'm not sure I understand where setting the defaults for the `CLSettings` are now (maybe some aren't needed).

glue-codes/fast-farm/src/FASTWrapper.f90
deslaughter: I think we may have to leave `FlowField` in `m` because it can be modified when calling `FlowField_GetVelAcc`. When I added the BoxExceed functionality it needed a flag to indicate if a warning had already been displayed about being outside of the box. If `FlowField` is in `p` then it will have `intent(in)` in most function calls so it can't be modified. I'm open to ideas about how to deal with this since I don't think we should store state in `FlowField`.

modules/inflowwind/src/InflowWind.txt
deslaughter: I didn't update the comment. `Point_WindNumber` is for the wind from `ExtInflow` which is an array of wind components corresponding to positions.

glue-codes/fast-farm/src/FASTWrapper.f90
andrew-platt: I agree we shouldn't store state info in `FlowField`.  Perhaps instead of issuing a warning during the simulation, we simply alert the user during `Init` that some points outside the wind box will be allowed.

modules/inflowwind/src/InflowWind.txt
andrew-platt: The comment was confusing me.  Maybe we change the comment on `Point_WindNumber` to `Wind vectors at a list of point locations (used by ExtInflow)` or similar?

andrew-platt (approved):

andrew-platt:
WindBox_extrap
Extrapolation of wind beyond the boundaries of the wind box.  See https://openfast.readthedocs.io/en/dev/source/user/inflowwind/driver.html#boxexceedallow-flag for details on the implementation.

PR #1526, opened 2023-04-14
SD: removing static improvement method (SIM) from SubDyn elastic output mesh (y3 mesh)

ebranlard:

This feature removes the static improvement method contribution from the Y3 mesh. The contribution is stilll present in the write outputs of SubDyn (displacements and elastic loads).

Additional cleanup of the code was performed as part of this pullrequest. 

More variables are allocated as part of Misc to avoid reallocation at each time step. (e.g. AllOuts and internal load vector)

SubDyn

This change came about while thinking about the future tight coupling algorithm of OpenFAST and could improve the stability. 

Results should be unchanged.


jjonkman (approved):
I have not reviewed the source code changes in great detail (I'm not an expert on the SubDyn source code anyway), but I agree with the need for this change to SubDyn.

Issue #1521, opened 2023-04-10
Intermittent failure in TSinflow_curl regression test


deslaughter:
**Bug description**
<A clear and concise description of the bug.>

The `TSinflow_curl` regression test fails on the `dev` branch when run with a GFortran compiled debug executable. The issue seems to be in `AWAE_Init` where it allocates lidar arrays starting with `AllocAry(m%y_IfW_High%lidar%LidSpeed, size(m%y_IfW_Low%lidar%LidSpeed)`.  Calling `size(m%y_IfW_Low%lidar%LidSpeed)` is invalid because `m%y_IfW_Low%lidar%LidSpeed` has not been allocated for this test. Likely due to changes in `Lidar.f90` lines 200-236.

**To Reproduce**
<Update the following list with your specific information.>
1. Compile debug version of `FAST.Farm` with 'gfortran'
2. Run regression test: `ctest -R TSinflow_curl`
3. Observe that test fails, but no error message is displayed in the log file
4. Run executable in GDB and observe the following error message:
`At line 1081 of file /projects/openfast.worktrees/ifw-mod/modules/awae/src/AWAE.f90
Fortran runtime error: Allocatable argument 'm' is not allocated`

**Expected behavior**
<A clear and concise description of what you expected to happen.>
Expected test to pass with debug executable.

**OpenFAST Version**
<Please provide as much detail as possible including git commit. The best information is a screenshot of the OpenFAST system description that prints when running OpenFAST:>

```
$> git rev-parse HEAD
a209327816cc84846a067691921ebae29582d5f7

 **************************************************************************************************
 FAST.Farm

 Copyright (C) 2023 National Renewable Energy Laboratory
 Copyright (C) 2023 Envision Energy USA LTD

 This program is licensed under Apache License Version 2.0 and comes with ABSOLUTELY NO WARRANTY.
 See the "LICENSE" file distributed with this software for details.
 **************************************************************************************************

 FAST.Farm-HEAD-HASH-NOTFOUND
 Compile Info:
  - Compiler: GCC version 11.3.0
  - Architecture: 64 bit
  - Precision: double
  - OpenMP: No
  - Date: Apr 10 2023
  - Time: 14:22:41
 Execution Info:
  - Date: 04/10/2023
  - Time: 14:36:40+0000
```

**System Information (please complete the following information):**
 - OS: Ubuntu 22.04
 - Compiler: GFortran 11.3
[...]

PR #1464, opened 2023-02-21
Lidar within InflowWind

Russell9798:
This PR requests adding LIDAR simulation within InflowWind. This change uses the existing, unused Lidar subroutines as a base to implement lidar measurement, which can take the form of single point beam(s), continuous beam or pulsed beam. Measurement frequency can be specified and the hub's motion on the lidar focal distance can be considered. 

The lidar measurements have been interfaced with ServoDyn to allow interface with ROSCO for Lidar-assisted control.

PR #1533, opened 2023-04-18
[BugFix] AWAE attempting to get size of possibly unallocated array

andrew-platt:
This is ready to merge.

PR #1464 introduced a new Lidar module into InflowWind.  There are some arrays that don't get allocated if it is not used, but the size of these arrays is checked inside the AWAE module.  Taking the `size` of an unallocated array will give an ambiguous result, which can lead to segmentation faults.

Some `if(allocated())` tests have been added to correct the issue.

#1521 
#1464


andrew-platt:
Closes #1521

deslaughter (approved):

PR #1534, opened 2023-04-18
Fix potential error when requesting wind below the ground (also fix build issue with SD syntax)

bjonkman:
This fixes a couple of issues:
- the SubDyn.f90 code did not build on my Windows Intel 2019 compiler due to the way the string constant was specified on a continuation line.
- InflowWind could potentially add shear below the ground or return NaN if trying to compute the power law below the ground. This change sets the mean velocity to 0 below ground.

None reported on OpenFAST code.

Any place that requests inflow below the ground; likely not a common occurrence.

Also note the question in IfW_FlowField.f90's `GetMeanVelocity` function: I am not sure why it is not adding the mean horizontal or vertical shear there.

This should not change any existing r-test cases since they likely don't request wind speeds below z=0.

### DIFF
--- a/modules/subdyn/src/SubDyn.f90
+++ b/modules/subdyn/src/SubDyn.f90
@@ -1166,8 +1166,8 @@ if (ErrStat2/=0) then
    deallocate(StrArray)
    CALL AllocAry(StrArray, nColumns, 'StrArray',ErrStat2,ErrMsg2); if (Failed()) return 
    CALL ReadCAryFromStr ( Line, StrArray, nColumns, 'Members', 'First line of members array', ErrStat2, ErrMsg2 ); if(Failed()) return
-   call LegacyWarning('Member table contains 6 columns instead of 7,  using default member directional cosines ID (-1) for all members.\
-   The directional cosines will be computed based on the member nodes for all members.')
+   call LegacyWarning('Member table contains 6 columns instead of 7,  using default member directional cosines ID (-1) for all members. &
+   &The directional cosines will be computed based on the member nodes for all members.')
    Init%Members(:,7) = -1
 endif
 ! Extract fields from first line
--- a/modules/inflowwind/src/IfW_FlowField.f90
+++ b/modules/inflowwind/src/IfW_FlowField.f90
@@ -333,6 +333,7 @@ contains
          U = 0.0_ReKi
       end select
       
+      !bjj: is there a reason we aren't adding the mean vertical and/or horizontal shear here, as in CalculateMeanVelocity() and Grid3D_AddMeanVelocity()?
    end function
 
 end subroutine
--- a/modules/inflowwind/src/IfW_FlowField.f90
+++ b/modules/inflowwind/src/IfW_FlowField.f90
@@ -310,6 +310,12 @@ contains
       type(Grid3DFieldType), intent(in)  :: GF
       real(ReKi), intent(in)           :: PosZ
       real(ReKi)                       :: U
+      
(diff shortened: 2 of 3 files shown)


bjonkman:
Seems like we have an intermittent failure with the r-tests again; this time with the rtest-interfaces check. I reran the failing case on [7a35023](https://github.com/bjonkman/openfast/actions/runs/4734366374) and it passed the second time (with no code changes)

deslaughter:
@bjonkman, thanks for taking a look at this. You're right that `GetMeanVelocity` isn't doing everything it needs to do. At one point I was pretty sure that `VLinShr` and `HLinShr` weren't used by Grid3D as they're set to 0 in all except one place. The `Grid3D_AddMeanVelocity` subroutine isn't used and should be removed. The `GetMeanVelocity` function should be removed and replaced with `CalculateMeanVelocity`.

bjonkman:
@deslaughter, I think the horizontal and vertical shear that are added in those routines is mostly for the native Bladed format 3D grid files; that's something Envision added, but I don't think OpenFAST has any tests for, yet.

andrew-platt:
Will the checks for above ground cause issues with MHK turbines?  Or are those currents based on the ground located at the seafloor?

andrew-platt (approved):

modules/subdyn/src/SubDyn.f90
```diff
@@ -1166,8 +1166,8 @@ SUBROUTINE SD_Input(SDInputFile, Init, p, ErrStat,ErrMsg)
    deallocate(StrArray)
    CALL AllocAry(StrArray, nColumns, 'StrArray',ErrStat2,ErrMsg2); if (Failed()) return 
    CALL ReadCAryFromStr ( Line, StrArray, nColumns, 'Members', 'First line of members array', ErrStat2, ErrMsg2 ); if(Failed()) return
-   call LegacyWarning('Member table contains 6 columns instead of 7,  using default member directional cosines ID (-1) for all members.\
-   The directional cosines will be computed based on the member nodes for all members.')
+   call LegacyWarning('Member table contains 6 columns instead of 7,  using default member directional cosines ID (-1) for all members. &
+   &The directional cosines will be computed based on the member nodes for all members.')
```
andrew-platt, line 1170: I didn't know you could continue a string to the next line this way.

deslaughter (approved):

modules/subdyn/src/SubDyn.f90
bjonkman: I don't typically use this format for continuing a string, but it's legal: https://www.intel.com/content/www/us/en/docs/fortran-compiler/developer-guide-reference/2023-1/free-source-form.html

andrew-platt: I learned something new today :)

bjonkman:
> Will the checks for above ground cause issues with MHK turbines? Or are those currents based on the ground located at the seafloor?

Basically, this just says if we are asking for mean velocities below the ground or MSL, they should return 0. The power law case returns NaN if z is negative and the power-law coefficient is less than 1 (typical use case), so that wouldn't make sense if z < 0. The log profile also doesn't make sense mathematically if Z <=0. 

I'm not sure how they've set up modeling for MHK turbines, but I don't think this should hurt.

andrew-platt:
Good point.  That would result in nonsensical results.

hkross:
We leave InflowWind alone for MHK and adjust the query points from the glue code before the are passed to InflowWind to put the origin at the seabed (as if it was the ground). So I don't think this will cause an issue.

PR #1538, opened 2023-04-19
Add support for Flang (Classic) compiler

deslaughter:
This pull request is ready to be merged pending regression and unit test results.

This PR adds support for the Classic Flang compiler (https://github.com/flang-compiler/flang) used by AMD ROCm (https://www.amd.com/en/graphics/servers-solutions-rocm). This compiler suite is used on the Crusher and Frontier super computers, possibly others.

This PR mainly affects the build system and NWTC-Library; however, slight changes were required in SubDyn, AeroDyn14, and BeamDyn as the compiler is not fully standards compliant.

SubDyn required removal of `move_alloc` in SD_FEM.f90, though the specific reason is unknown since this subroutine is used elsewhere without problems. For this specific case, the array sizes after being moved were incorrect which caused the subsequent code to fail. Also, when reading the SubDyn input file, carriage returns had to be explicitly removed before calling `ReadIAryFromStrSD`; otherwise, the function failed to parse the integers correctly. 

AeroDyn14 input file parsing was changed because the read function considers reading the end of the file to be an error instead of an end. Building the `dwm_driver_wind_farm` was also disabled as it relies on the `rename` intrinsic which is nonstandard Fortran.

BeamDyn failed when reading the blade input file as the whitespace around the mass and stiffness matrices were considered significant. As a result, code was added to read these lines as comments. This will make the parser fail if these lines do not exist in the input files.

Additional background on the Classic Flang compiler can be found at https://github.com/flang-compiler/flang/wiki. This is not the same Flang compiler as LLVM Flang (https://flang.llvm.org/docs/) though support will likely need to be added for it in the future.

13 out of 90 regression tests fail on this compiler due to a combination of producing different results and runtime failures. The runtime failures are due to an issue in the random number generator included by the compiler and must be resolved by Flang or AMD (https://github.com/flang-compiler/flang/issues/691 has not been included in AMD's fork of Flang).

In addition, the pFunit test framework does not support Flang so unit testing was not available. I'd like to look into replacing pFunit with https://github.com/fortran-lang/test-drive in the future to resolve this issue.


modules/nwtc-library/src/NWTC_Base.f90
```diff
@@ -67,7 +67,7 @@ MODULE NWTC_Base
 
       INTEGER(C_INTPTR_T)       :: FileAddr  = INT(0,C_INTPTR_T)                   !< The address of file FileName.         (RETURN value from LoadLibrary ) [Windows]
       TYPE(C_PTR)               :: FileAddrX = C_NULL_PTR                          !< The address of file FileName.         (RETURN value from dlopen ) [Linux]
-      TYPE(C_FUNPTR)            :: ProcAddr(NWTC_MAX_DLL_PROC)  = C_NULL_FUNPTR    !< The address of procedure ProcName.    (RETURN value from GetProcAddress or dlsym) [initialized to Null for pack/unpack]
+      TYPE(C_FUNPTR)            :: ProcAddr(NWTC_MAX_DLL_PROC)                     !< The address of procedure ProcName.    (RETURN value from GetProcAddress or dlsym) [initialized to Null for pack/unpack]
```
bjonkman, line 70: If this `ProcAddr` isn't initialized, does it cause issues with restart capabilities? I wonder if the uninitialized value becomes a problem when we pack/unpack the array like the comment (from a LONG time ago) mentions.

deslaughter: I agree that that it would be better to have it initialized. I'll see if it will accept initialization to another value. Unfortunately this compiler thinks that `C_NULL_FUNPTR` isn't valid in this context.

modules/subdyn/src/SubDyn.f90
```diff
@@ -1085,6 +1085,11 @@ SUBROUTINE SD_Input(SDInputFile, Init, p, ErrStat,ErrMsg)
 ! Reading reaction lines one by one, allowing for 1, 7 or 8 columns, with col8 being a string for the SSIfile
 do I = 1, p%nNodes_C
    READ(UnIn, FMT='(A)', IOSTAT=ErrStat2) Line; ErrMsg2='Error reading reaction line'; if (Failed()) return
+   j = index(line, achar(13))
```
andrew-platt, line 1088: Can you add a note here about the Flang compiler issues so we don't remove this on accident later?

modules/subdyn/src/SD_FEM.f90
```diff
@@ -1791,8 +1791,11 @@ SUBROUTINE DirectElimination(Init, p, ErrStat, ErrMsg)
    nDOF = p%nDOF_red
    if (p%reduced) then
       ! Temporary backup of M and K of full system
-      call move_alloc(Init%M,  MM)
-      call move_alloc(Init%K,  KK)
+      CALL AllocAry(KK, size(Init%K,1), size(Init%K,2), 'KK',  ErrStat2, ErrMsg2); if(Failed()) return; ! system stiffness matrix
```
andrew-platt, line 1794: Also add a comment here about Flang having trouble with `move_alloc` here.

andrew-platt (approved):

modules/nwtc-library/src/NWTC_Base.f90
ebranlard: Maybe it can be wrapped in a preprocessor directive to be safe? Otherwise, we might need to look at all the time ProcAddr is used and nullify it after the object is created.

I recall adding a safety in the `DLLTypeUnPack` to enforce that these variables are nullified after unpacking if no DLL is used.  https://github.com/OpenFAST/openfast/commit/39d37e6be556ec8e46c721eb23eab67d6d35be6e
But I don't know if such init are needed elsewhere.

ebranlard (approved):

modules/nwtc-library/src/NWTC_Base.f90
```diff
@@ -58,16 +58,17 @@ MODULE NWTC_Base
    
    INTEGER(IntKi), PARAMETER     :: NWTC_MAX_DLL_PROC  = 5                        !< maximum number of procedures that can be dynamically loaded from a DLL (see DLL_Type nwtc_base::dll_type)
    
+   TYPE(C_FUNPTR), PARAMETER     :: NULL_PROC_ADDR(NWTC_MAX_DLL_PROC) = C_NULL_FUNPTR  !< this is a hack so the Flang compiler will initialize ProcAddr to C_NULL_FUNPTR in DLL_Type (remove if no longer needed)
```
deslaughter, line 61: Creating a array of null function pointers seemed to be the only way for Flang to allow initialization of `ProcAddr` in `DLL_Type`. This will provide the same initialization as before, it just required an additional line of code.

modules/subdyn/src/SubDyn.f90
deslaughter: Added comment in latest commit

modules/nwtc-library/src/NWTC_Base.f90
deslaughter: I found a way to fix this in the latest commit by creating an array of null function pointers.

modules/nwtc-library/src/NWTC_Base.f90
andrew-platt: That's elegant!

andrew-platt:
Follow up on this PR.

I am getting a segmentation fault in the compiler (gfortran 10.4) when compiling `NWTC_Base.f90` on my Mac (Ventura 13.3, Intel).  This occurs in release and debug both.

Screenshot 2023-04-20 at 10 12 41 PM

This was working prior to this PR merge:

- 004a1fae4 -- works.  Just prior to merge of #1538 
- e083682dc -- fails.  Commit of merge.

modules/nwtc-library/src/NWTC_Base.f90
andrew-platt: After merging, I discovered this line causes a seg-fault with gfortran 10.4.0

PR #1549, opened 2023-04-27
[Bugfix] InflowWind: using derived types to read wind file headers failed when using ifort.

deslaughter:

**Bug description**
InflowWind did not read the wind file header data correctly when compiled with the Intel Fortran (ifort) compiler.

`InflowWind_IO.f90`

When InflowWind was refactored, the new code made use of derived types to read file header information for TurbSim and Bladed wind files. This allowed multiple variables to be read at one time and simplified the code. However, when compiled with ifort this read did not populate the derived type correctly due to the assumption of padding between the members of the type. GFortran did not have this issue. The solution was to add the `SEQUENCE` statement to the derived types indicating that the members should be considered without padding.

InflowWind tests are passing when compiled with ifort and Visual studio 17 on Windows.


bjonkman:
Thanks for fixing this, @deslaughter!

PR #1550, opened 2023-04-27
TurbSim bug fix: fix index into Z array for text grid files

bjonkman:
When requesting both tower points and text-formatted grid file outputs, an array could exceed the size of the array, causing an access violation. 

This was reported on the NREL Forum: https://forums.nrel.gov/t/difference-between-fast-and-openfast-results/4665/29

TurbSim, only when both `WrADTWR` and `WrFMTFF` are true (and probably only when the number of tower points is more than the number of horizontal grid points).

This doesn't get tested, so will not affect any results. Text-based full-field grid files are rarely used anywhere.


jjonkman (approved):
Thanks for fixing this bug!

PR #1589, opened 2023-05-24
Port openfast-registry to C++ and change pointers

deslaughter:
This pull request is nearly ready to be merged, but it will likely break SeaState so we may need to discuss

This pull request contains a port of the openfast-registry to C++. The new registry doesn't support the `dimspec` command used in AD14 so they were replaced in the input file. This module will be removed soon and this feature is not used by other modules so it didn't seem necessary to keep this feature. AeroDyn_Registry.txt had a malformed line which was acceptable to the previous parser, but didn't work for the new one so the input file was updated. All other registry files are unchanged.

The registry implements specifying a target and a regular pointer pointer on the field name using `&` and `*` respectively. The registry treats TARGETs (`&`) the same as allocatable fields with respect to pack/unpack/copy. Essentially, target's own their memory and it is packed/unpacked/copied like an allocatable field. POINTERS (`*`) do not own their memory and are not packed/unpacked. The pointer association is maintained by the copy function.

```
typedef  ^  ^  FlowFieldType    *FlowField   - - -  "Pointer to field, does not own memory, should only be used to point to targets"
typedef  ^  ^  FlowFieldType    &FlowField  - - -  "Target pointer, owns memory, will be deallocated when Destroy is called"  -
```

Mainly affects `openfast-registry`, but due to some insignificant differences all `_Types.f90` files have been regenerated.

**Merge Ordering**
Merge after #1008, but only when SeaState refactoring is ready for merging.


luwang00:
This pull request might create some conflict with the current implementation of SeaState (PR #1008 and PR #1578). Suggest holding off until PR #1008 and PR #1578 are merged in first. We can perhaps save this for version 4.1.

luwang00:
The commit associated with this pull request is cherry-picked into this [branch](https://github.com/luwang00/openfast/tree/f/SeaStateRefactoring). Currently working on updating SeaState and HydroDyn to be compatible with and to make full use of the new OpenFAST Registry capabilities with pointers to derived types. It should be done in time for v4.0.0 release.

PR #1596, opened 2023-05-31
Use pointers to couple InflowWind and AeroDyn

deslaughter:
This is a draft pull request for review of the implementation.

The goal of this PR is to allow AeroDyn to call InflowWind directly via a pointer to `FlowFieldType` derived data type and the `IfW_FlowField_GetVelAcc` function. This code adds logic to copy the FlowField pointer from the InflowWind Parameters derived type into the AeroDyn Parameters derived type during initialization. From there `AD_CalcOutputs` and `AD_UpdateStates` can call `IfW_FlowField_GetVelAcc` to get and update the wind velocities and acceleration. 

- AeroDyn
- AeroDyn Inflow
- OpenFAST Library (glue code)

The calls to get the wind velocity from InflowWind and transfer them to AeroDyn within `FAST_Solver` have been removed as they are no longer needed. To avoid disrupting Linearization, the wind velocity variables have been left as inputs in AeroDyn but will be moved to MiscVars as part of Linearization work in a future PR.

Test results are unchanged as expected.


modules/aerodyn/src/AeroDyn_Inflow.f90
```diff
@@ -256,36 +258,65 @@ subroutine ADI_CalcOutput(t, u, p, x, xd, z, OtherState, y, m, errStat, errMsg)
...
+
+   ! --- Outputs for driver
    y%PLExp = m%IW%PLExp
 
    ! --- Set outputs
-!TODO: this assumes one rotor!!!
-   y%WriteOutput(1:p%AD%rotors(1)%NumOuts+p%AD%rotors(1)%BldNd_TotNumOuts) = y%AD%rotors(1)%WriteOutput(1:p%AD%rotors(1)%NumOuts+p%AD%rotors(1)%BldNd_TotNumOuts)
-   y%WriteOutput(p%AD%rotors(1)%NumOuts+p%AD%rotors(1)%BldNd_TotNumOuts+1:p%NumOuts) = y%IW_WriteOutput(1:m%IW%p%NumOuts)
+   !TODO: this assumes one rotor!!!
+   y%WriteOutput = [y%AD%rotors(1)%WriteOutput, y%IW_WriteOutput]
```
ebranlard, line 319: Though I like the condensed syntax, I tend to be scared if that uses the stack.

modules/aerodyn/src/AeroDyn_Inflow.f90
```diff
@@ -405,125 +451,6 @@ subroutine concatOutputHeaders(WriteOutputHdr0, WriteOutputUnt0, WriteOutputHdr,
...
-   integer(IntKi)       :: errStat2      ! Status of error message
-   character(errMsgLen) :: errMsg2       ! Error message if errStat /= ErrID_None
-   errStat = ErrID_None
-   errMsg  = ''
-   if (IW%CompInflow==1) then
-      call InflowWind_CalcOutput(t, u_IfW, IW%p, IW%x, IW%xd, IW%z, IW%OtherSt, IW%y, IW%m, errStat2, errMsg2)
-      call SetErrStat(errStat2, errMsg2, errStat, errMsg, 'ADI_CalcOutput_IW') 
-   else
-      !$OMP PARALLEL DEFAULT(SHARED)
-      !$OMP DO PRIVATE(j,z) schedule(runtime)
```
ebranlard, line 487: Here the OMP parallelization is likely lost.

modules/aerodyn/src/AeroDyn_Driver_Subs.f90
```diff
@@ -22,7 +22,6 @@ module AeroDyn_Driver_Subs
    use AeroDyn_Inflow_Types
    use AeroDyn_Inflow, only: ADI_Init, ADI_ReInit, ADI_End, ADI_CalcOutput, ADI_UpdateStates 
    use AeroDyn_Inflow, only: concatOutputHeaders
-   use AeroDyn_Inflow, only: ADI_ADIW_Solve ! TODO remove me
```
ebranlard, line 25: Nice to see that it got removed :)

modules/inflowwind/src/InflowWind_Subs.f90
```diff
@@ -1747,42 +1719,27 @@ SUBROUTINE InflowWind_GetHubValues( Time, InputData, p, x, xd, z, OtherStates, m
    TYPE(InflowWind_OtherStateType),          INTENT(IN   )  :: OtherStates       !< Other states at Time
    TYPE(InflowWind_MiscVarType),             INTENT(INOUT)  :: m                 !< misc/optimization variables
    REAL(ReKi),                               INTENT(  OUT)  :: Velocity(3)       !< at InputPosition
-      
    INTEGER(IntKi),                           INTENT(  OUT)  :: ErrStat           !< Error status of the operation
    CHARACTER(*),                             INTENT(  OUT)  :: ErrMsg            !< Error message if ErrStat /= ErrID_None
+   
+   CHARACTER(*),PARAMETER     :: RoutineName="InflowWind_GetSpatialAverage"
```
ebranlard, line 1725: routinename needs changing here

modules/inflowwind/src/InflowWind_Subs.f90
```diff
@@ -1665,14 +1658,6 @@ END SUBROUTINE CalculateOutput
 !> this routine calculates a rotor-averaged mean velocity, DiskVel 
 SUBROUTINE InflowWind_GetSpatialAverage( Time, InputData, p, x, xd, z, OtherStates, m, MeanVelocity, ErrStat, ErrMsg )
```
ebranlard, line 1659: I would rename that function to GetRotorSpatialAverage. It threw me off, because the inputs and number of points are already hardcoded, so it's not a generic function to compute an average.

modules/inflowwind/src/InflowWind_Subs.f90
```diff
@@ -1681,63 +1666,50 @@ SUBROUTINE InflowWind_GetSpatialAverage( Time, InputData, p, x, xd, z, OtherStat
...
+   !----------------------------------------------------------------------------
 
-   do i=1,IfW_NumPtsAvg
-      MeanVelocity = MeanVelocity + m%y_Avg%VelocityUVW(:,i)
-   end do
-   MeanVelocity = MeanVelocity / REAL(IfW_NumPtsAvg,ReKi)
-        
-   
-   RETURN
+   CALL IfW_FlowField_GetVelAcc(p%FlowField, 0, Time, m%u_Avg%PositionXYZ, &
```
ebranlard, line 1699: This is a very important function, but somehow I didn't see during the review, I probably missed it. Where is it defined?

ebranlard (approved):
This looks great. I didn't review the changes to the registry. 

I didn't find the `GetVelAcc` function but I guess it was part of a previous pull request. I wanted to check that this function has some speedup if a simple flow field is used (shear),  but I think you indeed took care of it. I also wanted to see if we could easily add a flag to request OpenMP parallelization on the fly (likely harder). For even more speedup in OLAF we  would need to pass the flowfield to FVW. That can be done at another stage. 

I have my nerdy preference of having the "if failed" at the end of the line so that similar lines can easily be read on top of each other and I'm not sure I'm set on cutting lines with `&` (when they are medium size). Don't change that for me of course.

modules/aerodyn/src/AeroDyn_Inflow.f90
deslaughter: I agree that the calculation won't be in parallel in OMP. I'd like to explore adding OpenMP within the FlowField, which is inside InflowWind.

modules/aerodyn/src/AeroDyn.f90
```diff
@@ -1639,7 +1662,52 @@ subroutine AD_UpdateStates( t, n, u, utimes, p, x, xd, z, OtherState, m, errStat
...
+   call AD_SetExternalWindPositions(p, u, o, m%WindPos, node, errStat2, errMsg2)
+   if(Failed()) return
+
+   ! Calculate wind velocity/acceleration
+   call IfW_FlowField_GetVelAcc(p%FlowField, 1, t, m%WindPos, &
+                                m%WindVel, m%WindAcc, errStat2, errMsg2)
+   if(Failed()) return
+
+   ! Transfer wind velocity to nodes
+   node = 1
```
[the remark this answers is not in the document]
deslaughter: Yes, I'll add a comment since it was really confusing the first time I looked at it. The code that set's the positions increments the node number fore setting the value, and the code that sets the wind, increments the node position after setting the value. It would be good to make it consistent.

modules/inflowwind/src/InflowWind_Subs.f90
deslaughter: It's defined at the top of `IfW_FlowField.f90` which was added when InflowWind was restructured (#1516). The FlowFieldType is initialized in InflowWind and contains data for all of the wind types. The `GetVelAcc` function can then be used to query the velocity/acceleration at any point regardless of which wind type is being used.

modules/inflowwind/src/InflowWind_Subs.f90
deslaughter: I agree with renaming it. I'll get that in soon.

modules/aerodyn/src/AeroAcoustics_Types.f90
```diff
@@ -73,7 +73,7 @@ MODULE AeroAcoustics_Types
 ! =======================
 ! =========  AA_InputFile  =======
   TYPE, PUBLIC :: AA_InputFile
-    REAL(DbKi)  :: DT_AA      !< Time interval for aerodynamic calculations {or "default"} [s]
+    REAL(DbKi)  :: DT_AA      !< Time interval for aerodynamic calculations {or  [default"}"]
```
bjonkman, line 76: The comments/units here are a little bit off... probably due to the way quotation marks are used in the registry input file.

modules/aerodyn/src/AeroDyn_Types.f90
```diff
@@ -226,7 +227,7 @@ MODULE AeroDyn_Types
     CHARACTER(1024)  :: FVWFileName      !< FVW input filename [quoted string]
     CHARACTER(1024) , DIMENSION(:), ALLOCATABLE  :: AFNames      !< Airfoil file names (NumAF lines) [quoted strings]
     LOGICAL  :: UseBlCm      !< Include aerodynamic pitching moment in calculations? [flag]
-    LOGICAL  :: SumPrint      !< Generate a summary file listing input options and interpolated properties to "<rootname>.AD.sum"? [flag]
+    LOGICAL  :: SumPrint      !< Generate a summary file listing input options and interpolated properties to  [<rootname>.AD.sum"?"]
```
bjonkman, line 230: The comments/units here are a little bit off... probably due to the way quotation marks are used in the registry input file.

modules/aerodyn/src/AeroDyn_Types.f90
```diff
@@ -188,7 +189,7 @@ MODULE AeroDyn_Types
 ! =========  AD_InputFile  =======
   TYPE, PUBLIC :: AD_InputFile
     LOGICAL  :: Echo      !< Echo input file to echo file [-]
-    REAL(DbKi)  :: DTAero      !< Time interval for aerodynamic calculations {or "default"} [s]
+    REAL(DbKi)  :: DTAero      !< Time interval for aerodynamic calculations {or  [default"}"]
```
bjonkman, line 192: The comments/units here are a little bit off... probably due to the way quotation marks are used in the registry input file.

modules/aerodyn/src/AeroDyn_Types.f90
```diff
@@ -358,10 +359,13 @@ MODULE AeroDyn_Types
 ! =======================
 ! =========  AD_MiscVarType  =======
   TYPE, PUBLIC :: AD_MiscVarType
-    TYPE(RotMiscVarType) , DIMENSION(:), ALLOCATABLE  :: rotors      !< MiscVars for each rotor [-]
+    TYPE(RotMiscVarType) , DIMENSION(:), ALLOCATABLE  :: rotors      !< - [MiscVars for each rotor]
```
bjonkman, line 362: Looks like a problem (extra "-") in the AeroDyn registry input file here, too.

bjonkman:
Thanks, @deslaughter! I'm actually surprised the old registry parsed that without the escape character. There are probably other registry input files that have the same issue, but I just happened to see the AA and AD ones.

modules/aerodyn/src/AirfoilInfo_Types.f90
```diff
@@ -68,7 +68,7 @@ MODULE AirfoilInfo_Types
     REAL(ReKi)  :: k2      !< airfoil parameter in the x_cp_hat curve best-fit [ignored if UAMod<>1] [-]
     REAL(ReKi)  :: k3      !< airfoil parameter in the x_cp_hat curve best-fit [ignored if UAMod<>1] [-]
     REAL(ReKi)  :: k1_hat      !< Constant in the expression of Cc due to leading edge vortex effects.  [ignored if UAMod<>1] [-]
-    REAL(ReKi)  :: x_cp_bar      !< Constant in the expression of \hat(x)_cp^v [ignored if UAMod<>1, default = 0.2] [-]
+    REAL(ReKi)  :: x_cp_bar      !< Constant in the expression of hat(x)_cp^v [ignored if UAMod<>1, default = 0.2] [-]
```
andrew-platt, line 71: Since we use `\` as an escape character now, would we need `\\` to keep the `\` in the comment?

I'm not convinced it is worth bothering with here since we really don't use look at the `_Types.f90` files all that often.

modules/inflowwind/src/InflowWind.f90
```diff
@@ -290,8 +284,17 @@ SUBROUTINE InflowWind_Init( InitInp, InputGuess, p, ContStates, DiscStates, Cons
    ! Set flow field input data based on wind type
    !----------------------------------------------------------------------------
 
+   ! If flowfield is allocated, deallocate and allocate again to clear old data
+   if (associated(p%FlowField)) deallocate(p%FlowField)
```
andrew-platt, line 288: `allocated` or `associated` ?

deslaughter: Since `p%FlowField` is now a pointer, `associated` is used to check if the pointer has memory associated with it. `allocated` is used for allocatable arrays.

modules/aerodyn/src/AirfoilInfo_Types.f90
andrew-platt: I don't think we count on doxygen picking up the _Types.f90 files correctly, so I don't think we need to fix it.

modules/inflowwind/src/InflowWind.f90
andrew-platt: That makes sense.  At first glance it looked rather strange.

modules/map/src/MAP_Types.f90
```diff
@@ -1730,144 +1695,104 @@ SUBROUTINE MAP_CopyOtherState( SrcOtherStateData, DstOtherStateData, CtrlCode, E
...
-  IF (PRESENT(DEALLOCATEpointers)) THEN
-     DEALLOCATEpointers_local = DEALLOCATEpointers
-  ELSE
-     DEALLOCATEpointers_local = .true.
-  END IF
-  
 IF (ASSOCIATED(OtherStateData%H)) THEN
- IF (DEALLOCATEpointers_local) &
   DEALLOCATE(OtherStateData%H)
-  OtherStateData%H => NULL()
```
andrew-platt, line 1765: Don't we still need this?

andrew-platt, line 1765: Do we still need this?

bjonkman: I've seen some conflicting guidance on whether the `deallocate` statement also nullifies pointers or leaves the associated status undefined (as in initialization). [This comment](community.intel.com/t5/Intel-Fortran-Compiler/Question-about-deallocation-of-pointers/m-p/754651#M10138) from Intel's "Doctor Fortran" to the question `How to decide if a pointer has been allocated or pointed?` makes me think that we should include the statement that sets the pointer to NULL() afterwards.
image

modules/openfoam/src/OpenFOAM_Types.f90
```diff
@@ -2427,151 +2386,109 @@ SUBROUTINE OpFM_CopyInput( SrcInputData, DstInputData, CtrlCode, ErrStat, ErrMsg
...
-  IF (PRESENT(DEALLOCATEpointers)) THEN
-     DEALLOCATEpointers_local = DEALLOCATEpointers
-  ELSE
-     DEALLOCATEpointers_local = .true.
-  END IF
-  
 IF (ASSOCIATED(InputData%pxVel)) THEN
- IF (DEALLOCATEpointers_local) &
   DEALLOCATE(InputData%pxVel)
-  InputData%pxVel => NULL()
```
andrew-platt, line 2462: Given the comment from Dr. Fortran (comment from @bjonkman above), we may need this as well here.

modules/supercontroller/src/SuperController_Types.f90
```diff
@@ -2282,46 +2207,34 @@ SUBROUTINE SC_CopyInput( SrcInputData, DstInputData, CtrlCode, ErrStat, ErrMsg )
...
-  IF (PRESENT(DEALLOCATEpointers)) THEN
-     DEALLOCATEpointers_local = DEALLOCATEpointers
-  ELSE
-     DEALLOCATEpointers_local = .true.
-  END IF
-  
 IF (ASSOCIATED(InputData%toSCglob)) THEN
- IF (DEALLOCATEpointers_local) &
   DEALLOCATE(InputData%toSCglob)
-  InputData%toSCglob => NULL()
```
andrew-platt, line 2317: May also need this.

modules/openfast-library/src/FAST_Subs.f90
```diff
@@ -569,7 +568,11 @@ SUBROUTINE FAST_InitializeAll( t_initial, p_FAST, y_FAST, m_FAST, ED, BD, SrvD,
       Init%InData_IfW%WtrDpth          = p_FAST%WtrDpth
 
       Init%InData_IfW%NumWindPoints = 0
-      IF ( p_FAST%CompServo == Module_SrvD ) Init%InData_IfW%NumWindPoints = Init%InData_IfW%NumWindPoints + 1
+      
+      IF ( p_FAST%CompServo == Module_SrvD ) THEN
+         Init%InData_IfW%NumWindPoints = Init%InData_IfW%NumWindPoints + 1
```
andrew-platt, line 573: I suppose once we pass the wind pointer into SrvD, we won't need to set the `NumWindPoints` here (may still need for the c-bindings interfaces).

modules/aerodyn/src/AeroDyn_Inflow.f90
```diff
@@ -256,36 +258,68 @@ subroutine ADI_CalcOutput(t, u, p, x, xd, z, OtherState, y, m, errStat, errMsg)
...
    errMsg  = ""
 
-   ! --- CalcOutputs for IW (Sets u_AD%rotors(:)%InflowOnBlade, etc,  and m%IW%y)
-   y%IW_WriteOutput(:) = m%IW%y%WriteOutput(:)
+   !----------------------------------------------------------------------------
+   ! Calculate InflowWind outputs if module was initialized
+   !----------------------------------------------------------------------------
 
+   if (m%IW%CompInflow == 1) then
+      call InflowWind_CalcOutput(t, m%IW%u, m%IW%p, m%IW%x, m%IW%xd, m%IW%z,  &
```
andrew-platt, line 275: I assume we only need InflowWind_CalcOutput here to get the outputs from IfW?

PR #1618, opened 2023-06-09
[BugFix] OpenFAST Registry on Windows

deslaughter:

This PR resolves several issues with building the new OpenFAST Registry (C++) on Windows. The dependency on `std::filesystem` has been removed so it is compatible with older compilers (tested with VS2015). It also fixes a Visual Studio compiler issue with the strings in `templates.hpp` being too long.

Secondly it fixes an issue with the MAP library which occurs when combining C and Fortran sources in a single library and then generating a Visual Studio Solution via CMake. This is accomplished by splitting MAP into C and Fortran libraries. This requires that both libraries are added to the Simulink project.

- OpenFAST Registry (C++
- MAP
- Simulink


ebranlard (approved):
Thanks for doing that. The previous version didn't compile on windows with Visual Studio and with cmake & msys2. I've tried it and it now works.

andrew-platt:
Thanks for fixing this!

Issue #1484, opened 2023-03-08
Bug report


Stephen-cby:
**Bug description**
<A clear and concise description of the bug.>Issue with visualization of IEA-15-240-RWT-UMaineSemi
**To Reproduce**
<Update the following list with your specific information.>Steps to reproduce the behavior:
1. I am using OpenFAST v3.3.0 from https://github.com/OpenFAST/openfast/releases/download/v3.3.0/openfast_x64.exe
2. I am running IEA-15-240-RWT-UMaineSemi case from https://github.com/IEAWindTask37/IEA-15-240-RWT/archive/refs/tags/v1.1.3.zip
3. I want to get the visualization of this case and set WrVTK=2; VTK_type=1 in the input file. 
4. However, the error occurs:
**************************************************************************************************

OpenFAST-Unversioned from 5f3fb6ef74f48e75ca94000090737a41866fb264
Compile Info:
 - Compiler: Intel(R) Fortran Compiler 2021
 - Architecture: 64 bit
 - Precision: single
 - OpenMP: No
 - Date: Mar 08 2023
 - Time: 16:38:05
Execution Info:
 - Date: 03/08/2023
 - Time: 20:29:10+0800

OpenFAST input file heading:
    IEA 15 MW offshore reference model on UMaine VolturnUS-S semi-submersible floating platform

Running ElastoDyn.
Running AeroDyn.
Warning: Turning off Unsteady Aerodynamics because UA parameters are not included in airfoil
(airfoil has likely has constant polars). (node 1, blade 1)
Warning: Turning off Unsteady Aerodynamics because C_nalpha is 0. (node 2, blade 1)
Warning: Turning off Unsteady Aerodynamics because C_nalpha is 0. (node 4, blade 1)
Warning: Turning off Unsteady Aerodynamics because C_nalpha is 0. (node 5, blade 1)
Warning: Turning off Unsteady Aerodynamics because UA parameters are not included in airfoil
(airfoil has likely has constant polars). (node 1, blade 2)
Warning: Turning off Unsteady Aerodynamics because C_nalpha is 0. (node 2, blade 2)
Warning: Turning off Unsteady Aerodynamics because C_nalpha is 0. (node 4, blade 2)
Warning: Turning off Unsteady Aerodynamics because C_nalpha is 0. (node 5, blade 2)
Warning: Turning off Unsteady Aerodynamics because UA parameters are not included in airfoil
(airfoil has likely has constant polars). (node 1, blade 3)
[...]

PR #1621, opened 2023-06-13
[BugFix] HD wave visualization with 2nd order waves

andrew-platt:
This was reported in issue #1484

There has been a bug in HydroDyn for some time that if surface visualization was turned on with second order waves, the simulation would abort.

This was reported in #1484.

This only affects visualization of surfaces when second order waves are turned on.

This bug has been in the code for some time -- at least since v 3.3.0, but perhaps earlier.

No test results are affected (we don't extensively test the visualization on GitHub actions due to the resource expense).


andrew-platt:
Closes #1484

jjonkman (approved):
Thanks for fixing this bug!

PR #1624, opened 2023-06-14
Use pointers to couple InflowWind and FAST.Farm

deslaughter:

This commit is the initial attempt to use pointers to couple the wind velocities generated by AWAE in FAST.Farm to the InflowWind FlowField which is used by AeroDyn to calculate the wind velocities. The Vdist_High data in AWAE has been changed to a pointer and FlowField%Grid4D%Vel is associated with this member to alleviate the need to copy the data between the modules.

- FAST.Farm
- AWAE
- InflowWind
- Openfast Library

Many of these areas were impacted because pointers were added to InitInput structures to make use of standard initialization functions. It would have been possible to skip these changes and directly associate the pointers in FAST.Farm, but I thought doing it this way would be clearer.

FAST.Farm test results are unchanged though the regression tests run slightly faster.


andrew-platt (commented):
Looks good!  I'll run a couple of cases I have for WAT and let you know if I find anything.

PR #1629, opened 2023-06-15
MHK: add parameters for readability

bjonkman:
This PR is for code readability. 
- It replaces the `MHK=={0|1|2}` statements using parameters for the 3 allowed values of MHK. I am not thrilled with putting the MHK parameters in NWTC Library, but they are used in many different modules so that's where I put them (at least for now).
- It also simplifies some of the logic statements. For example `IF (A) THEN ... ELSEIF (NOT A)` statements are now `IF (A) THEN... ELSE`, avoiding a second check that must always be true. 

None.

This impacts the places that check the value of the `MHK` variable: ElastoDyn, AeroDyn, InflowWind, and OpenFAST glue code

This shouldn't affect any tests. Tests have have failed are related to GH Actions getting the wrong patch version of python.


andrew-platt (approved):

andrew-platt:
@hkross, do you want to take a quick look?

andrew-platt:
I think it makes sense to change the AeroDyn_Driver_Subs.f90 as well.

bjonkman:
AD driver has been updated as well.

PR #1630, opened 2023-06-15
Make CMake module libs STATIC

deslaughter:

This PR changes the CMakeLists.txt files in each module so that the module libraries are built as STATIC even if CMake is configured with `-DBUILD_SHARED_LIBS=ON`. Interface libraries (C-Bindings) and `openfastlib` are always built as shared libraries so this change will not affect downstream users. This option is part of CMake: [BUILD_SHARED_LIBS](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html)

CMakeLists.txt files of all modules

This change arose from a problem installing WEIS which built OpenFAST with `-DBUILD_SHARED_LIBS=ON`. At one point, the only way to build `openfastlib` as a shared library was to set `-DBUILD_SHARED_LIBS=ON`. This library is currently always built as shared but many users still set `BUILD_SHARED_LIBS=ON` which causes all module libraries to be created as shared libraries so they are dynamically loaded at run time. This commit changes the CMakeLists.txt files to build all module libs as `STATIC` regardless of how `BUILD_SHARED_LIBS` is set. As a result, all module libraries will be statically linked into `openfast`, `openfastlib` and `FAST.Farm` which should improve performance and reduce complexity. This also prevents a problem with building the Simulink MEX file which wouldn't work if `BUILD_SHARED_LIBS=ON`.

### DIFF
--- a/glue-codes/openfast-cpp/CMakeLists.txt
+++ b/glue-codes/openfast-cpp/CMakeLists.txt
@@ -29,7 +29,10 @@ find_package(ZLIB REQUIRED)
 find_package(HDF5 REQUIRED COMPONENTS C HL)
 find_package(yaml-cpp REQUIRED)
 
-add_library(openfastcpplib src/OpenFAST.cpp src/SC.cpp)
+add_library(openfastcpplib
+  src/OpenFAST.cpp
+  src/SC.cpp
+)
 set_property(TARGET openfastcpplib PROPERTY POSITION_INDEPENDENT_CODE ON)
 target_link_libraries(openfastcpplib
   openfastlib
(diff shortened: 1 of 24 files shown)


andrew-platt (approved):

deslaughter:
@andrew-platt , `BUILD_SHARED_LIBS` is a CMake option, https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html, which it looks like is duplicated in the main CMakeLists.txt file. So it could still be set even if it was removed. In this PR all the libraries except `openfastcpplib` are set to `SHARED` or `STATIC`. I wasn't sure which way that library should be built, though I expect it needs to be `SHARED` for the downstream applications. I'm open to removing it and seeing if it causes any issues. Let me know what you think.

deslaughter:
I've decided to leave this PR as is and revisit removing the `BUILD_SHARED_LIBS` option from the main `CMakeLists.txt` file in the future.

PR #1639, opened 2023-06-20
InflowWind pointers in AeroDyn, OpenFOAM, and Lidar

deslaughter:

A test case needs to be added to demonstrate that Lidar still works.

This PR integrates the InflowWind FlowField pointer into OpenFOAM and InflowWind Lidar. It also modifies the AeroDyn to remove the need to collect all node positions for the wind calculation and then unpack the data into the inputs. The `IfW_GetVelAcc` routine has been modified to accept a flag which allows the bounding box to be exceeded. This feature is enabled for wind velocity calls from OLAF and Lidar. 

- AeroDyn
- InflowWind
- OpenFOAM

Regression test results are unchanged. OpenFOAM changes have not been tested.


modules/aerodyn/src/AeroDyn.f90
```diff
@@ -6998,11 +7053,11 @@ SUBROUTINE Perturb_u( p, n, perturb_sign, u, du )
       u%BladeMotion(3)%TranslationAcc(fieldIndx,node) = u%BladeMotion(3)%TranslationAcc(fieldIndx,node) + du * perturb_sign
 
    CASE (25) !Module/Mesh/Field: u%InflowOnBlade(:,:,1) = 25;
```
andrew-platt, line 7055: Comments for case 25-27 could be updated

modules/inflowwind/src/IfW_FlowField.f90
```diff
@@ -107,13 +110,26 @@ subroutine IfW_FlowField_GetVelAcc(FF, IStart, Time, PositionXYZ, VelocityUVW, A
...
+            Position(:, i) = GetPrimePosition(PositionXYZ(:, i))
+         end do
+      end if
+   else if (present(PosOffset)) then
+      Position = PositionXYZ + spread(PosOffset, 2, NumPoints)
+   else 
       Position = PositionXYZ
    end if
 
+   ! If
```
andrew-platt, line 128: ?

deslaughter: The if statement above it was rearranged a couple of times, must have missed deleting part of it that was commented out

andrew-platt (approved):

ebranlard (approved):

modules/aerodyn/src/AeroDyn.f90
```diff
@@ -1681,29 +1675,88 @@ subroutine AD_CalcWind(t, u, p, o, m, ErrStat, ErrMsg)
...
+      if (p%rotors(iWT)%MHK > 0) then
+         PosOffset = [0.0_ReKi, 0.0_ReKi, p%rotors(iWT)%WtrDpth]
+      else
+         PosOffset = 0.0_ReKi
+      end if
+
+      ! Hub
+      if (u%rotors(iWT)%HubMotion%Committed) then
+         call IfW_FlowField_GetVelAcc(p%FlowField, StartNode, t, &
+            real(u%rotors(iWT)%HubMotion%TranslationDisp + u%rotors(iWT)%HubMotion%Position, ReKi), &
```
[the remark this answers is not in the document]
deslaughter: I like the idea of providing an interface for IfW_FlowField_GetVelAcc, it could solve some issues that came up during the integration which made it seem hard to use. I'll play around with it a little and do another commit.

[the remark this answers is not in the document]
deslaughter: I took a look at providing multiple interfaces to IfW_GetVelAcc, but it requires declaring whether Position, Velocity, and Accel are always R4Ki or R8Ki. This would require many changes throughout the code. I'll keep thinking about a better solution as I continue to integrate the Flow Field.

PR #1643, opened 2023-06-21
IfW: add VTK output of slice in XY to driver

andrew-platt:
This PR is ready to merge.

This allows visualizing a slice through hub-height (or other height) when comparing with a simulation.  The user can specify the height at which to export the slice.  At the moment it is a very slow process due to the looping and because it exports over the entire time range (this could be improved later).

Screenshot 2023-06-21 at 2 26 44 PM
This figure shows the XY plane at the hub height for a Mann wind field (HAWC format).  The vertical plane is exported from the driver using the `WrVTK` option.  Both are taken at 23.625 seconds.

none

InflowWind driver only

No test cases are affected.  I did not add a test case for this since it is a very slow code to execute.

### DIFF
--- a/modules/inflowwind/src/InflowWind_IO.f90
+++ b/modules/inflowwind/src/InflowWind_IO.f90
@@ -39,7 +39,8 @@ public :: IfW_SteadyWind_Init, &
 public :: Uniform_WriteHH, &
           Grid3D_WriteBladed, &
           Grid3D_WriteHAWC, &
-          Grid3D_WriteVTK
+          Grid3D_WriteVTK, &
+          Grid3D_WriteVTKsliceXY
 
 type(ProgDesc), parameter :: InflowWind_IO_Ver = ProgDesc('InflowWind_IO', '', '')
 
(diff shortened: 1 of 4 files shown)


modules/inflowwind/src/InflowWind_IO.f90
```diff
@@ -2924,4 +2925,79 @@ subroutine Grid3D_WriteHAWC(G3D, FileRootName, unit, ErrStat, ErrMsg)
...
+ 
+      do iy = 1, G3D%NYGrids
+         do ix = 1, G3D%NSteps      ! time and X are interchangeable
+            write (unit, '(3(f10.2,1X))') G3D%Vel(:, iy, iz, ix)
+         end do
+      end do
+
+   enddo
+
+   close (unit)
```
deslaughter, line 3000: Should `close (unit)` be inside the time step do loop?

modules/inflowwind/src/InflowWind_IO.f90
```diff
@@ -2924,4 +2925,79 @@ subroutine Grid3D_WriteHAWC(G3D, FileRootName, unit, ErrStat, ErrMsg)
...
 end subroutine Grid3D_WriteHAWC
 
+
+subroutine Grid3D_WriteVTKsliceXY(G3D, FileRootName, XYslice_height, unit, ErrStat, ErrMsg)
+
+   type(Grid3DFieldType), intent(in)   :: G3D            !< Parameters
+   character(*), intent(in)            :: FileRootName   !< RootName for output files
+   real(ReKi),    intent(in)           :: XYslice_height
+   integer(IntKi), intent(in)           :: unit           !< Error status of the operation
+   integer(IntKi), intent(out)          :: ErrStat        !< Error status of the operation
```
deslaughter, line 2935: Are `ErrStat` and `ErrMsg` initialized in this function? Maybe I missed it.

deslaughter (commented):
That's a cool feature. Saw a couple of things that should probably be fixed.

modules/inflowwind/src/InflowWind_IO.f90
andrew-platt: They aren't explicitly initialized, but get set either by `OpenFOutFile` or by the check on the slice height.  For completeness I can add them.

I also just noticed that there is no upper bounds check on the height, which could lead to an invalid index on `iz`.  I'll fix that also.

modules/inflowwind/src/InflowWind_IO.f90
andrew-platt: yes it should!  I had removed the file loop, then re-added it.

ebranlard:
If it's not too much work, maybe XZ and YZ could be added too (like fast farm). 


On a side note, plotting slices of TurbSim and MonnBoxes can easily be done using the python-toolbox:
https://github.com/OpenFAST/python-toolbox/blob/787092df79a558b849c45aa9c53adfb737beb386/pyFAST/input_output/examples/Example_TurbSimBox.py#L51

andrew-platt:
Adding the XY and YZ slices should be pretty straightforward.  I'll give that a shot.  Also, it might make sense to allow exporting multiple slices instead of just one.

Issue #1651, opened 2023-06-26
openFAST v3.5.0 crashing when setting BldNd_BladesOut to 0 in ED.dat


TimoHuelsmann:
**Bug description**
<A clear and concise description of the bug.>
The bug occurred when running openFAST v3.5.0 with the IEA15MW monopile model. 
To reduce the total number of output channels, the blade node outputs were tried to be eliminated in the ElastoDyn.dat, by setting the BldNd_BladesOut to 0 (as hinted in the readthedocs manual). 
The simulation aborts immediately, stating the error  _forrtl: severe (157): Program Exception - access violation_ in the console.

The error did not occur, when also removing the OutList-entries associated with the node output setting.

**To Reproduce**
<Update the following list with your specific information.>

To achieve this error, the ElastoDyn.dat section
```
---------------------- NODE OUTPUTS --------------------------------------------
 1   BldNd_BladesOut  - Blades to output
 99   BldNd_BlOutNd   - Blade nodes on each blade (currently unused)
              OutList     - The next line(s) contains a list of output parameters.  See OutListParameters.xlsx, ElastoDyn_Nodes tab for a listing of available output channels, (-)
"TDx"    - local flapwise (translational) deflection (relative to the undeflected position) of node
"TDy"    - local edgewise (translational) deflection (relative to the undeflected position) of node
"RDz"
END of input file (the word "END" must appear in the first 3 columns of this last OutList line)
---------------------------------------------------------------------------------------
```
has been changed to 
```
---------------------- NODE OUTPUTS --------------------------------------------
0   BldNd_BladesOut  - Blades to output
99   BldNd_BlOutNd   - Blade nodes on each blade (currently unused)
              OutList     - The next line(s) contains a list of output parameters.  See OutListParameters.xlsx, ElastoDyn_Nodes tab for a listing of available output channels, (-)
"TDx"    - local flapwise (translational) deflection (relative to the undeflected position) of node
"TDy"    - local edgewise (translational) deflection (relative to the undeflected position) of node
"RDz"
END of input file (the word "END" must appear in the first 3 columns of this last OutList line)
---------------------------------------------------------------------------------------
```
. The error occurred when running the simulation from windows power shell, as well as when running the simulation with Simulink.

**Expected behavior**
<A clear and concise description of what you expected to happen.>
I would have expected that it is enough to set the BldNd_BladesOut  to  0 to remove the blade node outputs from the .out-files.

[...]

PR #1654, opened 2023-06-26
ED/AD: fix segfault when BldNd_BladesOut==0

andrew-platt:
This PR is ready for merging.

If the number of blade node outputs is set to 0, a segmentation fault due to an array bounds issue occurs in both ElastoDyn and BeamDyn.  This is a temporary fix -- the logic really should be improved elsewhere in the code for checking if any blades are output.  However, we don't actually have logic in place yet for only outputting specified blades, so this temporary fix can be improved when that is done.

#1651

AeroDyn15 and ElastoDyn -- this only fixes a potential segfault if the number of blades to output in the nodal outputs is set to zero.


andrew-platt:
Fixes #1651

deslaughter (approved):

Issue #1622, opened 2023-06-13
Inconsistent Output and Assignment Error in fx_getMats Function during OC4 Wind Turbine Linearization in OpenFAST


whisperingkahuna:
**Bug description**
I am using the latest version of OpenFAST to carry out the linearization of the OC4 wind turbine. Three degrees of freedom are set as true, namely FlapDOF1, DrTrDOF, and GenDOF.

The wind speed is set at 18m/s. The option for CalcSteady is set as true, and the NLinTimes parameter is set to 36.

The process of linearization has been completed successfully, and the output has been generated as expected. However, when I try to execute the fx_getMats function from fx_getMats.m file in order to get the averaged matrices, I encounter the following error:

        fx_getMats(FileNames)
        Unable to perform assignment because the left and right sides have a different number of elements.

Error in fx_getMats (line 119)
matData.WindSpeed(iFile) = data(iFile).WindSpeed;

According to Jason, the following discrepancy in the .lin files could be causing the issue:

Current output: Wind Speed: *********** m/s
Expected output: Wind Speed: 18.0000 m/s

Even after trying again on a different computer with fresh files, the problem persists. Could this be a bug?

**Screenshots, if applicable**
Output screenshot
![plot_wind_2023-06-12_16-44-38](https://github.com/OpenFAST/openfast/assets/28399678/c35499e5-e6ab-4e1c-a8e4-890c5f22bd4a)

**OpenFAST Version**

![image (4)](https://github.com/OpenFAST/openfast/assets/28399678/d2994702-44b7-4307-a2b3-f02120480e0d)

**System Information (please complete the following information):**
 - OS: <Windows 11, 10>

**Additional context**
<linearization files for reference
[5MW_OC4Semi_WSt_WavesWN_lin files.zip](https://github.com/OpenFAST/openfast/files/11732524/5MW_OC4Semi_WSt_WavesWN_lin.files.zip).>

jjonkman:
Dear @whisperingkahuna,

Stars (***) would show up in an output if the internal value stored cannot be displayed in the internal Fortran format used for writing the output file.  I would not expect that you’d receive this issue with this wind speed output.  I’m not aware of a bug with this output, but perhaps something was introduced recently?

Can you also share your input files so that we can reproduce the error?
 
Best regards,

whisperingkahuna:
Dear @jjonkman jjonkman,
Thank you for your response.

Here are the files.

[...]

PR #1655, opened 2023-06-26
[BuFix] InflowWind was not setting InitOut MWS data.

andrew-platt:
This bugfix is **NOT** ready for merging.  Need to check if the `.lin` files in the regression tests are affected.

This fixes a bug in the reported windspeed from InfloWind to the glue code.  The wind speed was getting set to `HUGE()` which did not appear correctly in the `.lin` files.

#1622

Linearization was not reporting the correct wind speed in the `.lin` files.

~The regression tests may need to be rerun as the reported wind speed in the `.lin` files is `0.0000 m/s`.  At the moment this doesn't make much sense to me since the reported value should be `HUGE()`.~
The linearization regression tests are run without inflow.


deslaughter (commented):
@andrew-platt , I had intended to remove FileDat and pass InitOutData%WindFileInfo to the initialization functions instead, but I missed it. Could you make that change instead?

Issue #366, opened 2019-11-29
BeamDyn - Solution does not converge after the maximum number of iterations


JelmerPolman:
Dear all,

I have some problems with my BeamDyn model. The following error keeps popping up at some simulations: "FAST_Solution:FAST_AdvanceStates:B1:BD_GA2:BD_DynamicSolutionGA2:Solution does not converge after the maximum number of iterations". I have reduced the time step to 0.0001 seconds, which results in unreasonable simulation times. 

In order to troubleshoot I simulated the following testcase. Blade 1 is pitched 90 degrees and is in the three o'clock position and loaded with 1 g gravity. The rotor does not turn and AeroDyn is turned of so that only the deflection due to gravity is simulated. After a few time steps the simulation aborts.

![command_window](https://user-images.githubusercontent.com/50822421/69871368-e6919280-12b2-11ea-9b0d-2c87a20fb53f.JPG)

QuasiStaticInit is set to true. One can see that the blade deflections are indeed static (1.3 m), but the root bending moment in x-direction (although very small at 17 N.m) is not.

![results](https://user-images.githubusercontent.com/50822421/69871528-530c9180-12b3-11ea-82f4-36b461437a12.png)

The only thing that really helps to prevent this error is reducing the element order of BeamDyn. An element order of 5 for example runs without a problem. However, this order does not provide accurate results. I did a convergence study and calculated a reference solution with Ansys. From a comparison it follows that I need a element of order of at least 11 to get correct results.

![L2_norm](https://user-images.githubusercontent.com/50822421/69872394-923be200-12b5-11ea-9b70-52d9cebb8ce8.png)

If we look at the blade deflections, it can be clearly be seen that a low element order does not give accurate results. However, a high element order leads to a Runge's like oscillation in the torsion of the blade tip.

![out_of_plane_deflection](https://user-images.githubusercontent.com/50822421/69871977-800d7400-12b4-11ea-8d61-5f21f894f86c.png)

![in_plane_deflection](https://user-images.githubusercontent.com/50822421/69871981-826fce00-12b4-11ea-8c84-18dc35ca23fa.png)

![torsion](https://user-images.githubusercontent.com/50822421/69871893-489ec780-12b4-11ea-9b1c-ecc6e57e7998.png)

What can be the cause for OpenFAST to abort, althoug almost al channels are static? What steps can I take to reduce the oscillations in the blade torsion. Could this also be the cause for the needed very small time step?

I have attatched the input files here:

[...]

PR #1663, opened 2023-06-28
Resolve BeamDyn initial strain for rotated blade

deslaughter:

In the BeamDyn driver, and in the glue-code, if the blade was initially rotated then an initial strain was introduced which produced forces when no loads or gravity had been applied. The cause was traced to the calculation of R0 (initial rotation) at the quadrature points which was interpolated via the shape functions from the element nodal rotations. The shape function interpolation is not exact and introduced strains on the order of 1e-7, but with the large stiffness of the blade, forces of 6200 N could be produced at the root for the IEA 15MW blade. This was resolved by directly calculating the quadrature point initial rotations in uu0(4:6) via interpolated node position and twist, which are well behaved. This was only a problem for non-straight blades such as the IEA 15MW where the tangent vector was not aligned with the Z axis.

This PR also fixes a bug where the parameter `p%UsePitchAct` wasn't initialized from the input file until after it was used in `SetOutParam`. The initialization has been moved to the `SetParameters` subroutine in `BeamDyn.f90`.

This changes resolves Issue #366 when QuasiStaticInit is set to `False`. The original issue files have been updated to work with OpenFAST 3.5.0-dev [bd_sim.zip](https://github.com/OpenFAST/openfast/files/11898275/bd_sim.zip) and the simulation completes with a DT of 0.01 seconds (previously required DT to be 0.0001 seconds). Will need to investigate why this test fails when QuasiStaticInit is `True`.

BeamDyn

The BeamDyn unit test `test_BD_QuadraturePointData.F90` was updated to account for the slight difference in R0, E1, Stiffness that resulted from this change.

### DIFF
--- a/modules/beamdyn/src/BeamDyn.f90
+++ b/modules/beamdyn/src/BeamDyn.f90
@@ -177,7 +177,7 @@ SUBROUTINE BD_Init( InitInp, u, p, x, xd, z, OtherState, y, MiscVar, Interval, I
          return
       end if
 
-      ! Set the initial displacements: p%uu0, p%rrN0, p%E10
+      ! Set the initial displacements: p%uu0, p%E10
    CALL BD_QuadraturePointDataAt0(p)
       if (ErrStat >= AbortErrLev) then
          call cleanup()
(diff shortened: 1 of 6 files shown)


andrew-platt (approved):
This is a great solution!  Thanks for figuring this one out!

deslaughter:
The regression tests were failing randomly because of a difference in the way E10 and the Jacobian were calculated. The source of the problem was an inexplicable difference in the results from `matmul` and `dot_product`. With this corrected, no change is required to the regression test references.

ebranlard:
If it makes sense, is there anything we could add to the documentation based on what's implemented in the code and what was in the documentation? Since you spent some time figuring things out, it could be valuable to make the documentation clearer maybe.

andrew-platt:
I forgot there was some BD theory documentation in the docs.  It is probably worth modifying the two equations that are affected there.  The documentation in the code was updated accordingly, but keeping that in sync with the other docs would be good.

deslaughter:
The equations in the documentation still apply to the code, just the implementation changed slightly. The part that is missing is how `R0` is constructed from the tangent vector and the blade twist. I can add that if you think it would be worthwhile since it isn't currently in the docs.

andrew-platt:
Thanks for the clarification.  Given that it is just a subtlety of how `R0` is derived, I don't think we need to add it to the documentation.  I don't think the documentation doesn't need to cover all the tiny nuances.

PR #1679, opened 2023-07-11
AD: outputs of velocity in different coordinate systems and additional documentation in BEMTUncoupled

ebranlard:

As we are preparing to update the BEMT model of AeroDyn 15, it is convenient to output variables in "well-defined" coordinates systems that are independent of the underlying coordinate system used for the momentum balance. This pull request also adds additional documentation in the functions used inBEMTUncoupled. 

AeroDyn. 

This pull request introduces the following coordinate systems:
- "i" inertial (already used)
- "p": poalr coordinate system (same as hub but shifted by the azimuthal offset of each blade)
- "l" local polar grid. Contains an additional azimuthal rotation compared to the rotating hub corresponding to the sweep at a given radial node. 
- "a": the airfoil section coordinate system with xa normal to the chord. 

With these coordinate systems, the following velocity outputs are added (For BEM but also OLAF): 
- Induced velocitties:  `Vind*i, Vind*p` (Previously called `Uin, Uit, Uir`) , `Vind*l`, `VInd*a`
- Structural velocity: `STV*i, STV*p, STV*l`, `STV*a`
- Undisturbed velocities: `VUnd*i`, etc.
- Disturbed velocity: `VDis*i`, etc.
- Forces: `F*i`, etc
- Moments: `F*i`, etc
- Buoyancy loads. 

The following BEM specific outptus are added (to more easily debug these outputs are compare between the old and new BEM) (this required to add them as optional outputs to some BEM functions):
- AxInd_qs: the axial induction factor as obtained by the quasi-steady BEM (before DBEMT and Skew)
- TanInd_qs: the axial induction factor as obtained by the quasi-steady BEM (before DBEMT and Skew)
- F: the tip/hub loss factor
- k, kp: the dimensionless load coefficients that appear when manipulating the thrust and torque coefficients from the Blade Element Theory. 
- CT_qs: The quasi-steady thrust coefficient. This is important as this is the coefficient which is corrected for high-thrust. With these outputs, we can plot the a_qs-CT_qs relationship internally used by BEM for verification.  

In addition, the following is done:
- the variable `psi` is renamed as `psi_s` as this is the skew azimuthal position, to avoid confusion with the azimuthal angle of the blade. 
- Additional documentation is provided in BEMTUncoupled
- A cleanup of the "future" BEM code is carried out. 

**Documentation**

The coordinate systems are now documented ([here](https://ebranlard-openfast.readthedocs.io/en/ad-moreout-newbemprep/source/user/aerodyn/coordsys.html)). 

Further focumentation will follow as part of the release of the new BEM code (v4.0?) accompanied with a paper. 

The test results should be unchanged. 

Checklist:

 - [x] Documentation of coordinate systems ([here](https://ebranlard-openfast.readthedocs.io/en/ad-moreout-newbemprep/source/user/aerodyn/coordsys.html))
- [x] Adding outputs for Buoyancy
- [x] Rename Fix RtAeroF*g  to RtAeroF*i
- [x] Final review


modules/aerodyn/src/BEMT.f90
```diff
@@ -1452,9 +1475,12 @@ subroutine BEMT_CalcOutput_Inductions( InputIndex, t, CalculateDBEMTInputs, Appl
          !............................................
          ! get BEMT inductions (axInduction and tanInduction):
          !............................................
-         call calculate_Inductions_from_BEMT(p, phi, u, OtherState, AFInfo, axInduction, tanInduction, ErrStat2, ErrMsg2)
+         call calculate_Inductions_from_BEMT(p, phi, u, OtherState, AFInfo, axInduction, tanInduction, ErrStat2, ErrMsg2, k_out, kp_out, F_out)
```
andrew-platt, line 1478: Will we run into compatibility issues with older compilers passing optional arguments like this to another subroutine?

andrew-platt (approved):

modules/aerodyn/src/BEMT.f90
ebranlard: Thanks. Yeah, I was feeling shaky while implementing this. Right now I don't really have another way of doing it.... So... fingers crossed?

ebranlard:
I might add a couple of small changes maybe even some more outputs, and potentially do some renaming in the coming days.
But I know rc3.5.1 is coming soon. The main thing I want to add is the loads along the blade in the inertial frame and potentially another coordinate system.

modules/aerodyn/src/BEMT.f90
andrew-platt: Since those arguments are optional in `calculate_Inductions_from_BEMT`, you could do something like:
`If (present(k_out)) then
   call calculate_Inductions_from_BEMT(p, phi, u, OtherState, AFInfo, axInduction, tanInduction, ErrStat2, ErrMsg2, k_out, kp_out, F_out)
else
             call calculate_Inductions_from_BEMT(p, phi, u, OtherState, AFInfo, axInduction, tanInduction, ErrStat2, ErrMsg2)
endif`

or something similar.  From what I could tell, it looks like you have either none or all of those arguments in the parent routine.

andrew-platt:
Sounds good!  We can keep this open until rc-3.5.1 is closer to ready (it's taking longer to get some pieces in that expected).

modules/aerodyn/src/BEMT.f90
ebranlard: That's a great suggestion, I've now updated the code.

ebranlard (commented):
This is now ready.

docs/source/user/aerodyn/appendix.rst
```diff
@@ -51,27 +51,34 @@ The blade data input file contains the nodal discretization, geometry, twist, ch
...
 
-This is a list of output parameters for the AeroDyn module.  The names are grouped by meaning, but can be ordered in the OUTPUTS section of the AeroDyn input file as you see fit.  :math:`B \alpha N \beta`, refers to output node :math:`\beta` of blade :math:`\alpha`, where :math:`\alpha` is a number in the range [1,3] and :math:`\beta` is a number in the range [1,9], corresponding to entry :math:`\beta` in the :math:`\textit{BlOutNd}` list.  :math:`\textit{TwN}\beta` refers to output node :math:`\beta` of the tower and is in the range [1,9], corresponding to entry :math:`\beta` in the :math:`\textit{TwOutNd}` list. A comprehensive, up-to-date list of all possible output parameters is given in the Excel file :download:`OutListParameters.xlsx <../../../OtherSupporting/OutListParameters.xlsx>`.
 
-The local tower coordinate system is shown in :numref:`ad_tower_geom` and the local blade coordinate system is shown in :numref:`ad_blade_local_cs` below.  Figure :numref:`ad_blade_local_cs` also shows the direction of the local angles and force components.
+AeroDyn has regular outputs (see :numref:`AD-Nodal-Outputs`) and nodal outputs (see :numref:`AD-Outputs`).
```
andrew-platt, line 55: The links are switched.

andrew-platt (approved):

docs/source/user/aerodyn/appendix.rst
ebranlard: thanks for spotting that!

PR #1631, opened 2023-06-16
Initial AeroMap changes for ElastoDyn and BeamDyn

bjonkman:
**Feature or improvement description**
This PR updates ElastoDyn and BeamDyn with changes needed for AeroMap generation (which will be added in a later pull request, after AeroDyn and InflowWind are also modified). In AeroMaps, linearization is done on only one blade, with a subset of states. Thus, the linearization code in these modules must change.

The flag that activates the new code is set to false, so the code should behave as it did before.

I also updated some comments in FVW and modified some error handling in the ED and BD nodal outputs (it will now still output some of the requested outputs if the number requested exceeds size limits)

**Related issue, if one exists**
None

**Impacted areas of the software**
ElastoDyn and BeamDyn, particularly with linearization.

**Additional supporting information**

**Test results, if applicable**
This should not change test results.

PR #1715, opened 2023-08-03
AD: Initial AeroMap changes for AeroDyn and misc UA/DBEMT changes

ebranlard:

This pull request introduces:
- If statements in AeroDyn to perform the linearization differently when the AeroMap flag is set
- Uses `UA%lin_xIndx` in UA and AeroDyn to store the indices of the states in the operating point vector `x_op`
- Uses `utimes` in DBEMT in cases the inputs are at multiple times (and not just dt and t+dt). 
- Uses double precision variables for some of the critical variables of UA
- Prepares the ground for a followup pull request that will affect the test results  (introducing feature to avoid linearization of non active UA states, use of a stiff integrator, introducing `u%AvgDiskVel`)

AeroDyn, UA, DBEMT

Keeping track of state indices can be difficult, in particular if we desactivate some UA nodes. 
The `AeroMap` feature is progressively being introduced in OpenFAST. See #1631

Should be unaffected

### DIFF
--- a/modules/aerodyn/src/FVW_Subs.f90
+++ b/modules/aerodyn/src/FVW_Subs.f90
@@ -462,7 +462,6 @@ logical function have_nan(p, m, x, z, u, label)
    character(len=*),                intent(in) :: label !< label for print
    integer :: iW
    have_nan=.False.
-!bjj: If we used Is_NaN (or a version of it for this data type) instead of isnan, I'd get fewer compiler warnings that "Fortran 2003 does not allow this intrinsic procedure."
    do iW = 1,size(p%W)
       if (any(isnan(x%W(iW)%r_NW))) then
          print*,trim(label),'NaN in W(iW)%r_NW'//trim(num2lstr(iW))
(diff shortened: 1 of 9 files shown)


modules/aerodyn/src/FVW_Subs.f90
```diff
@@ -481,11 +480,11 @@ logical function have_nan(p, m, x, z, u, label)
          have_nan=.True.
       endif
       if (any(isnan(x%W(iW)%Eps_NW))) then
-         print*,trim(label),'NaN in E_NW'
+         print*,trim(label),'NaN in E_NW'//trim(num2lstr(iW))
```
andrew-platt, line 483: Could we change the `print*` to `call WrScr` instead?  The `print*` doesn't show up properly with Simulink.

ebranlard: Those are debug print statements, and are never called, so I'm not worrying about them

andrew-platt: Perfect!  No need to worry about them then.  [apologies for my laziness in not checking the context better]

andrew-platt (approved):

Issue #1746, opened 2023-08-24
Bug report--Error when conduct simulation using OLAF & turbulent wind (TurbSim)


ZL6462:
Dear all, 

I met a bug when conduct simulation using OLAF & turbulent wind (TurbSim). The simulation object is the monopile IEA-15MW wind turbine downloaded from "https://github.com/IEAWindTask37/IEA-15-240-RWT/tree/master/OpenFAST". The error informantion is "forrtl: severe (157): Program Exception - access violation10:53:31."

![image](https://github.com/OpenFAST/openfast/assets/98078424/90a390a5-99f8-42ca-8968-bbd3501ddcbc)

After a preliminary judgment, I think the error may occur when the wake of the rotor beyond the boundary of the wind field generated by TurbSim. I have tried to enlarge the wind field to avoid the error, but with the development of the wake, there are still  some vortices enter the ground and therefore cross the boundary of the wind field, so the error occurred again.

Some Information of the executable program is given as follows:
1. OpenFAST version 3.5.0
2. Complied with "Intel(R) Fortran Compiler 1800" with OpenMP on.

Thank you for your attention.

ebranlard:
Hi @ZL6462 

Could you package all your input files (including the input file for turbsim) into a zip file and share them here? 

If you know how to compile OpenFAST, can you also  try to compile the branch [rc-3.5.1](https://github.com/OpenFAST/openfast/tree/rc-3.5.1) and see if you still have the issue?

ebranlard:
You can also compile by activating the `traceback` in Visual studio.  
Right click on FASTLib in the solution Explorer > `Properties` > `Fortran` > `Run-time` > `Generate Traceback Information  = Yes`

This might give us a better idea of where the access violation occurs.

ZL6462:
Dear @ebranlard,

Follow your suggestion, 

1. All the input files are attached here:  [input files.zip](https://github.com/OpenFAST/openfast/files/12437456/input.files.zip)
2. I have activate the "traceback", and the error INFO are shown as follow, it maybe the same as my judgement, namely the wake of the rotor beyond the boundary of the wind field grid.
3. I have also compile the version "rc-3.5.1", and same error occur.

Maybe, we cannot call this error a bug, because under the turbulent wind condition, it is the wake of the rotor is of course somewhat disordered, so it may beyond the wind field boundary (for example when the wake enter the ground). Even though, I still wonder how to avoid this if possible.

Thank for your attention.

[...]

PR #1748, opened 2023-08-25
IfW: Bug Fix: compute vel avg profile when exceed allowed

ebranlard:
This pull request ready to be merged

When using OLAF, VelAvg was not computed, and box extrapolation using VelAvg would result in  access violations. 

#1746

InflowWind, OLAF

We could also implement periodic boundary conditions (for lateral extent) instead of using the average profile.


andrew-platt:
Good enough solution for now :).  Thanks for addressing this!

deslaughter:
@ebranlard Thanks for fixing!

Issue #1750, opened 2023-08-28
Bug report - MAP++ Problems with horizontal lines


RBergua:
**Bug description**
Some wave tank testing campaigns use horizontal mooring lines instead of catenary lines (see for example Section 2.4 from https://www.mdpi.com/2075-1702/11/9/865). In this case, the lines can be considered massless (monofilament fishing lines attached to a spring) and no hydrodynamic loading is considered (Cd = 0, Ca = 0, Cm = 0). This massless nature (i.e., no inertial loading in the lines) and lack of hydrodynamic forces would make the system a perfect candidate to be analyzed with MAP++ (quasi-static approach). 
Unfortunately, when running the system with MAP++, the system response is not the expected one. For example, when running a roll free decay, the outputs from MAP++ and MoorDyn v2 are clearly different. See the below plot for reference:
![image](https://github.com/OpenFAST/openfast/assets/33709140/bfca0f2f-3130-4d9c-bcae-cab44658bce9)

As it can be observed, the period in MAP++ is longer than in MoorDyn v2.

When looking at the fairlead tensions in the three mooring lines, the general trend is very similar. However, some differences can be observed between both solutions.
![image](https://github.com/OpenFAST/openfast/assets/33709140/856ddc78-5732-49c2-82ca-6314a4a855db)

Interestingly, when looking at the 3 component forces (e.g., Fx, Fy, Fz) in the fairleads, clear differences between MAP++ and MoorDyn v2 can be observed. For the roll free day, it is expected that the vertical forces from lines 1 and 2 (located downwind) oscillate around positive and negative values. While the vertical force for line 3 (located upwind) should remain more or less constant. Somehow, when looking at the fairlead 1 vertical force, it can be observed that the loading in MoorDyn v2 oscillates to positive and negative values (as expected). However, in MAP++ the force is mirrored around 0 N. MAP++ cannot return negative forces in this vertical direction. Note that this negative component does not indicate line slack. There is still a large tension in the line (see plot above for reference). This negative Fz only indicates that the vertical component in this horizontal line becomes negative.
![image](https://github.com/OpenFAST/openfast/assets/33709140/3fb11710-555b-49bb-852c-381eb28a8ede)

This impacts the system dynamics as shown in the first screenshot.    

**To Reproduce**
The input file used for MAP++ is attached below for reference:
`---------------------- LINE DICTIONARY ---------------------------------------`
[...]

PR #1752, opened 2023-08-30
MAP++ linear spring approach improvement

RBergua:
Small modification for the linear spring approach in MAP++. This modification allows to study conventional catenary systems as well as lines that have the anchor point above the fairlead connection.

The default MAP++ version can only apply vertical forces in the downward direction. This is the force that we would expect in a conventional mooring line system. However, there are some conditions when the force should be applied in the upward direction (e.g., when then anchor location is above the fairlead connection).

The proposed modifications allows MAP++ to apply positive or negative forces in the vertical direction when using the linear spring approach.

For reference MAP++ can solve a mathematical problem by means of a catenary formulation or a linear spring formulation. The modifications presented here, only impact the linear spring formulation.

This modification solves the problems for the linear spring approach in MAP++ reported here: https://github.com/OpenFAST/openfast/issues/1750

The modifications only affect MAP++ when the `linear_spring` flag is selected in the line properties section.

Sometimes, when performing scale tests of floating offshore wind turbines in a wave tank, horizontal mooring lines are used. Usually monofilament fishing lines connected in series to a spring are used to reproduce a similar elastic restoring force as the corresponding catenary line. See for reference section 2.4 of the paper: https://doi.org/10.3390/machines11090865. For this kind of applications, the linear spring approach from MAP++ is a convenient approach to simulate the system.

Several tests cases have been performed with satisfactory results. The outputs from the modified code (MAP++) has been also compared against MoorDyn v2. See for reference: https://github.com/OpenFAST/openfast/issues/1750

### DIFF
--- a/modules/map/src/lineroutines.c
+++ b/modules/map/src/lineroutines.c
@@ -667,7 +667,7 @@ MAP_ERROR_CODE solve_linear_spring_cable(Line* line, char* map_msg, MAP_ERROR_CO
   };
 
   *(line->H.value) = sqrt(force.x*force.x + force.y*force.y);
-  *(line->V.value) = fabs(force.z);
+  *(line->V.value) = -force.z; /* vertical force can be positive or negative. Important for systems with the anchor point above the fairlead connection */
   return MAP_SAFE;
 };
 


andrew-platt:
Is this something we may need to change for MoorDyn as well?

RBergua:
MoorDyn works as expected. This problem is only related to MAP++.

andrew-platt:
Thanks @RBergua!

andrew-platt (approved):

PR #1785, opened 2023-09-27
Wake-added turbulence in FAST.Farm

ebranlard:

Implementation of the wake-added turbulence into FAST.Farm

FAST.Farm

The feature and input file changes has been documented, but it requires the calibration of its two main constants. 
r-tests and additional documentation will follow after the calibration. 

The following additional inputs are added to the input file:
```python
--- WAKE-ADDED TURBULENCE ---
0                  WAT                - Switch between wake-added turbulence box options {0: no wake added turbulence, 1: predefined turbulence box, 2: user defined turbulence box} (switch)
"Inflow/WAT/isoturb_1024x32x16.u" WAT_BoxFile  - Filepath to the file containing the u-component of the turbulence box (either predefined or user-defined) (quoted string)
1024, 32, 16       WAT_NxNyNz         - Number of points in the x, y, and z directions of the WAT_BoxFile [used only if WAT=2, derived value if WAT=1] (-)
5.5, 4.0, 4.0      WAT_DxDyDz         - Distance (in meters) between points in the x, y, and z directions of the WAT_BoxFile [used only if WAT=2, derived value if WAT=1] (m)
default            WAT_ScaleBox       - Flag to scale the input turbulence box to zero mean and unit standard deviation at every node [DEFAULT=False] (flag)
0.5                WAT_k_Def          - Calibrated parameter for the influence of the wake deficit in the wake-added turbulence [>=0.0, no upper bound] or DEFAULT [DEFAULT=NONE]
0.5                WAT_k_Grad         - Calibrated parameter for the influence of the radial velocity gradient of the wake deficit in the wake-added turbulence [>=0.0, no upper bound] or DEFAULT [DEFAULT=NONE]
default            WAT_D_BrkDwn       - Downstream distance in rotor diameter after which WAT scaling has reached 100% capacity [>=0] or DEFAULT [DEFAULT=1] (D)
```

Test results are unchanged.


andrew-platt:
Temporary links to readthedocs changes for proofing:
https://ap-openfast.readthedocs.io/en/f-ffwat/source/user/fast.farm/InputFiles.html#wake-added-turbulence-wat
https://ap-openfast.readthedocs.io/en/f-ffwat/source/user/fast.farm/FFarmTheory.html#wake-added-turbulence-wat

docs/source/user/fast.farm/InputFiles.rst
```diff
@@ -624,11 +625,58 @@ If DEFAULT is used, then **Mod_Projection=2** when **Mod_Wake=2**,
...
+Currently uncalibrated, no default value. Anticipated value is expected to be around 0.5.
+
+**WAT_k_Grad** [float]
+Parameter for the influence of the radial velocity gradient of the wake deficit in the wake-added turbulence 
+[>=0.0, no upper bound].
+Currently uncalibrated, no default value. Anticipated value is expected to be around 0.5.
+
+**WAT_D_BrkDwn** [float, dimensionless]
+Downstream distance in rotor diameter after which WAT scaling has reached 100% capacity.
+DEFAULT is 1 (i.e. 1D)
```
jjonkman, line 676: We've discussed potentially replacing input `WAT_D_BrkDwn` with one of the filter functions involved in the eddy viscosity formulation.  I would guess F_vShr(x) would make sense because it represents the development of turbulent stresses generated by the wake shear layer, which seems to be what `WAT_D_BrkDwn` is trying to achieve.  The only issue I see is the DEFAULT values of `C_vShr_DMin`, `C_vShr_DMax`, `C_vShr_FMin`, and `C_vShr_Exp` used to define F_vShr(x) result in a filter function that does not approach unity until 25D.  This would impact the calibration of `WAT_k_Def` and `WAT_k_Grad`.

docs/source/user/fast.farm/InputFiles.rst
```diff
@@ -750,6 +798,12 @@ printed using the “F10.4” format. **OutFmt** is not used when
 **OutFileFmt** = 2. See :numref:`FF:Output:Time` for
 details on time-series results files.
 
+
+**OutAllPlanes** [-] Output all wake planes in VTK at all time steps. 
+Note: this option requires intensive writing to disk and will drastically slow down the simulation.
+DEFAULT is False.
```
jjonkman, line 804: `OutAllPlanes` should be labled as a flag.

docs/source/user/fast.farm/InputFiles.rst
```diff
@@ -408,6 +408,7 @@ The wake will adopt a "curled" shape in skewed inflow.
 When Wake_Mod=2,3, the stability of the algorithm will depend on the choice of **dr** and **DT_Low** (see the guidelines (see the guidelines given in :numref:`FF:ModGuidance`).
 
 
+**RotorDiamRef** [float, in m]: Reference turbine rotor diameter for wake calculations.
```
jjonkman, line 411: How is RotorDiamRef used in the code?  Does it replace EstimatedRotorRadius and/or D_rotor_filt?

docs/source/user/fast.farm/FFarmTheory.rst
```diff
@@ -1291,6 +1292,135 @@ the solution of the momentum equation),
...
+
+The scaling factor, expressed in terms of the wake deficit :math:`V_x^{Wake}`, is determined at each wake plane as:
+
+.. math::
+
+   \begin{aligned}
+      k_{} (x,y,z) = 
+          \frac{k_\text{def} }{\overline{U}} \left| V_x^{Wake}(x,y,z) \right|
+        + \frac{k_\text{grad}D}{2\overline{U}} \left[\left|{\frac{\partial {V_x^{Wake}(x,y,z)}}{\partial r}}\right| +  \left|{\frac{1}{r}\frac{\partial {V_x^{Wake}(x,y,z)}}{\partial \theta}}\right|  \right]  
+        \end{aligned}
```
jjonkman, line 1337: k should be pre-multiplied by F_vShr(x) if we choose do so (see below).

docs/source/user/fast.farm/FFarmTheory.rst
```diff
@@ -1291,6 +1292,135 @@ the solution of the momentum equation),
...
+
+where :math:`D` is a reference diameter, and :math:`\bar{U}` is the mean velocity taken as the filtered velocity at the turbine location normal to the rotor disk.
+The coordinates :math:`x,y,z` and :math:`r,\theta` are taken in the meandering frame of reference.
```
jjonkman, line 1340: Suggest clarifying that the last term (involving r,\theta) is zero for the polar wake and nonzero for the curled wake.

docs/source/user/fast.farm/FFarmTheory.rst
```diff
@@ -1291,6 +1292,135 @@ the solution of the momentum equation),
...
+
+      \alpha\epsilon^{2/3}\approx \frac{1}{0.688 L^{2/3}} 
+
+We have found that there is no dependency on the length scale. We nevertheless recommend to set it to the rotor diameter if the users generate their own boxes.
+
+
+**Predefined boxes**
+
+A recommended practice for the high-resolution domain of FAST.Farm is to chose a grid
+spacing equal to the maximum chord of the blade.
```
jjonkman, line 1365: This is no longer true.  See: https://github.com/OpenFAST/openfast/issues/1822.  Will this impact WAT?

docs/source/user/fast.farm/FFarmTheory.rst
```diff
@@ -1291,6 +1292,135 @@ the solution of the momentum equation),
...
+      \frac{d\boldsymbol{B}}{dt} = \boldsymbol{U}_\text{farm}(t)
+
+where:
+
+.. math::
+
+      \boldsymbol{U}_\text{farm} = \operatorname{mean}\{ \overline{V}^\text{Low}_\text{Amb}[i_w], i_w =1\cdots n_{WT}\}
+
+where :math:`\overline{V}^\text{Low}_\text{Amb}[i_w]` is the rotor
+averaged ambient wind speed.
```
jjonkman, line 1399: Suggest noting here that the convection of the turbulence box may change if you add or remove turbines from the wind farm.

docs/source/user/fast.farm/FFarmTheory.rst
```diff
@@ -1291,6 +1292,135 @@ the solution of the momentum equation),
...
+
+Wake-Added Turbulence (WAT)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Wake-added turbulence is the additional small-scale turbulence generated
+from the turbulent mixing in the wake. 
+It is modeled by scaling up a background (undisturbed) turbulence.
+
+
+The theory for WAT will be presented in more details in the future as part of a journal publication.
```
jjonkman, line 1307: We can address this later, but I would say that the journal paper should document the WAT model at high-level, and the details regarding the implementation in FAST.Farm should be stated here.  (This would be consistent with the rest of the FAST.Farm theory section documented here.)

jjonkman (commented):
I've reviewed the documentation part of this PR.  Please find my edits and comments below.

docs/source/user/fast.farm/InputFiles.rst
```diff
@@ -624,11 +625,58 @@ If DEFAULT is used, then **Mod_Projection=2** when **Mod_Wake=2**,
...
+
+**WAT_BoxFile**: [quoted string]
+Filepath to the file containing the u-component of the turbulence box 
+(either predefined or user-defined). This turbulence box is expected to be 
+in the Mann box format. 
+The filename should be of the form "Label_1024x32x16.u" where the extension ".u" 
+is required. FAST.Farm will assume that files with extensions ".v" and ".w" are present,
+and these files will also be read. 
+When **WAT=1**, the three digits separated by "x" in the filename are assumed to be 
+the number of points in the x y and z dimension of the box.
```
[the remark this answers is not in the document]
ebranlard: Currently we don't use the ".bin" extension, the extensions are ".u", ".v", ".w", so I think the documentation should be good as it is. 

I've added the precision regarding the quotation and relative/absolute paths. 

There is no predefined dimensions, these can be any dimensions chosen by the user when generating the Mann box. The idea is to push people to use this convention in the filename because Mann Boxes do not store this information anywhere, and it is a critical information to be able to read the box correctly. Most likely, we will provide the user with a bsaeline box with dimensions 512x512x64, so I've changed the description to accomodate for that.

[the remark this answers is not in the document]
ebranlard: I personally think we should get rid of `WAT_NxNyNz` and force the user to use this filename convention.

docs/source/user/fast.farm/InputFiles.rst
```diff
@@ -624,11 +625,58 @@ If DEFAULT is used, then **Mod_Projection=2** when **Mod_Wake=2**,
...
+Number of points in the x, y, and z directions of the WAT_BoxFile.
+These are used only if WAT=2. 
+
+**WAT_DxDyDz**: [three floats, comma separated] 
+Distances (in meters) between points in the x, y, and z directions of the WAT_BoxFile 
+These are used only if WAT=2.
+When **WAT=1** the dimensions in each directions are taken as :math:`dx=dy=dz=0.03*\text{RotorDiamRef}`.
+
+**WAT_ScaleBox**: [flag]  
+When set to True, the input turbulence box is scaled so that it has zero mean and unit standard deviation at every node.
```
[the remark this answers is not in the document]
ebranlard: I've added the following precision:

The Mann Box generator typically ensures that the wind speed mean is close to zero on average.
The predefined boxes provided with OpenFAST have been generated such that their mean is close to zero and standard deviation is close to one *on average* (see :numref:`FF:WAT`, for how the Mann Box inputs are determined). 
The flag **WAT_ScaleBox** enforces that these properties are true *at every node* of the box, irrespecitvely of how the Mann Box was generated.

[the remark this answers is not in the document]
ebranlard: Note: zero mean is not a problem and is typical for Mann Boxes, but for the WAT reader we adapted how the conversion of time / space is done to convect the box at the speed we want.

docs/source/user/fast.farm/InputFiles.rst
ebranlard: I think it might be risky to rely on the same inputs indeed. 

But it's true that we could use the function `EddyFilter`, and call it with `C_Dmin=0` and `C_Dmax = WAT_D_BrkDwn`
As for the `FMin` and the exponent, we could use the values from the input file, but they might represent something a bit different, maybe it's best to hardcode them to say `FMin=0` and `Exp=1`.

docs/source/user/fast.farm/InputFiles.rst
ebranlard: I've added the following: 

It is used to:
verify that the high-res boxes are sufficiently large cover the rotor, 
determine the default value of **f_c** (see below), 
determine the physical grid spacing of the `WAT_BoxFile` when **WAT=1** (see below).

docs/source/user/fast.farm/FFarmTheory.rst
ebranlard: I've added the scaling, with the following text:  

The function :math:`k_\text{scale}` is used to ramp up the onset of WAT, with a value of 0 at the rotor and 1 when the wakes are beyond the input file parameter **WAT_D_BrkDwn**.
A linear scaling is currently used.

docs/source/user/fast.farm/FFarmTheory.rst
ebranlard: I think we can keep this for WAT as the chord max is a good recommendation for the loads. But we could use D/15 as well (about double)

jjonkman: Never mind.  My prior comment was incorrect.  The D/15 guidance is for dr, not DS_High.  I agree that setting DS_High equal to max chord is good guidance.

docs/source/user/fast.farm/InputFiles.rst
jjonkman: I'm wondering if we should hold off on deciding this until the calibration is further along.  I recall thinking before that the calibration may require different values at different downstream distances, which may work well if we have a filter function that is downstream-distance-dependent.

andrew-platt:
This PR is replaced by https://github.com/OpenFAST/openfast/pull/2202

Issue #1787, opened 2023-09-27
Questions about the operation of ad_VerticalAxis_OLAF and Turbsim


engineerYuXN:
Dear all
         I have built a vertical axis wind turbine with multiple rotors in ad_VerticalAxis_OLAF, and I want to use the turbulent wind generated by turbsim to do the calculations, but no matter how I change the size of the wind field (increase the area of the wind field or decrease the area of the wind field), I always get an error and stop the calculations around the middle of the calculations, and the specific error is reported as follows.
![aerodyn报错](https://github.com/OpenFAST/openfast/assets/87838413/3029e519-3b53-46a3-b716-7c088ac09f57)
The following diagram shows the relevant settings about the model
![22](https://github.com/OpenFAST/openfast/assets/87838413/d41e639b-33f1-4c64-b3a6-5d1f8873dce1)
The following diagram shows the relevant settings about the ad_driver.dvr
![33](https://github.com/OpenFAST/openfast/assets/87838413/8c033802-25a5-4712-b210-fba5093c4962)
![44](https://github.com/OpenFAST/openfast/assets/87838413/5d1e797d-6c24-446f-bca1-9c072708526c)
![55](https://github.com/OpenFAST/openfast/assets/87838413/21eafa25-8944-4ae0-8fb7-6ba6d89df97c)
![66](https://github.com/OpenFAST/openfast/assets/87838413/8faa24e3-61f0-49d3-927b-8893d1b514c2)
The following diagram shows the relevant settings about the Turbsim.inp
![77](https://github.com/OpenFAST/openfast/assets/87838413/be079dbf-905f-41ca-ba31-cfefbaa9bb09)
![88](https://github.com/OpenFAST/openfast/assets/87838413/ad75664a-ce0a-4dc3-9af5-8c4ca8c2eff5)
I hope you can give me some suggestions for improvement based on my input file and explain the reason for the error.

Thank your for your attention.
Best regards,

engineerYuXN:
@jjonkman @ebranlard

ebranlard:
Hi @engineerYuXN 

There has been a couple of bugs in InflowWind lately. I can see that you are using 3.4.1. If you can, try to compile the version for 3.5.1, which is in the branch rc-3.5.1 of this repository. 

You can also try to increase you TurbSim height, so that the box starts right above the gound (say 0.05m above the ground instead of 5m) . 
Let us know if that helps,

Emmanuel

engineerYuXN:
[...]

PR #1791, opened 2023-10-04
Apply the FakeGroundEffect to all FWspan points

rcorniglion:
Ready to be merged.

Bug fix. The FakeGroundEffect is applied to all FW nodes.  Current code did not apply the FakeGroundEffect to all FW points.

Possibly related to is issue #1787 since FW points can go below the ground.

aerodyn

ad_VerticalAxis_OLAF with longer simulation time, free far-wake and turbine closer to the ground shows that FW points can go below the ground. This is fixed with the correction.


ebranlard (commented):
Thanks for doing that, the loop indeed needs to go to `FWnSpan+1`

Preferably we should have only one warning, but since the far wake is less and less used, I think that's fine.

ebranlard (approved):

andrew-platt:
considering backporting this to rc-3.5.4

Issue #1781, opened 2023-09-23
Bug report-Error when running the case "ad_QuadRotor_OLAF".


ZL6462:
Dear all, 

When I try to run the case "ad_QuadRotor_OLAF" from "https://github.com/OpenFAST/r-test/tree/main/modules/aerodyn/ad_QuadRotor_OLAF", an error occur:
![image](https://github.com/OpenFAST/openfast/assets/98078424/b94dcdf5-dcbc-4d41-9bb6-b4bf7dd34f07)
The version of the aerodyn.exe is 3.5.0, and I have also try some other version such as v3.3.0 and v3.1.0 with their corresponding r-test version, and all filed.

Thank you for your attention.

andrew-platt:
How did you compile `aerodyn.exe`?  We didn't provide an executable in the 3.5.0 release, so I want to verify I can reproduce this issue with the same compilation method you used.

andrew-platt:
I can confirm this issue exists in 3.5.0 and in the rc-3.5.1 branch.  Fixed in PR #1811.

@ZL6462, thanks for lettings us know about this issue!

PR #1811, opened 2023-10-16
AD15 driver: segmentation fault with intel compiler when no blades

andrew-platt:
This is ready to merge.

Issue #1781 highlights a bug that appears with the intel compiler.  The `DiskAvgValues` routine can trigger a segmentation fault if there are no blades on a turbine.  The `ad_QuadRotor_OLAF` case has 5 effective turbines in the model, but only 4 rotors.  When the 5th rotor is evaluated by the `DiskAvgValues` with the Intel compiler, a segmentation fault or underflow can occur (I've seen both during debugging).  I'm not entirely sure how, but the line `m%AvgDiskVelDist = m%AvgDiskVelDist / real( p%NumBlades * p%NumBlNds, ReKi )` (AeroDyn.f90 line 2859) was triggering the access violation in mentioned bug report (reproduced on v3.5.0).

#1781

AeroDyn driver, only when a multi-rotor case is evaluated with one rotor having no blades.

This doesn't appear to be an issue with any of the gcc compilers, which is why we had not seen it in testing.

### DIFF
--- a/modules/aerodyn/src/AeroDyn.f90
+++ b/modules/aerodyn/src/AeroDyn.f90
@@ -2848,6 +2848,8 @@ subroutine DiskAvgValues(p, u, m, x_hat_disk, y_hat_disk, z_hat_disk, Azimuth)
    ! calculate disk-averaged velocities
    m%AvgDiskVel = 0.0_ReKi
    m%AvgDiskVelDist = 0.0_ReKi ! TODO potentially get rid of that in the future
+   if (p%NumBlades <= 0) return  ! The Intel compiler gets array bounds issues in this routine with no blades.
+
    do k=1,p%NumBlades
       do j=1,p%NumBlNds
          m%AvgDiskVelDist = m%AvgDiskVelDist + m%DisturbedInflow(:,j,k)


ebranlard (approved):
Thanks for doing that. I think it's important that AeroDyn still works without blades (for future OLAF developments). Thanks for doing that.

andrew-platt:
closes #1781

andrew-platt:
Looking at issue #1781 a bit more closely, this PR may not be completely related.  In the bug report, the issue occurs on line 323 of AeroDyn_Driver_Subs.f90, which is a call to ADI_UpdateStates.  The fix proposed here addresses an issue during an initialization (AeroDyn_Driver_Subs.f90 line244) to ADI_CalcOutput.

So it may be that in diagnosing issue #1781, I found a different issue that is now fixed here. Since I don't see any further issues with this test case, it is possible that it is compiler specific and I don't reproduce it, or that it was fixed in another PR included in rc-3.5.1 to address some uninitialized variables.  As my time is limited, I'll assume the latter until I know otherwise.

modules/aerodyn/src/AeroDyn.f90
```diff
@@ -2848,6 +2848,8 @@ subroutine DiskAvgValues(p, u, m, x_hat_disk, y_hat_disk, z_hat_disk, Azimuth)
    ! calculate disk-averaged velocities
    m%AvgDiskVel = 0.0_ReKi
    m%AvgDiskVelDist = 0.0_ReKi ! TODO potentially get rid of that in the future
+   if (p%NumBlades <= 0) return  ! The Intel compiler gets array bounds issues in this routine with no blades.
+
```
bjonkman, line 2852: Should we make sure that all of the `intent(out)` variables in this routine have been set prior to returning? Otherwise, they may return uninitialized values, depending on compiler settings.

andrew-platt: Good idea.  In this case it is probably ok since none of these values really get used when there are no blades.  Still as a matter of principle it is worth doing.

I'll post another PR setting some of the uninitialized vars.

andrew-platt: I took a quick look in the code at where we need to set things if `p%NumBlades=0`.  This is probably going to take more time than I have at the moment, so I propose doing a PR for such changes on `dev` after the `dev-unstable` merge.

PR #1818, opened 2023-10-19
VS-build updates for rc-3.5.1 and other minor updates

andrew-platt:
This PR is ready to merge.

In preparation for the 3.5.1 release, a few minor issues needed addressing including the following:

- Correct the VS project for `AeroDyn_Inflow_C_Binding` so that it compiles (missing files)
- Add non-OpenMP compile options for `AeroDyn`, `InflowWind`, and `AeroDyn_Inflow_C_Binding`
- Set regression test plots to full width on larger displays -- this had been set to two columns which made the plots unreadably small
- Update the `MoorDyn_Types.f90` file.  Some comments had changed but it had not been regenerated.
- Regenerate the `FAST_Types.f90` file. The `NodeClusterType` was set as an integer in the registry file, but was a logical in the _Types file.

Build systems.

No test results change.


modules/moordyn/src/MoorDyn_Types.f90
RyanDavies19, line 1: @andrew-platt I should have updated the registry when I changed these types but inside MoorDyn the rod roll and pitch are radians rather than degrees.

RyanDavies19: There are a couple changes here to comments I made that are reverted. Mostly corrections to internal MD variable units and meanings. Can I PR to your branch with corrections?

andrew-platt: Sure.  Did you update the MoorDyn_Registry.txt file, or just the MoorDyn_Types.f90 file?  Ideally the MoorDyn_Registry.txt file should be edited and then the registry.exe used to generate the updated MoorDyn_Types.f90 file.

RyanDavies19: I did not update the registry file. I will do that now. Matt had mentioned doing that but it slipped my mind yesterday.

andrew-platt: Thanks for the updates on that!

deslaughter (approved):

Issue #1863, opened 2023-11-08
Seg Fault in HydroDyn in v3.5.1


dzalkind:
**Bug description**
I'm getting a seg fault when running OpenFAST v3.5.1 (via conda installation, both locally and on a linux CI) when running HydroDyn and our IEA-22MW reference model.

**To Reproduce**
<Update the following list with your specific information.>
1. `conda install openfast`
2. clone/download the I[EA-22MW reference model](https://github.com/IEAWindTask37/IEA-22-280-RWT/tree/main/OpenFAST/IEA-22-280-RWT-Monopile)
3. run `openfast IEA-22-280-RWT-Monopile.fst`
4. See the error

**Expected behavior**
A simple test simulation with most modules enabled.

**Screenshots, if applicable**
```
 **************************************************************************************************
 OpenFAST

 Copyright (C) 2023 National Renewable Energy Laboratory
 Copyright (C) 2023 Envision Energy USA LTD

 This program is licensed under Apache License Version 2.0 and comes with ABSOLUTELY NO WARRANTY.
 See the "LICENSE" file distributed with this software for details.
 **************************************************************************************************

 OpenFAST-HEAD-HASH-NOTFOUND
 Compile Info:
  - Compiler: GCC version 12.3.0
  - Architecture: 64 bit
  - Precision: single
  - OpenMP: Yes, number of threads: 16/16
  - Date: Oct 20 2023
  - Time: 23:54:21
 Execution Info:
  - Date: 11/08/2023
  - Time: 13:33:58-0700

 OpenFAST input file heading:
     Generated with AeroElasticSE FAST driver

 Running ElastoDyn.
 Nodal outputs section of ElastoDyn input file not found or improperly formatted.
 Running AeroDyn.
 AD15 Nodal Outputs: Nodal output section of AeroDyn input file not found or improperly formatted.
 Skipping nodal outputs.
 Running InflowWind.
 Running HydroDyn.

Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:
#0  0x10aeaad96
#1  0x10aea9edd
#2  0x7ff8101a35ec
#3  0x109645540
#4  0x10961af00
#5  0x1088163c3
#6  0x10881fd76
#7  0x1086d6a65
#8  0x109e95b5e
#9  0x7ff80fe1c41e
```

**OpenFAST Version**
```
[...]

PR #1864, opened 2023-11-09
Cleanup HydroDyn and SeaState code

bjonkman:
This PR removes duplicate copies of data and pointers of many variables shared between SeaState and HydroDyn. These shared data are now stored in the `WaveField` data type (stored as a SeaState parameter).

It also removes duplicated error checks, outdated code (including outdated driver codes for some HydroDyn submodules), and outdated comments. Hopefully these changes will make it easier to read the code, though more updates could be done. 

Some particular changes to note: 
- Parameter `WtrDpth` has been replaced with `EffWtrDpth` in places where it had added `MSL2SWL` to the variable. `WtrDpth` is now *ALWAYS* the value from the input file. This should help remove confusion about whether that variable is being used before or after adding  `MSL2SWL`.
- Several subroutines were modified so that SeaSt_Interp doesn't give the same warning about clamped boundaries hundreds of times when that error condition occurs.

HydroDyn and SeaState; glue codes that connect the two modules are also modified.

Regression test results have not changed. 

Someone should add a test where `MSL2SWL` is non-zero, though. We should also add tests for the new features that have been added recently.


andrew-platt:
@luwang00, can you also review?

modules/hydrodyn/src/Morison.f90
```diff
@@ -2126,7 +2117,7 @@ SUBROUTINE Morison_Init( InitInp, u, p, x, xd, z, OtherState, y, m, Interval, In
       tMG       = -999.0
       An_drag   = 0.0
       
-      IF ( (InitInp%InpJoints(i)%Position(3)-p%MSL2SWL) >= -p%WtrDpth ) THEN
+      IF ( (InitInp%InpJoints(i)%Position(3)-p%WaveField%MSL2SWL) >= -InitInp%WaveField%EffWtrDpth ) THEN !bjj: ask Lu if this is correct. I wonder if this check is supposed to be against WtrDpth
```
luwang00, line 2120: This should be ok. Based on my notes, `InitInp%InpJoints(i)%Position(3)` is measured from the MSL. Alternatively, we can simplify the check to `InitInp%InpJoints(i)%Position(3) >= -InitInp%WaveField%WtrDpth`.

modules/seastate/src/SeaState_Input.f90
```diff
@@ -121,7 +122,7 @@ subroutine SeaSt_ParseInput( InputFileName, OutRootName, defWtrDens, defWtrDpth,
       if (Failed())  return;
 
       ! Z_Depth - Depth of the domain the Z direction.
-   call ParseVarWDefault ( FileInfo_In, CurLine, 'Z_Depth', InputFileData%Z_Depth, defWtrDpth+InputFileData%MSL2SWL, ErrStat2, ErrMsg2, UnEc )
+   call ParseVarWDefault ( FileInfo_In, CurLine, 'Z_Depth', InputFileData%Z_Depth, defWtrDpth+InputFileData%MSL2SWL, ErrStat2, ErrMsg2, UnEc ) !bjj: wouldn't the default be better with InputFileData%WtrDpth + InputFileData%MSL2SWL since we may have specified a WtrDpth already?
```
luwang00, line 125: @bjonkman I think you are right. We should change the default `Z_Depth` to `InputFileData%WtrDpth + InputFileData%MSL2SWL`.

modules/seastate/src/SeaState_Input.f90
```diff
@@ -743,8 +661,8 @@ subroutine SeaStateInput_ProcessInitData( InitInp, p, InputFileData, ErrStat, Er
 
 
        ! WaveHs - Significant wave height
-
-   if ( ( InputFileData%Waves%WaveMod /= 0 ) .AND. ( InputFileData%Waves%WaveMod /= 4 ) .AND. ( InputFileData%Waves%WaveMod /= 5 ) ) then   ! .TRUE. (when WaveMod = 1, 2, 3, or 10) if we have plane progressive (regular), JONSWAP/Pierson-Moskowitz spectrum (irregular) waves, or white-noise waves, but not user-defined or GH Bladed wave data.
+       !bjj: is this check still appropriate? do we need to add something for WaveMod 6 or 7? Otherwise, fix the comment on the next line
+   if ( ( InputFileData%WaveMod /= WaveMod_None ) .AND. ( InputFileData%WaveMod /= WaveMod_UserSpctrm ) .AND. ( InputFileData%WaveMod /= WaveMod_ExtElev ) ) then   ! .TRUE. (when WaveMod = 1, 2, 3, or 10) if we have plane progressive (regular), JONSWAP/Pierson-Moskowitz spectrum (irregular) waves, or white-noise waves, but not user-defined or GH Bladed wave data.
```
luwang00, line 665: @bjonkman Yes, I think we should add the conditions `InputFileData%WaveMod /= WaveMod_ExtFull` and `InputFileData%WaveMod /= WaveMod_UserFreq` to the `if` statement; however, it would be easier to read if we just switch the `if` and `else` branches.

modules/seastate/src/SeaState_Input.f90
```diff
@@ -759,122 +677,93 @@ subroutine SeaStateInput_ProcessInitData( InitInp, p, InputFileData, ErrStat, Er
...
-         call SetErrStat( ErrID_Info,'WvLowCOff adjusted to '//trim(num2lstr(TmpFreq))//' rad/s, based on WaveDT.',ErrStat,ErrMsg,RoutineName)
+      if ( InputFileData%WvHiCOff > TmpFreq ) then
+         InputFileData%WvHiCOff =  TmpFreq
+         call SetErrStat( ErrID_Info,'WvHiCOff adjusted to '//trim(num2lstr(TmpFreq))//' rad/s, based on WaveDT.',ErrStat,ErrMsg,RoutineName)
       end if
    end if
 
-   if (InputFileData%Waves%WaveMod > 2 .and. InputFileData%Waves%WaveMod /= 6) then
-      if ( InputFileData%Waves%WvLowCOff >= InputFileData%Waves%WvHiCOff ) then
+   !bjj: do we even need this check on WaveMod? Even if it's not being used, we can check that low < Hi, right?
```
luwang00, line 712: @bjonkman Agreed. There's no harm in requiring `WvHiCOff > WvLowCOff` even if they are not used. We can remove the `WaveMod` checks here.

modules/hydrodyn/src/HydroDyn_Output.f90
```diff
@@ -1006,9 +1006,10 @@ SUBROUTINE HDOUT_Init( HydroDyn_ProgDesc, OutRootName, InputFileData, y,  p, m,
    ! Check that the variables in OutList are valid      
    !-------------------------------------------------------------------------------------------------      
       
-   
-   CALL SetOutParam(InputFileData%OutList, p, ErrStat, ErrMsg )
-   IF ( ErrStat >= AbortErrLev ) RETURN
+   if (allocated(InputFileData%OutList)) then
```
luwang00, line 1009: Not sure if this is related to Issue #1863 of Version 3.5.1.

modules/seastate/src/SeaState_Input.f90
ebranlard: As a user I prefer if the sanity checks are only done for the options that are activated. I like to put garbage in the inputs I don't use, to make it visually clear that I'm not using them.

luwang00: If we are to only perform the checks for the relevant `WaveMod` options, I think the ones that should be included are 2 (`WaveMod_JONSWAP`), 3 (`WaveMod_WhiteNoise`), 4 (`WaveMod_UserSpctrm`), 5 (`WaveMod_ExtElev`), and 7 (`WaveMod_UserFreq`) if I'm not mistaken. Not sure why 2 (JONSWAP) was not included in the original check. We might also want to overwrite `WvLowCOff` and `WvHiCOff` for the two regular-wave options in the code here just to make sure we are not cutting out that one wave component.

modules/hydrodyn/src/HydroDyn_Output.f90
deslaughter: This is a good check to perform, just in case OutList hasn't been allocated, but it's not causing the referenced issue.

modules/hydrodyn/src/WAMIT.f90
```diff
@@ -902,13 +901,14 @@ SUBROUTINE WAMIT_Init( InitInp, u, p, x, xd, z, OtherState, y, m, Interval, ErrS
       else
            ! Initialize the variables associated with the incident wave:
 
-         SELECT CASE ( InitInp%WaveMod ) ! Which incident wave kinematics model are we using?
-         CASE ( 0 )  ! No waves, NOTE: for this case we are forcing ExctnDisp = 0, so only p%WaveExctn needs to be allocated, not p%WaveExctnGrid
+         SELECT CASE ( InitInp%WaveField%WaveMod ) ! Which incident wave kinematics model are we using?
```
deslaughter, line 904: Is there a reason that `InitInp%WaveField` isn't `p%WaveField`? It looks like it was changed in most other places.

deslaughter (approved):

modules/hydrodyn/src/HydroDyn_Input.f90
```diff
@@ -1191,42 +1165,24 @@ SUBROUTINE HydroDynInput_ProcessInitData( InitInp, Interval, InputFileData, ErrS
...
+         IF ( ( InitInp%WaveField%WaveStMod /= 0 ) .AND. ( InitInp%WaveField%WaveStMod /= 1 ) .AND. &
+              ( InitInp%WaveField%WaveStMod /= 2 ) .AND. ( InitInp%WaveField%WaveStMod /= 3 ) ) THEN
             ErrMsg  = ' WaveStMod must be 0, 1, 2, or 3.'
             ErrStat = ErrID_Fatal
             RETURN
          END IF
       ELSE
-         IF ( ( InitInp%WaveStMod /= 0 ) .AND. ( InitInp%WaveStMod /= 1 ) .AND. &
-              ( InitInp%WaveStMod /= 3 ) ) THEN
+         IF ( ( InitInp%WaveField%WaveStMod /= 0 ) .AND. ( InitInp%WaveField%WaveStMod /= 1 ) .AND. &
```
bjonkman, line 1177: This check is different than the one in SeaSt_Input.f90/SeaStateInput_ProcessInitData(). In SeaState, it says WaveStMod must be 0 when WaveMod is WaveMod_ExtFull (6). Here in HydroDyn, it says it can be 0, 1, or 3. @luwang00 , do you know what the check should actually be?

modules/seastate/src/SeaState_Input.f90
```diff
@@ -589,129 +580,56 @@ subroutine SeaStateInput_ProcessInitData( InitInp, p, InputFileData, ErrStat, Er
...
+   IF ( InputFileData%WaveMod /= WaveMod_None .AND. InputFileData%WaveMod /= WaveMod_ExtFull ) THEN
+      IF ( (InputFileData%WaveStMod /= 0) .AND. (InputFileData%WaveStMod /= 1) .AND. &
+           (InputFileData%WaveStMod /= 2) .AND. (InputFileData%WaveStMod /= 3) ) THEN
          CALL SetErrStat( ErrID_Fatal,'WaveStMod must be 0, 1, 2, or 3.',ErrStat,ErrMsg,RoutineName)
          RETURN
       END IF
-   ELSE ! Wave stretching is not supported when WaveMod = 0 or 6.
-      InputFileData%Waves%WaveStMod = 0_IntKi
+   ELSE ! Wave stretching is not supported when WaveMod = 0 (WaveMod_None) or 6 (WaveMod_ExtFull).
+      InputFileData%WaveStMod = 0_IntKi
```
bjonkman, line 614: This is not consistent with the check in HydroDyn, where it says WaveStMod can be 0, 1, or 3 for WaveMod_ExtFull.

modules/hydrodyn/src/HydroDyn_Input.f90
luwang00: @bjonkman I believe the check performed in HydroDyn here is correct. Only extrapolation stretching (WaveStMod=2) cannot be used with full-field data. All other wave-stretching methods should work.

modules/seastate/src/SeaState_Input.f90
luwang00: @bjonkman This check in SeaState should be corrected following the check in HydroDyn above. WaveStMod=0, 1, and 3 should be allowed with WaveMod_ExtFull.

modules/hydrodyn/src/Morison.f90
bjonkman: Addressed in 67c77c: I updated the check to be `InitInp%InpJoints(i)%Position(3) >= -InitInp%WaveField%WtrDpth`. I think that's easier to read.

modules/seastate/src/SeaState_Input.f90
bjonkman: Addressed in 67c77c: updated default value for `Z_Depth`

modules/seastate/src/SeaState_Input.f90
bjonkman: Addressed in 67c77c

modules/seastate/src/SeaState_Input.f90
bjonkman: Addressed in 67c77c: The check has been updated for relevant WaveMod options. For WaveMod options that are supposed to ignore these values, I set them to -HUGE() and +HUGE() so that when they ARE used in HydroDyn, they allow all frequencies.

modules/hydrodyn/src/HydroDyn_Output.f90
bjonkman: This was mostly to fix issues in Debug mode where it would error due to passing an invalid array reference.

modules/hydrodyn/src/WAMIT.f90
bjonkman: Addressed in 67c77c: it was just an oversight. I think I had added the parameter p%WaveField after switching a few of the variables.

modules/hydrodyn/src/HydroDyn_Input.f90
bjonkman: Addressed in 67c77c: This check was moved to SeaState, to replace the conflicting check there (and removed from HydroDyn)

modules/seastate/src/SeaState_Input.f90
bjonkman: Addressed in 67c77c: updated with the check from HydroDyn

PR #1882, opened 2023-11-17
Remove IfW data from AD15 inputs

andrew-platt:
This PR is ready to merge.

This is the first of 3 PR's to fix the linearization in AD15 after converting the IfW data handling to pointers.  The three PRs are:

1. remove IfW data from AD15 input, exact match to prior regression tests (this PR)
2. Fix the linearization in AD15 that was broken in previous work and switch IfW inputs to AD15 to "extended inputs".  This should reproduce all linearization results.
3. remove extrapolation of IfW data -- many regression tests change as a result

*PR 1*
The IfW data is removed from the AD15 inputs. To match prior regression tests, the registry is temporarily modified to generate an extrap/interp routine for the new data structure `AD_InflowType`. This allows extrapolating the wind data for BEMT which gives a an exact match to the previous behavior. 

Note: the input positions for the blade motion mesh stored in `u(1)` for the `T + dt` time step are used to calculate the wind.  This yields very slightly different for wind based on extrapolated positions vs. extrapolated wind based on previous time steps.  This causes the `5MW_Land_BD_DLL_WTurb` test case to fail, but the test will pass with a correction step in both the reference and the test case.  See below.

Internal data management of AD15 only.  No other modules affected.

This series of PRs will allow us to merge the `dev-unstable-pointers` branch into `dev`

All regression tests match exactly with the exception of `5MW_Land_BD_DLL_WTurb` test case.  This case has very small differences with the previous results (samples plotted below).  However, this is due to calculating the wind at `T + dt` using the extrapolated positions, rather than extrapolating the wind from previous time steps.  Running this test case with on correction step with both this branch and the `dev-unstable-pointers` branch yields identical results.


andrew-platt:
The `5MW_Land_BD_DLL_WTurb` case failed with the following channels:

- `LSSGagMza`
- `YawBrMzp`
- `TwrBsMzt`
- `B1RootMzr`
- `B2RootMzr`
- `B3RootMzr`
- `B1TipRDzr`
- `B2TipRDzr`
- `B3TipRDzr`

The differences are very small.  For example, for the B2RootMzr: 
Screenshot 2023-11-17 at 5 41 14 PM

Zoomed in:
B2RootMzr18sec

These differences completely vanish if a correction step is used in both the reference and the test case.

deslaughter (approved):
I really like the `RotInflow` type and it will come in handy when integrating MHK

PR #1889, opened 2023-11-23
New spring element in SubDyn: 6 by 6 stiffness matrix

RBergua:
I believe this pull request is ready to be merged.

One new spring element has been included in SubDyn. 
The spring is a new member type. Accordingly, SubDyn now supports as members: beams, cables, rigid elements and springs.
Previously, SubDyn could only include a stiffness matrix as a boundary condition.

The spring element has the next properties:
1. It provides a 6 by 6 stiffness matrix (including diagonal and cross-coupling terms). The stiffness matrix is assumed to be symmetric (Kij = Kji).
image
2. It doesn't have a mass/inertia associated. If interested, the user may include a lumped mass at a given JointID.
3. The spring element must be defined between nodes (i.e., JointIDs) that are coincident. The code checks if this condition is true and aborts the program if the condition is not met.
4. Since the spring element is defined between two coincident nodes, the user must provide the spring orientation by means of the direction cosine.
5. When defining a spring element, the user must provide 22 inputs (the PropSetID and 21 coefficients to define the upper half of the stiffness matrix).
6. The spring is not subdivided by means of NDiv. 
7. The sensors definition for the spring element (e.g., force, moment, displacement, rotation...) are the same as for the beams.
8. The spring can be connected to beams, kinematic joints (e.g., revolute joint, universal joint, and spherical joint), the interface joint and rigid links. Avoid connecting one spring directly to one cable.

The new SubDyn input file includes a new section for the spring element. See below for reference:
image

Internally, the spring element relates 2 nodes with 6 DOFs each. Accordingly, mathematically the spring element has a 12 by 12 dimension.
image

Only SubDyn is impacted. Since this is a new element, the verification tests should pass as the legacy elements in the code have not been modified.

Note that the spring element is defined by joints that are coincident. This challenges the glue-code approach. For example, if HydroDyn is coupled to SubDyn, the hydro loads from HydroDyn will be mapped to the closest structural node in SubDyn. For reference, the link between HydroDyn and SubDyn nodes is performed according to the system definition (i.e., without any deflections taken into account). This means that, unfortunately, the hydro loads around the spring element will be lumped at only one node (i.e., at only one side of the spring element). The node chosen by the glue-code will be the one that appears first during the indexing. The user can minimize this problem by defining an additional structural node at each side of the spring relatively close to it. This will improve the handling of the loads specially if the structural discretization was quite coarse in the first place.    

Having the capability to include a stiffness matrix is very helpful for modelers. This stiffness matrix can be used for example to model:
1. Elastic couplings.
2. Bearings. Some OpenFAST users include part of the wind turbine drivetrain in SubDyn. The loads could be transferred to the supporting structure through one or more spring elements. Tapered roller bearings could also be represented taking advantage of the cross-coupling terms.  
3. Sensors with compliance for models at scale.
4. The spring element could also be used to approximate revolute joints, universal joints or spherical joint. For example, one revolute joint could be considered by defining the diagonal coefficients as stiff (e.g., 1E12, don't provide very high values or the system will be ill-conditioned), except the direction with the DOF of interest (e.g., K44). It's important to remember at this point that SubDyn assumes that structural motions within the body are small.
5. In a land-based system, a user modeling the tower in ElastoDyn could represent the foundation stiffness by adding a single spring element in SubDyn, with the _upper_ joint as the interface and the _lower_ joint as the reaction (effectively having a ground spring). In this setup, the user should define Nmodes = 0 because the Craig–Bampton fixed-interface method will have no internal DOFs. The flexibility is represented through the Guyan modes. To prevent numerical issues, the user should run the simulation with tight coupling in OpenFAST (v5.x or later).
6. Many more!

Several test cases have been reviewed for the new spring element in SubDyn. Below some representative test cases are included. Results compare the output from SubDyn vs reference results (computed analytically or by means of different codes like HOTINT, SAMCEF and OpenModelica).

**Checklist**
- [x] Consider adding a new SubDyn driver r-test (with minimal outputs, few time steps, etc. for limited run time and disk usage, but still testing the new feature)
- [x] Document the new feature in the documentation
- [x] Add the API changes to the documentation
- [x] Update the input files in the r-test
- [x] Final review


RBergua:
**Test Case 1** (https://github.com/OpenFAST/r-test/tree/main/modules/subdyn/SD_Spring_Case1): Spring and lumped mass.
One spring with a vertical stiffness of 3947.8 N/m and a lumped mass of 100 kg is considered. For the spring, one side is connected to the interface joint and the other side is attached to the lumped mass. The gravity acceleration is 9.81 m/s^2. For this condition, the system should oscillate around -0.2485 m (-9.81*100/3947.8) with a frequency of 1 Hz.
image

**Test Case 2** (https://github.com/OpenFAST/r-test/tree/main/modules/subdyn/SD_Spring_Case2): Same spring and lumped mass as test case 1, but 5% damping considered.
image

**Test Case 3**: Comparison between SubDyn revolute joint (https://github.com/OpenFAST/r-test/tree/main/modules/subdyn/SD_Revolute_Joint) and spring element (https://github.com/OpenFAST/r-test/tree/main/modules/subdyn/SD_Spring_Case3).
One pendulum-type system is evaluated. See schematic representation below for reference. One revolute joint or a spring element is defined at point B. The rotational stiffness is 1E6 Nm/rad. The AB beam is 1 m long and the BC beam is 4 m long. Both beams have the same sectional properties (2 m diameter for a full circle beam) and material properties (Young Modulus = 210E9 N/m^2, Poisson ratio = 0.3 and density = 7860 kg/m^3). 
The SubDyn interface joint is located at point A. Accoding to SubDyn’s methodology, it’s a floating case (there are no reaction joints). The Guyan modes correspond to rigid body modes and the gravitational loads are applied to the Guyan modes. This means that the system will only experience the elastic restoring from the revolute joint or the spring element, but not the gravitational restoring.
image

One force to reproduce a free-decay is applied at point C. This nonlinear force is as follows:
#Time_[s] , Fx_[N]  , Fy_[N] , Fz_[N] , Mx_[Nm] , My_[Nm] , Mz_[Nm]
0.0       ,  0.0    , 0.0    , 0.0    , 0.0     , 0.0     , 0.0
25        ,  2.5E4  , 0.0    , 0.0    , 0.0     , 0.0     , 0.0
50        ,  2.5E4  , 0.0    , 0.0    , 0.0     , 0.0     , 0.0
50.01     ,  0.0    , 0.0    , 0.0    , 0.0     , 0.0     , 0.0
100.0     ,  0.0    , 0.0    , 0.0    , 0.0     , 0.0     , 0.0

The angle rotated at the point B is computed by looking at the displacement of point C with regarding to point B.
image  

**Test Case 4**: Beams and springs.
One system with 2 beams is considered (https://github.com/OpenFAST/r-test/tree/main/modules/subdyn/SD_2Beam_Cantilever). See the below schematic representation for reference. The 2 beams use a Timoshenko formulation and have the same dimensions (5 m long each, 1 m external diameter and 0.025 m thickness) and material properties (Young Modulus = 210E9 N/m^2, Poisson ratio = 0.3 and density = 7860 kg/m^3). The node 1 is clamped (interface joint for the SubDyn standalone model).
The first 4 eigenfrequies of the system are: f1 = 9.8 Hz, f2 = 9.8 Hz, f3 = 56.5 Hz and f4 = 56.5 Hz.
Gravity acceleration is not considered and a steady force along the x direction of 2E6 N is applied at node 3.
image

The deflection of the system looks as follows:
image

As it can be observed, the deflection along the x-direction does not involve a deflection in the z-direction duet to the linear assumption used in SubDyn.

This time, one spring element is introduced between the two beams (https://github.com/OpenFAST/r-test/tree/main/modules/subdyn/SD_2Beam_Spring). See the below representation for reference. The spring is defined between nodes 2 and 3 (coincident). The spring provides some compliance in the x-direction (k = 2E7 N/m), the rotation around the y-axis (k = 1E8 Nm/rad) and it’s relatively stiff (e.g., 1E12 N/m or Nm/rad) in the other directions.
image

The new eigenfrequencies for the system are: f1 = 6.3 Hz, f2 = 9.8 Hz, f3 = 27.9 Hz and f4 = 48.5 Hz.
The deflection of the system looks as follows:
image

As expected, the spring experiences a deflection in the x-direction of 0.1 m (2E6N/(2E7N/m)) and a rotation around the y-axis that brings the node 4 at the free-end close to a x-position of 0.95 m.

ebranlard (approved):
I've reviewed the code and it looks good to me. The r-test needs to be updated it seems

ebranlard:
The changes to the SubDyn input files should be updated in the r-test repository. 
And also documented here: 
https://openfast.readthedocs.io/en/dev/source/user/api_change.html

And likely some small documentation changes are needed.

RBergua:
I have updated the documentation for this new spring element. I guess the current r-test will run just modifying the SubDyn input file accordingly. This is, adding below the RIGID LINK PROPERTIES section, the new entries:
```
----------------------- SPRING ELEMENT PROPERTIES -------------------------------------
             0   NSpringPropSets - Number of spring properties
PropSetID   k11     k12     k13     k14     k15     k16     k22     k23     k24     k25     k26     k33     k34     k35     k36     k44      k45      k46      k55      k56      k66    
  (-)      (N/m)   (N/m)   (N/m)  (N/rad) (N/rad) (N/rad)  (N/m)   (N/m)  (N/rad) (N/rad) (N/rad)  (N/m)  (N/rad) (N/rad) (N/rad) (Nm/rad) (Nm/rad) (Nm/rad) (Nm/rad) (Nm/rad) (Nm/rad)
```

modules/subdyn/src/FEM.f90
```diff
@@ -1157,6 +1165,159 @@ SUBROUTINE ElemK_Cable(A, L, E, T0, DirCos, K)
...
+!! The spring element can include diagnal and cross-coupling positions. 
+!! Assuming that the stiffness is symmetric (21 stiffness coefficients). The stiffness matrix could also be non-symmetric, if desired.  
+SUBROUTINE ElemK_Spring(k11, k12, k13, k14, k15, k16, k22, k23, k24, k25, k26, k33, k34, k35, k36, k44, k45, k46, k55, k56, k66, DirCos, K)
+   REAL(ReKi), INTENT( IN) :: k11, k12, k13, k14, k15, k16, k22, k23, k24, k25, k26, k33, k34, k35, k36, k44, k45, k46, k55, k56, k66
+   REAL(FEKi), INTENT( IN) :: DirCos(3,3) !< From element to global: xg = DC.xe,  Kg = DC.Ke.DC^t
+   REAL(FEKi), INTENT(OUT) :: K(12, 12) 
+   ! Local variables
+   REAL(FEKi)                            :: DC(12, 12)
+   
+   K(1:12,1:12) = 0.0_FEKi
```
deslaughter, line 1178: This might be a silly suggestion, but could the indexing be simplified by building the 6x6 K matrix, then applying the element to global rotation DC (6x6) to it, and finally copying the resulting 6x6 into kg (12x12) four times? This method would be equivalent.

modules/subdyn/src/SD_FEM.f90
```diff
@@ -681,14 +687,17 @@ SUBROUTINE SD_Discrt(Init,p, ErrStat, ErrMsg)
        Init%PropsB(1:Init%NPropB, 1:PropSetsBCol) = TempProps(1:Init%NPropB, 1:PropSetsBCol)
     endif
 
-    ! --- Cables and rigid link properties (these cannot be subdivided, so direct copy of inputs)
+    ! --- Cables, rigid link and spring properties (these cannot be subdivided, so direct copy of inputs)
     Init%NPropC = Init%NPropSetsC
     Init%NPropR = Init%NPropSetsR
+	Init%NPropS = Init%NPropSetsS
```
deslaughter, line 693: Please use spaces for indentation (several lines in this file use tabs).

modules/subdyn/src/SubDyn.f90
```diff
@@ -1250,11 +1250,29 @@ SUBROUTINE SD_Input(SDInputFile, Init, p, ErrStat,ErrMsg)
       CALL ReadAry( UnIn, SDInputFile, Init%PropSetsR(I,:), PropSetsRCol, 'RigidPropSets', 'RigidPropSets ID and values ', ErrStat2, ErrMsg2, UnEc ); if(Failed()) return
    ENDDO   
    IF (Check( Init%NPropSetsR < 0, 'NPropSetsRigid must be >=0')) return
+   !----------------------- SPRING ELEMENT PROPERTIES --------------------------------
+   CALL ReadCom  ( UnIn, SDInputFile,                  'Spring element properties'                                 ,ErrStat2, ErrMsg2, UnEc ); if(Failed()) return
+   CALL ReadIVar ( UnIn, SDInputFile, Init%NPropSetsS, 'NPropSetsS', 'Number of spring properties' ,ErrStat2, ErrMsg2, UnEc ); if(Failed()) return
+   CALL ReadCom  ( UnIn, SDInputFile,                  'Spring element properties Header'                          ,ErrStat2, ErrMsg2, UnEc ); if(Failed()) return
+   CALL ReadCom  ( UnIn, SDInputFile,                  'Spring element properties Unit  '                          ,ErrStat2, ErrMsg2, UnEc ); if(Failed()) return
+   IF (Check( Init%NPropSetsS < 0, 'NPropSetsSpring must be >=0')) return
```
deslaughter, line 1258: Should this set an error instead of just returning?

modules/subdyn/src/SubDyn_Tests.f90
```diff
@@ -323,6 +323,7 @@ subroutine Test_Transformations(ErrStat,ErrMsg)
       character(ErrMsgLen), intent(out) :: ErrMsg
 
       real(ReKi), dimension(3) :: P1, P2, e1, e2, e3
+	  integer(IntKi) :: eType
```
deslaughter, line 326: Please use spaces for indentation.

modules/subdyn/src/SD_FEM.f90
```diff
@@ -982,6 +1013,34 @@ SUBROUTINE SetElementProperties(Init, p, ErrStat, ErrMsg)
          p%ElemProps(i)%Rho    = Init%PropsR(P1, 2)
          p%ElemProps(i)%D      = min(sqrt(1/Pi)*4, L*0.05) ! For plotting only
 
+      else if (eType==idMemberSpring) then
+         if (DEV_VERSION) then
+            print*,'Member',I,'is a spring element'
+         endif
+		 p%ElemProps(i)%Area = 0	! Spring elements have no area
```
deslaughter, line 1020: Please use spaces for indentation.

deslaughter (commented):
Found a few minor things. Overall it looks really good.

docs/source/user/subdyn/input_files.rst
```diff
@@ -525,22 +527,26 @@ An example of rigid link properties table is given below
...
-
-
-
 
+Spring Properties
+~~~~~~~~~~~~~~~~
+Members that are specified as spring elements (**MType=5**), 
+have their properties defined in the spring element properties table. 
+The table lists for each spring property: the property ID (**PropSetID**), the diagonal stiffness 
+coefficients (**K11**, **K22**, **K33**, **K44**, **K55**, **K66**), and the cross-coupling
```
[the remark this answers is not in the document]
RBergua: The description could be simplified by stating:
`The table lists for each spring property: the property ID (**PropSetID**) and the stiffness coefficients (**K11**, **K12**, **K13**, **K14**, **K15**, **K16**, **K22**, **K23**, **K24**, **K25**, **K26**,  **K33**, **K34**, **K35**, **K36**, **K44**, **K45**, **K46**, **K55**, **K56**, **K66**).`

Like this it would appear in the same order as in the input file.

docs/source/user/subdyn/theory.rst
```diff
@@ -279,19 +279,21 @@ The following joints are supported:
...
-A member is one of the three following types:
+A member is one of the four following types:
 
 - Beams (*MType=1*), Euler-Bernoulli (*FEMMod=1*) or Timoshenko (*FEMMod=3*)
 
 - Pretension cables (*MType=2*)
 
 - Rigid link (*MType=3*)
 
+- Spring element (*MType=5*)
```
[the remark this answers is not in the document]
RBergua: My bad. MType = 4 is reserved for the arbitrary beams. The spring element is MType = 5. I just fixed this.

modules/subdyn/src/FEM.f90
RBergua: Yes, you could rotate the 6x6 K element matrix and then copy the resulting matrix four times.

modules/subdyn/src/SubDyn.f90
RBergua: I followed the same logic as for the cable (line 1228) to be consistent:
`   IF (Check( Init%NPropSetsC < 0, 'NPropSetsCable must be >=0')) return`

deslaughter: Ah, I missed the call to `Check`, sorry about that.

docs/source/user/api_change.rst
```diff
@@ -9,6 +9,22 @@ The changes are tabulated according to the module input file, line number, and f
...
+SubDyn                                        57\* NSpringPropSets  0                         - Number of spring properties
+SubDyn                                        58\*                                            PropSetID   k11     k12     k13     k14     k15     k16     k22     k23     k24     k25     k26     k33     k34     k35     k36     k44      k45      k46      k55      k56      k66
+SubDyn                                        59\*                                            (-)      (N/m)   (N/m)   (N/m)  (N/rad) (N/rad) (N/rad)  (N/m)   (N/m)  (N/rad) (N/rad) (N/rad)  (N/m)  (N/rad) (N/rad) (N/rad) (Nm/rad) (Nm/rad) (Nm/rad) (Nm/rad) (Nm/rad) (Nm/rad)   
+============================================= ==== =============== ========================================================================================================================================================================================================
+
+\*Exact line number depends on number of entries in various preceeding tables.
+
 OpenFAST v3.5.0 to OpenFAST v3.5.1
```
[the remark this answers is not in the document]
andrew-platt: I made a couple of modifications to the api_change.rst in my branch (will do PR soon to @RBergua's branch).

modules/subdyn/src/SubDyn_Tests.f90
andrew-platt: Fixed in my branch

andrew-platt:
See PR https://github.com/RBergua/openfast/pull/2 for updates (brings this PR up to dev -- only necessary to get r-test to the right point).

verlivkra:
Will the spring elements be included in the next release of OpenFAST? And when is that coming?

RBergua:
The spring elements for SubDyn have been merged into OpenFAST dev branch and will be available in OpenFAST v.4.0.0. Such OpenFAST version should be released relatively soon.

PR #1911, opened 2023-12-01
SubDyn summary file: fixes

RBergua:
I believe this pull request can be merged.

SubDyn can output in the summary file (`SumPrint = 1`) the cosine matrices (`OutCOSM = True`).

In the summary file, the direction cosine for a spring element is not properly written. The summary file looks as:
Wrong

As can be observed, the summary file shows only zeros for the spring element. This is because the `subroutine OutSummary` tries to get the direction cosine based on the joint 1 and joint 2 of the member (and joint 1 and joint 2 are coincident for springs). In this PR, the `subroutine OutSummary` has been modified to properly account for the spring element (and arbitrary beams in the future). 

However, I believe that if the user defines a given direction cosine for a circular beam, the `subroutine GetDirCos` will also compute the direction cosine based on joint 1 and joint 2, effectively overwriting whatever the user provided as input. 

SubDyn summary file.

Proper summary file:
Right

### DIFF
--- a/modules/subdyn/src/SubDyn.f90
+++ b/modules/subdyn/src/SubDyn.f90
@@ -3837,12 +3837,12 @@ SUBROUTINE OutSummary(Init, p, m, InitInput, CBparams, Modes, Omega, Omega_Gy, E
    WRITE(UnSum, '(A9,9(A15))')  '#Member ID', 'DC(1,1)', 'DC(1,2)', 'DC(1,3)', 'DC(2,1)','DC(2,2)','DC(2,3)','DC(3,1)','DC(3,2)','DC(3,3)'
    DO i=1,p%NMembers
       mType = Init%Members(I, iMType)
-	  iNode1 = FINDLOCI(Init%Joints(:,1), Init%Members(i,2)) ! index of joint 1 of member i
+      iNode1 = FINDLOCI(Init%Joints(:,1), Init%Members(i,2)) ! index of joint 1 of member i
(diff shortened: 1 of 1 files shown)


modules/subdyn/src/SubDyn.f90
```diff
@@ -3835,11 +3836,25 @@ SUBROUTINE OutSummary(Init, p, m, InitInput, CBparams, Modes, Omega, Omega_Gy, E
    WRITE(UnSum, '(A, I6)') '#Direction Cosine Matrices for all Members: GLOBAL-2-LOCAL. No. of 3x3 matrices=', p%NMembers 
    WRITE(UnSum, '(A9,9(A15))')  '#Member ID', 'DC(1,1)', 'DC(1,2)', 'DC(1,3)', 'DC(2,1)','DC(2,2)','DC(2,3)','DC(3,1)','DC(3,2)','DC(3,3)'
    DO i=1,p%NMembers
-      iNode1 = FINDLOCI(Init%Joints(:,1), Init%Members(i,2)) ! index of joint 1 of member i
+      mType = Init%Members(I, iMType)
+	  iNode1 = FINDLOCI(Init%Joints(:,1), Init%Members(i,2)) ! index of joint 1 of member i
```
andrew-platt, line 3840: tab character at start of line

andrew-platt (changes requested):
Could you change out the tab characters for spaces?  Otherwise I approve

andrew-platt:
For @andrew-platt:

- [ ] Check that readthedocs renders when this PR is merged (updates to webhook might have broken the intergration)

andrew-platt (approved):

PR #1913, opened 2023-12-03
ADI: probably memory leak in ADI_UpdateStates

andrew-platt:
This is ready to merge.

The ADI_UpdateStates routine makes a copy of the AD_Input, but never destroyed that data after usage.  This could lead to memory continuously getting allocated during a simulation with either ADI_C_Binding or the AD driver.

This issue was reported by e-mail.

- AeroDyn driver
- AeroDyn_Inflow_C_Binding library interface (for C-code coupling)


deslaughter (changes requested):
Destroying the input at the end of `ADI_UpdateStates` gets rid of most of the memory leaks as checked with valgrind. The following needs to be added at line 760 of `fVW.f90` to get rid of the other memory leaks: `call FVW_DestroyInput(uInterp, ErrStat2, ErrMsg2); if(Failed()) return`

deslaughter (approved):

PR #1919, opened 2023-12-05
Use int64 to index arrays for Registry pack/unpack routines

deslaughter:
This PR is ready to be merged

Using int32 (IntKi) for indexing into arrays imposes a limit on the array size of approximately 2 billion elements. Most arrays in OpenFAST will never reach this size, but the `AllOutData` matrix which contains the output data for the entire simulation, and the byte buffer array used for packing and unpacking checkpoint data can exceed this size rather easily, which results in a crash. By using int64 (B8Ki) for indexing, far larger arrays may be used (2^63 elements). Which will eliminate crashing, but may cause OpenFAST to use all of the memory in the computer.

`openfast-registry` and all module `_Types.f90` files including `NWTC_Library_Types.f90` and `ModMesh_Mapping.f90`

No test results were affected.


modules/nwtc-library/CMakeLists.txt
```diff
@@ -14,8 +14,19 @@
 # limitations under the License.
 #
 
+# if (GENERATE_TYPES)
+#   generate_f90_types(src/Registry_NWTC_Library_typedef_nomesh.txt ${CMAKE_CURRENT_LIST_DIR}/src/NWTC_Library_Types.f90 -noextrap)
+# endif()
+ 
 if (GENERATE_TYPES)
-  generate_f90_types(src/Registry_NWTC_Library_typedef_nomesh.txt ${CMAKE_CURRENT_LIST_DIR}/src/NWTC_Library_Types.f90 -noextrap)
+  # Generate Registry_NWTC_Library.txt by concatenating _base.txt and _mesh.txt
```
andrew-platt, line 22: Clever!  I like this solution.

modules/openfast-library/src/FAST_Types.f90
```diff
@@ -6895,17 +6895,17 @@ subroutine FAST_PackOutputFileType(Buf, Indata)
...
    call RegPack(Buf, allocated(InData%TimeData))
    if (allocated(InData%TimeData)) then
-      call RegPackBounds(Buf, 1, lbound(InData%TimeData), ubound(InData%TimeData))
+      call RegPackBounds(Buf, 1, lbound(InData%TimeData, kind=B8Ki), ubound(InData%TimeData, kind=B8Ki))
       call RegPack(Buf, InData%TimeData)
    end if
    call RegPack(Buf, allocated(InData%AllOutData))
    if (allocated(InData%AllOutData)) then
-      call RegPackBounds(Buf, 2, lbound(InData%AllOutData), ubound(InData%AllOutData))
+      call RegPackBounds(Buf, 2, lbound(InData%AllOutData, kind=B8Ki), ubound(InData%AllOutData, kind=B8Ki))
```
andrew-platt, line 6908: This new registry looks a bit different than the old one, but I think this is the spot we had issues in the old one.

andrew-platt (approved):

PR #1932, opened 2023-12-11
Blade resolved loads from CFD

andrew-platt:
This PR is ready to merge.

A new module named `ExtLoads` interfaces with CFD to transfer loads from a CFD solver into OpenFAST.  This module is an alternative to AeroDyn.

To stabilize the CFD solution, AeroDyn loads are used during initialization and then blended into the CFD solution after the transients die out.  The time AD15 loads are used is user configurable (set in radians of rotation -- typically a few revolutions). This requires that AeroDyn is used in addition to the ExtLoads module.

During the simulation, the CFD code will call OpenFAST multiple times for each CFD timestep.  This may require rewinding the OpenFAST simulation to the prior CFD timestep and recalculating all the intervening times using updated CFD information.  To accomplish this, the size of the state variables (and other things) was increased from 2 to 4 so that a saved copy of the previous and current state information (3,4) at the CFD time step can be used.  When rewinding OpenFAST back to the previous CFD time, this data will be copied over to the previous and current state information (1,2).

Other additional features include:
* the ability to rewind multiple timesteps when called through the cpp interface from CFD

* Cpp interface
* glue code
   * additional copies of states, inputs, miscvars to store a checkpoint (copy of previous state and this state at some specific time in history)
   * additional logic for the `ExtLoads` module
* FAST_Library.f90
   * name changes
      * `FAST_ExtInfw_Solution0` renamed `FAST_CFD_Solution0`
      * `FAST_ExtInfw_Step` renamed `FAST_CFD_Step` and split into multiple routines
   * routines with new variables
      * `FAST_ExtInfw_Restart` new variable `numElementsTower_c`
   * new routines
      * `FAST_ExtLoads_Restart`
      * `FAST_CFD_Store_SS`
      * `FAST_CFD_Reset_SS`
      * `FAST_CFD_Prework`
      * `FAST_CFD_UpdateStates`
      * `FAST_CFD_AdvanceToNextTimeStep`
      * `FAST_CFD_WriteOutput`
 * FAST_Subs.f90
    * new routines -- split out from `FAST_Solution`
       * `FAST_Prework`
       * `FAST_UpdateStates`
       * `FAST_AdvanceToNextTimeStep`
       * `FAST_WriteOutput`
    * new routines for substep management
       * `FAST_Store_SS`
       * `FAST_Reset_SS`
       * `FAST_InitIOarrays_SS`
 

**Test results**

**To do**

- [ ] update api_change.rst
- [ ] verify documentation renders correctly


modules/openfast-library/src/FAST_Registry.txt
```diff
@@ -39,23 +40,24 @@ param	FAST	-	INTEGER	Module_Unknown	-	-1	-	"Unknown"	-
...
+param	^	-	INTEGER	Module_HD	-	11	-	"HydroDyn"	-
+param	^	-	INTEGER	Module_SD	-	12	-	"SubDyn"	-
+param	^	-	INTEGER	Module_ExtPtfm	-	13	-	"External Platform Loading MCKF"	-
+param	^	-	INTEGER	Module_MAP	-	14	-	"MAP (Mooring Analysis Program)"	-
+param	^	-	INTEGER	Module_FEAM	-	15	-	"FEAMooring"	-
+param	^	-	INTEGER	Module_MD	-	16	-	"MoorDyn"	-
+param	^	-	INTEGER	Module_Orca	-	17	-	"OrcaFlex integration (HD/Mooring)"	-
+param	^	-	INTEGER	Module_IceF	-	18	-	"IceFloe"	-
+param	^	-	INTEGER	Module_IceD	-	19	-	"IceDyn"	-
+param	^	-	INTEGER	NumModules	-	20	-	"The number of modules available in FAST"	-
```
andrew-platt, line 60: This is incorrect.  Change to `19`

modules/openfast-library/src/FAST_Registry.txt
```diff
@@ -653,7 +700,7 @@ typedef	^	FAST_ModuleMapType	MeshMapType	SubStructure_2_SStC_P_P	{:}	-	-	"Map Su
 # ED --> SrvD -- PlatformPtMesh motion to SrvD%PtfmMotionMesh for passing to DLL
 typedef	^	FAST_ModuleMapType	MeshMapType	ED_P_2_SrvD_P_P	-   	-	-	"Map ElastoDyn     platform point mesh motion to ServoDyn      point mesh -- for passing to controller"
 # ED/BD <-> AD (blades)
-typedef	^	FAST_ModuleMapType	MeshMapType	BDED_L_2_AD_L_B	{:}	-	-	"Map ElastoDyn BladeLn2Mesh point meshes OR BeamDyn BldMotion line2 meshes to AeroDyn14 InputMarkers OR AeroDyn BladeMotion line2 meshes"
+typedef	^	FAST_ModuleMapType	MeshMapType	BDED_L_2_AD_L_B	{:}	-	-	"Map ElastoDyn/BeamDyn BladeLn2Mesh point meshes OR BeamDyn BldMotion line2 meshes to AeroDyn14 InputMarkers OR AeroDyn BladeMotion line2 meshes"
```
andrew-platt, line 703: Revert this line.  text is now incorrect.

modules/openfast-library/src/FAST_Registry.txt
```diff
@@ -404,7 +406,9 @@ typedef	^	^	IceD_InputType	u	{:}	-	-	"System inputs"
 typedef	^	^	IceD_OutputType	y	{:}	-	-	"System outputs"
 typedef	^	^	IceD_MiscVarType	m	{:}	-	-	"Misc/optimization variables"
 typedef	^	^	IceD_InputType	Input	{:}{:}	-	-	"Array of inputs associated with InputTimes"
+typedef	^	^	IceD_InputType	Input_bak	{:}{:}	-	-	"Backup Array of inputs associated with InputTimes"
```
andrew-platt, line 409: instead of `Input_bak`, should we consider naming this `Input_Saved`?

modules/openfast-library/src/FAST_Solver.f90
```diff
@@ -582,6 +654,55 @@ SUBROUTINE AD_InputSolve_IfW( p_FAST, u_AD, y_IfW, y_ExtInfw, ErrStat, ErrMsg )
...
+  ErrMsg = ''
+  
+  pi = acos(-1.0)
+  NumBl  = size(u_AD%rotors(1)%InflowOnBlade,3)
+  Nnodes = size(u_AD%rotors(1)%InflowOnBlade,2)
+
+  do k=1,NumBl
+     do j=1,Nnodes
+        !Get position first
+        z = u_AD%rotors(1)%BladeMotion(k)%Position(3,j) + u_AD%rotors(1)%BladeMotion(k)%TranslationDisp(3,j)
```
andrew-platt, line 684: This will need revision with `dev-unstable-pointers` when the inflow is no longer passed.

modules/openfast-library/src/FAST_Subs.f90
```diff
@@ -5188,6 +5616,2334 @@ SUBROUTINE FAST_InitIOarrays( t_initial, p_FAST, y_FAST, m_FAST, ED, BD, SrvD, A
 
 END SUBROUTINE FAST_InitIOarrays
 !----------------------------------------------------------------------------------------------------------------------------------
+!> Routine that calls FAST_InitIOarrays_SS for one instance of a Turbine data structure. This is a separate subroutine so that the FAST
+!! driver programs do not need to change or operate on the individual module level.
+SUBROUTINE FAST_InitIOarrays_SS_T(t_initial, Turbine, ErrStat, ErrMsg )
```
andrew-platt, line 5621: This name might be a bit confusing since `_SS_` is used elsewhere for steady-state solving.

modules/openfast-library/src/FAST_Subs.f90
```diff
@@ -5188,6 +5616,2334 @@ SUBROUTINE FAST_InitIOarrays( t_initial, p_FAST, y_FAST, m_FAST, ED, BD, SrvD, A
...
+   CHARACTER(*),             INTENT(  OUT) :: ErrMsg              !< Error message if ErrStat /= ErrID_None
+
+      CALL FAST_AdvanceToNextTimeStep(t_initial, n_t_global, Turbine%p_FAST, Turbine%y_FAST, Turbine%m_FAST, &
+                  Turbine%ED, Turbine%BD, Turbine%SrvD, Turbine%AD14, Turbine%AD, Turbine%IfW, Turbine%ExtInfw, &
+                  Turbine%HD, Turbine%SD, Turbine%ExtPtfm, Turbine%MAP, Turbine%FEAM, Turbine%MD, Turbine%Orca, &
+                  Turbine%IceF, Turbine%IceD, Turbine%MeshMapData, ErrStat, ErrMsg )
+
+END SUBROUTINE FAST_AdvanceToNextTimeStep_T
+!----------------------------------------------------------------------------------------------------------------------------------
+!> This routine advances the time step from n_t_global to n_t_global + 1 and does all the relvant copying of data
```
andrew-platt, line 7621: This is redundant with parts of `FAST_Solution`.  Consider simplifying `FAST_Solution` to call this.

modules/openfast-library/src/FAST_Subs.f90
```diff
@@ -5188,6 +5616,2334 @@ SUBROUTINE FAST_InitIOarrays( t_initial, p_FAST, y_FAST, m_FAST, ED, BD, SrvD, A
...
+   CHARACTER(*),             INTENT(  OUT) :: ErrMsg              !< Error message if ErrStat /= ErrID_None
+
+      CALL FAST_UpdateStates(t_initial, n_t_global, Turbine%p_FAST, Turbine%y_FAST, Turbine%m_FAST, &
+                  Turbine%ED, Turbine%BD, Turbine%SrvD, Turbine%AD14, Turbine%AD, Turbine%ExtLd, Turbine%IfW, Turbine%ExtInfw, &
+                  Turbine%HD, Turbine%SD, Turbine%ExtPtfm, Turbine%MAP, Turbine%FEAM, Turbine%MD, Turbine%Orca, &
+                  Turbine%IceF, Turbine%IceD, Turbine%MeshMapData, ErrStat, ErrMsg )
+
+END SUBROUTINE FAST_UpdateStates_T
+!----------------------------------------------------------------------------------------------------------------------------------
+!> This routine takes data from n_t_global and predicts the states and output  at n_t_global+1
```
andrew-platt, line 7501: This is redundant with parts of `FAST_Solution`.  Consider simplifying `FAST_Solution` to call this.

modules/beamdyn/src/Registry_BeamDyn.txt
```diff
@@ -44,6 +44,7 @@ typedef   ^        InitOutputType CHARACTER(ChanLen)  WriteOutputUnt {:}  -   -
 typedef   ^        InitOutputType ProgDesc            Ver             -   -   -   "This module's name, version, and date"  -
 typedef   ^        InitOutputType R8Ki                kp_coordinate  {:}{:} - -   "Key point coordinates array"            -
 typedef   ^        InitOutputType IntKi               kp_total        -   -   -   "Total number of key points" -
+typedef   ^        InitOutputType R8Ki                QPtN           {:}   -   -   "Quadrature (QuadPt) point locations in natural frame [-1, 1]" -
```
bjonkman, line 47: Is there a reason that a part of the code outside of the BeamDyn module would need the BD quadrature points?  Also wondering why `kp_coordinates` are InitOutputType (I realize that is an old issue and not part of this PR.)

andrew-platt: I think this was mistakenly kept during the merge. It shouldn't be there as it isn't used anywhere else.  Thanks for catching this!

modules/beamdyn/src/Registry_BeamDyn.txt
```diff
@@ -44,7 +44,6 @@ typedef   ^        InitOutputType CHARACTER(ChanLen)  WriteOutputUnt {:}  -   -
 typedef   ^        InitOutputType ProgDesc            Ver             -   -   -   "This module's name, version, and date"  -
 typedef   ^        InitOutputType R8Ki                kp_coordinate  {:}{:} - -   "Key point coordinates array"            -
```
bjonkman, line 45: can `kp_coordinate` and `kp_total` be removed here, too?

andrew-platt: Done!

modules/beamdyn/src/Registry_BeamDyn.txt
andrew-platt: Removed!

andrew-platt:
There are discrepancies in some models run with the cpp driver and the OpenFAST Fortran executable.  So we will need to fix whatever the underlying issue is before merging this.

Marking this as draft until we resolve the problem.

cmake/FindNetCDF.cmake
deslaughter, line 1: It might be good to update this from https://github.com/Kitware/VTK/blob/master/CMake/FindNetCDF.cmake since it seems to have changed a little. The biggest improvement is that it provides `NetCDF::NetCDF`: A target to use with `target_link_libraries`.

glue-codes/openfast-cpp/src/OpenFAST.H
deslaughter, line 1: If this is too difficult, you can skip it, but it would probably be better to name this file `OpenFAST.hpp` in case we create an `OpenFAST.h` as a C header file. On Windows these two files would be in conflict when installed because the filesystem is not case sensitive.

modules/openfast-library/src/FAST_Registry.txt
```diff
@@ -420,76 +424,101 @@ typedef	^	^	BD_MiscVarType	m	{:}	-	-	"Misc/optimization variables"
...
+typedef	^	^	BD_InputType	Input_Saved	{:}{:}	-	-	"Backup Array of inputs associated with InputTimes"
 typedef	^	^	DbKi	InputTimes	{:}{:}	-	-	"Array of times associated with Input Array"
+typedef	^	^	DbKi	InputTimes_Saved	{:}{:}	-	-	"Backup Array of times associated with Input Array"
 
 # ..... ElastoDyn data .......................................................................................................
-typedef	FAST	ElastoDyn_Data	ED_ContinuousStateType	x	{2}	-	-	"Continuous states"
-typedef	^	^	ED_DiscreteStateType	xd	{2}	-	-	"Discrete states"
-typedef	^	^	ED_ConstraintStateType	z	{2}	-	-	"Constraint states"
-typedef	^	^	ED_OtherStateType	OtherSt	{2}	-	-	"Other states"
+typedef	FAST	ElastoDyn_Data	ED_ContinuousStateType	x	{4}	-	-	"Continuous states"
```
deslaughter, line 432: I really wish we had defined a NumStates parameter, so the state arrays would be the same size across modules. Something to remember for the future.

deslaughter (approved):

modules/openfast-library/src/FAST_Registry.txt
andrew-platt: That's worth doing. I'll make the change.

glue-codes/openfast-cpp/src/OpenFAST.H
andrew-platt: Let's leave this for now.  It might annoy SOWFA and Nalu-Wind people.

modules/openfast-library/src/FAST_Registry.txt
andrew-platt: `NumStateTimes`

cmake/FindNetCDF.cmake
andrew-platt: Since we don't currently have any issues with cmake or spack builds, let's leave it for now.

andrew-platt:
Screen Shot 2023-12-14 at 4 11 31 PM

Simulation from NALU-Wind coupled to OpenFAST through the `ExtLoads` module.

modules/openfast-library/src/FAST_Registry.txt
```diff
@@ -39,23 +40,24 @@ param	FAST	-	INTEGER	Module_Unknown	-	-1	-	"Unknown"	-
...
-param	^	-	INTEGER	Module_SD	-	11	-	"SubDyn"	-
-param	^	-	INTEGER	Module_ExtPtfm	-	12	-	"External Platform Loading MCKF"	-
-param	^	-	INTEGER	Module_MAP	-	13	-	"MAP (Mooring Analysis Program)"	-
-param	^	-	INTEGER	Module_FEAM	-	14	-	"FEAMooring"	-
-param	^	-	INTEGER	Module_MD	-	15	-	"MoorDyn"	-
-param	^	-	INTEGER	Module_Orca	-	16	-	"OrcaFlex integration (HD/Mooring)"	-
-param	^	-	INTEGER	Module_IceF	-	17	-	"IceFloe"	-
-param	^	-	INTEGER	Module_IceD	-	18	-	"IceDyn"	-
-param	^	-	INTEGER	NumModules	-	18	-	"The number of modules available in FAST"	-
+param	^	-	INTEGER	Module_ExtLd	-	8	-	"AeroDyn"	-
```
bjonkman, line 48: This module name should be fixed.

modules/openfast-library/src/FAST_Subs.f90
```diff
@@ -5561,11 +8017,77 @@ SUBROUTINE FAST_Solution(t_initial, n_t_global, p_FAST, y_FAST, m_FAST, ED, BD,
...
+   ErrStat = ErrID_None
+   ErrMsg  = ""
+
 
    !----------------------------------------------------------------------------------------
    !! Check to see if we should output data this time step:
-   !----------------------------------------------------------------------------------------   
-   CALL WriteOutputToFile(n_t_global_next, t_global_next, p_FAST, y_FAST, ED, BD, AD14, AD, IfW, ExtInfw, SeaSt, HD, SD, ExtPtfm, &
+   !----------------------------------------------------------------------------------------
+   CALL WriteOutputToFile(n_t_global, m_FAST%t_global, p_FAST, y_FAST, ED, BD, AD14, AD, IfW, ExtInfw, SeaSt, HD, SD, ExtPtfm, &
```
bjonkman, line 8090: Not sure why you are passing variable `t_global` when using `m_FAST%t_global` in this routine. The call from `FAST_WriteOutput_T` uses `t_initial`, so that is also confusing.

modules/openfast-library/src/FAST_Subs.f90
```diff
@@ -27,7 +27,7 @@ MODULE FAST_Subs
    USE VersionInfo
 
    IMPLICIT NONE
-   
+
```
bjonkman, line 30: For future pull requests, it would be nice if we didn't remove all the trailing spaces from existing source files. It creates a lot of unnecessary changes that can cause merge conflicts.

modules/openfast-library/src/FAST_Subs.f90
andrew-platt: There was some confusion on this during the merge.  I'll get this fixed.

modules/openfast-library/src/FAST_Subs.f90
andrew-platt: Agreed.  That snuck in during one of the later PRs and I didn't catch it in time.

andrew-platt: For future comparisons, there is a `hide whitespace` option in the file compare in a GitHub PR.  Unfortunately, it hides all whitespace changes, not just the trailing ones.

bjonkman: Yeah, I can _view_ these changes in the PR without the white spaces fine. I've just had experience with merge conflicts getting worse due to one file removing white space while another changes something else on the line. It's not _hard_ to fix those conflicts, just potentially time consuming (some merge-conflict editors make this easier than others).

andrew-platt: I agree.  I've also found it can really make resolving merge conflicts much more time consuming.  I'll pay more attention to that going forward.

modules/openfast-library/src/FAST_Solver.f90
```diff
@@ -235,7 +278,23 @@ SUBROUTINE ED_InputSolve( p_FAST, u_ED, y_ED, p_AD14, y_AD14, y_AD, y_SrvD, u_AD
             CALL Transfer_Line2_to_Point( y_AD%rotors(1)%BladeLoad(k), u_ED%BladePtLoads(k), MeshMapData%AD_L_2_BDED_B(k), ErrStat2, ErrMsg2, u_AD%rotors(1)%BladeMotion(k), y_ED%BladeLn2Mesh(k) )
                CALL SetErrStat(ErrStat2, ErrMsg2, ErrStat, ErrMsg, RoutineName)
          END DO
-                  
+      
+      ELSE IF (p_FAST%CompAero == Module_ExtLd ) THEN
+
+         DO K = 1,SIZE(u_ED%BladePtLoads,1) ! Loop through all blades (p_ED%NumBl)
```
bjonkman, line 284: This code to blend AD and ExtLd data is repeated in both BeamDyn and ElastoDyn routines, and actually called twice in ED because `ExtLd_ConvertOpDataForOpenFAST` is called for both tower and blade loads. Since the code is just adding AeroDyn output data to the ExtLd output mesh, it seems like that should have been done in a different routine so that these routines to set BeamDyn and ElastoDyn inputs don't actually change the ExtLd outputs. That would mean you don't need the extra p, u, and m variables from ExtLd, making it match the way the other modules are handled here. Otherwise, that blending should be stored someplace other than the ExtLd output type.

modules/extloads/src/ExtLoadsDX_Registry.txt
```diff
@@ -0,0 +1,44 @@
...
+typedef ^       InputType      R8Ki            nacDef        {:}     -       -       "Deformations the nacelle - to send to external driver"
+typedef ^       InputType      R8Ki            bldRootDef    {:}     -       -       "Deformations of the blade root nodes - to send to external driver"
+typedef ^       InputType      R8Ki            twrRefPos     {:}     -       -       "Reference position of the tower nodes - to send to external driver"
+typedef ^       InputType      R8Ki            bldRefPos     {:}     -       -       "Reference position of the all blade nodes - to send to external driver"
+typedef ^       InputType      R8Ki            hubRefPos     {:}     -       -       "Reference position of the tower nodes - to send to external driver"
+typedef ^       InputType      R8Ki            nacRefPos     {:}     -       -       "Reference position of the all blade nodes - to send to external driver"
+typedef ^       InputType      R8Ki            bldRootRefPos {:}     -       -       "Reference position of the blade root nodes - to send to external driver"
+typedef ^       InputType      IntKi           nBlades       {:}     -       -       "Number of blades"
```
bjonkman, line 31: These integers (and possibly some other R8Ki variables below) seem like they should be parameters of the module, not inputs.

andrew-platt: I agree!

@gantech, let's figure out how to resolve this.

modules/openfast-library/src/FAST_Solver.f90
andrew-platt: @gantech, FYI.

modules/extloads/src/ExtLoadsDX_Registry.txt
andrew-platt: Maybe a better solution: PR #2001

modules/openfast-library/src/FAST_Subs.f90
```diff
@@ -5153,61 +5581,2048 @@ SUBROUTINE FAST_InitIOarrays( t_initial, p_FAST, y_FAST, m_FAST, ED, BD, SrvD, A
...
+   CHARACTER(*), PARAMETER                 :: RoutineName = 'FAST_Solution'
+
+
+   ErrStat = ErrID_None
+   ErrMsg  = ""
+
+   n_t_global_next = n_t_global+1
+   t_global_next = t_initial + n_t_global_next*p_FAST%DT  ! = m_FAST%t_global + p_FAST%dt
+
+   y_FAST%WriteThisStep = NeedWriteOutput(n_t_global_next, t_global_next, p_FAST)
```
bjonkman, line 7464: I'm wondering why this line wasn't included in `FAST_Prework`. 

`y_FAST%WriteThisStep` is used in `FAST_UpdateStates`. While it doesn't matter in this glue code, I assume the reason this routine was split into `FAST_Prework`, `FAST_UpdateStates`, etc, is that the individual routines would be called outside `FAST_Solution`, in which case, `y_FAST%WriteThisStep` would probably not be set properly. 

Is there another reason for keeping it here?

andrew-platt: Good question.  Since the `AMR-Wind` has the ability to reset OpenFAST back multiple timesteps, it may be that they didn't want to have the write outputs updated during intermediate steps that may be redone. @gantech, do you happen to remember?

bjonkman: If they aren't directly setting `y_FAST%WriteThisStep`, then it is probably just using the initial value of `true`. If you don't need file output at every integration step, you will get performance improvements by setting this value.

andrew-platt: Corrected with #2076 (on the `dev-unstable-pointers` branch which will be merged to `dev` soonish).

PR #1949, opened 2023-12-19
Release 3.5.2

andrew-platt:
**Feature or improvement description**
Pull request to merge `rc-3.5.2` into `main` and create a tagged release for v3.5.2.

See the milestone and project pages for additional information

    https://github.com/OpenFAST/openfast/milestone/12

Test results, if applicable
See GitHub Actions

### Release checklist:
- [x] Update the documentation version in docs/conf.py
- [x] Update the versions in docs/source/user/api_change.rst
- [x] Verify readthedocs builds correctly
- [ ] Create a tag in OpenFAST
- [ ] Create a merge commit in r-test and add a corresponding tag
- [ ] Compile executables for Windows builds
    - [ ] FAST_SFunc.mexw64
    - [ ] OpenFAST-Simulink_x64.dll
    - [ ] openfast_x64.exe
    - [ ] DISCON.dll (x64)
    - [ ] AeroDyn_Driver
    - [ ] AeroDyn_Inflow_C_Binding
    - [ ] BeamDyn_Driver
    - [ ] HydroDyn_Driver
    - [ ] HydroDyn_C_Binding (x64)
    - [ ] InflowWind_Driver
    - [ ] IfW_C_Binding (x64)
    - [ ] MoorDyn_Driver
    - [ ] FAST.Farm (x64)

# Changelog

### Build systems

#1948 Pass Python_EXECUTABLE to pfunit, add error check on Python version

### AeroDyn

#1913 ADI: memory leak in ADI_UpdateStates

### AWAE

#1963 FAST.Farm, Mod_AmbWind=3: add error if HR grid not centered on turbine in Y dimension

### HydroDyn

#1872 Fix segfault in HD when no outputs specified

## Regression tests

#1886  Update floating MHK case input files

## Input file changes

No input files change with this release, as this only includes minor bugfixes.

Full list of changes: https://openfast.readthedocs.io/en/main/source/user/api_change.html

Full input file sets: https://github.com/OpenFAST/r-test/tree/v3.5.2 (example input files from the regression testing)

PR #1950, opened 2023-12-20
Option to turn off unit_tests from cmake

andrew-platt:
This PR is ready to merge.

Previously anytime the `BUILD_TESTING` option in CMake was turned on, unit tests would also be run.  However, Kestrel cannot currently run the unit testing, so an option to turn that off separately is needed.

A new CMake variable `BUILD_UNIT_TESTING` is now turned on if `BUILD_TESTING` is on.  This option can be separately specified for scenarios where pfunit used in testing is not available.

#1948

Unit testing.

We are currently using `pfunit` version 3.2.10, which is 6+ years old (current version is 4.8).  We don't have resources at present to deploy the newer `pfunit` or fix the issues on Kestrel, so a workaround is needed for these rare cases.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,7 +67,10 @@ endif()
 option(BUILD_TESTING "Build the testing tree." OFF)
 if(BUILD_TESTING)
   option(CODECOVERAGE "Enable infrastructure for measuring code coverage." OFF)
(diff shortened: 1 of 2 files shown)


andrew-platt:
Update PR #1949 notes and the `v3.5.2.md` file with links to this PR after merging.

deslaughter (approved):

PR #1951, opened 2023-12-20
[BugFix] FF mod_AmbWind=3 failed if a turbine not aligned exactly at center of high res grid

andrew-platt:
This is ready to merge.

**Description**

If a turbine was shifted ever so slightly in the +/-Y direction from a high resolution grid point, the box of grid points sent to InflowWind for calculation would be shifted so one edge was just outside the wind box.  This would immediately fail.

For example, the FF input file:

```
--- WIND TURBINES ---
1             NumTurbines   - Number of wind turbines (-) [>=1] [last 6 columns below used only for Mod_AmbWind=2 or 3]
WT_X            WT_Y            WT_Z            WT_FASTInFile   X0_High         Y0_High         Z0_High         dX_High         dY_High         dZ_High
(m)             (m)             (m)             (string)        (m)             (m)             (m)             (m)             (m)             (m)
1.20200000e+02	-315.1	0	"FFTest_WT1.fst"	-1.00300000e+02	-540	1	4	5	5
```

puts the turbine at -315.1 m in Y, which is 0.1 m from the center of the high resolution grid located at -315 m (`Y0_high + 0.5*dY_high*(nY_high-1)`).  During  the call to InflowWind in AWAE, the high res grid points were shifted by the turbine XY position instead of the center of the high res grid.  This resulted in all grid points passed to InflowWind being shifted by -0.1 m too far in Y, so one side of passed grid was outside the wind box.

Note that the code fix also shifts the box differently in the X direction.  This changed results for the one regression test.

Also discovered that the hub center was not correctly calculated.  This actually doesn't matter since that is used in calculating the disk average values, which are ignored entirely in AWAE.

@abhineet-gupta figrued out where the problem likely was.

https://forums.nrel.gov/t/gf-wind-array-boundaries-violated-grid-too-small-in-y-direction/6149/5

FAST.Farm with `Mod_AmbWind=3`

Updated the ModAmb_3 test case to verify this fixes the problem. Properly setting the offset to the center of the high resolution box also shifted the box 20 m in the X direction for the regression test, so the results changed very slightly.

### DIFF
--- a/modules/awae/src/AWAE.f90
+++ b/modules/awae/src/AWAE.f90
@@ -1361,6 +1361,7 @@ subroutine AWAE_UpdateStates( t, n, u, p, x, xd, z, OtherState, m, errStat, errM
    character(*), parameter                      :: RoutineName = 'AWAE_UpdateStates'
 !   real(DbKi)          :: t1, t2
    integer(IntKi)                               :: n_high_low, nt, n_hl, i,j,k,c
+   real(ReKi)                                   :: offsetBoxCtr(3)      ! for shifting high res boxes before sending coordinates to InflowWind
    
    errStat = ErrID_None
    errMsg  = ""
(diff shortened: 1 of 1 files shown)


andrew-platt:
@abhineet-gupta, please take a look at this and verify it works correctly.

abhineet-gupta:
This works. I can successfully run my simulation now.

jjonkman (changes requested):
I'm not sure I understand this change.  With `Mod_AmbWind` = 3, a separate instance of InflowWind is used for the low-resolution domain and the high-resolution domain associated with each turbine.  InflowWind always assumes that its wind domain is defined relative to its local (0,0,0) origin, which for the low-resolution domain is the FAST.Farm origin and for each high-resolution domain is the origin of the tower base of a turbine.  This is why the original code subtracted `p%WT_Position` from `p%Grid_high`.  The new code assumes this origin is the lateral center of high-res domain, not the wind turbine origin, which is not compatible with the assumptions in InflowWind.

My guess is the issue reported in the forum post is related to an issue in the FAST.Farm case creation python scripts rather than an issue with the FAST.Farm source code.

andrew-platt:
A few details:

- When `ModAmbWind=3`, the position of the high-resolution domain (corner) is set by the coordinates given in the FAST.Farm input file: `X0_High` etc, and the discretization `dX_High` etc.
- In 3.4.1, it was possible to shift the position of the turbine by a small amount relative to the box using the `WT_X` etc. coordinates.  
- The wind turbine reference position is hard-coded in InflowWind as `(0,0,RefHt)`.  This was the case in 3.4.1 as well.

However the behavior changed. In 3.4.1 it was possible to have the turbine misaligned slightly with the center of the box, but this is no longer possible. For example, shifting a turbine from `WT_Y=315m` to `WT_Y=315.2m` was perfectly acceptable.  What exactly changed is a bit of a mystery: reference position handling within InflowWind and AWAE are identical between 3.4.1 and present dev.

One other potentially useful detail is about a bug that was fixed during restructuring InflowWind for v3.5.0.  Before this, the interpolation allowed points just outside the grid in one direction.  When we added the ability to extrapolate beyond the box boundary (only allowed with OLAF and Lidar), we discovered that the interpolation algorithm was not treating the grid boundaries correctly and would allow interpolation of points within one `dy` outside the grid boundary.  Now the grid boundaries are more strictly enforced (perhaps too rigidly).  So it may be that 3.4.1 allowed small shifts in turbine positions only because this interpolation bug (this was discovered during #1516 when extrapolation beyond the box boundary was implemented https://openfast.readthedocs.io/en/dev/source/user/inflowwind/driver.html#boxexceedallow-flag).

bjonkman:
> the interpolation algorithm ... would allow interpolation of points within one `dy` outside the grid boundary

I have a vague recollection of adding this "feature" when I was at NREL. I believe there was a request to allow extrapolation of the wind field outside the box, but I wasn't comfortable with extrapolating too far, so we just added the one point. I think someone had created grids that were just a tiny bit too small for their project and this fixed their issue.

andrew-platt:
Thanks for the info @bjonkman!  I couldn't imagine this had been a true bug with the box boundaries.  I think this is a feature we want to add back in -- maybe only a half grid delta at each edge in Y.


[FF_highResGrid_original_offset.pdf](https://github.com/OpenFAST/openfast/files/13863571/FF_highResGrid_original_offset.pdf)
FAST.Farm high res grid and OpenFAST wind grid (`Mod_AmbWind=3`) in v3.4.1 and before. This is the case where the grid sizing and spacing for the IfW grid and high-res grid are identical.  If the turbine is slightly shifted from the center of the high-res grid, some of the points end up outside the box, which wasn't a problem when we had the ability to sample just outside the grid.


[FF_highResGrid_new_offset.pdf](https://github.com/OpenFAST/openfast/files/13863666/FF_highResGrid_new_offset.pdf)
With v3.5.0, I tightened up the boundaries on the grid, so now a slightly offset turbine will have high-res points outside the grid.  I think we'll need to allow a bit larger boundary for the interpolation (half dY).

andrew-platt:
We decided that the best solution for `Mod_AmbWind=3` is to enforce that the high-res grid center is aligned with the _InflowWind_ grid.  Closing this PR in favor of a different approach.

Issue #1849, opened 2023-11-02
MoorDyn v2: echo file missing


RBergua:
**Bug description**
It seems that MoorDyn v2 is not producing an Echo file despite the flag being defined as True.  

**To Reproduce**
Any model from the r-test repository (https://github.com/OpenFAST/r-test/tree/main/glue-codes/openfast) that uses MoorDyn. 

**Expected behavior**
An output Echo file (*.ech).

**OpenFAST Version**
This issue has been observed in OpenFAST v3.5.1 and OpenFAST dev.

andrew-platt:
@RyanDavies19 and @mattEhall, can you address this?

RyanDavies19:
@RBergua the echo file capability was removed when MoorDyn was upgraded to v2 (#1086). Good point noticing that the flags are still in place. I will work on removing them with some other work I am doing.

RyanDavies19:
@andrew-platt MoorDyn creates a log summary file that includes most all of the inputs. There are a couple things that still could be added though.

RBergua:
@RyanDavies19 Is there a specific flag to enable this log summary? I'm only getting result files (*.out) from MoorDyn at the moment.

andrew-platt:
If the log file only gets written when MoorDyn successfully initializes, then we will need to add the echo back in.  We will need some way to troubleshoot the input files in that case.

RyanDavies19:
The flag is '1       writeLog' in the options section and a higher integer triggers more detailed output (at least in the C code it does). I talked with Matt about this today and am planning to finish getting all the details outputted to the log after I get pinned bodies completed.

RyanDavies19:
@RBergua this can be closed, my PR removed echo file flags from things and updated the log file to include everything from the inputs

PR #1967, opened 2024-01-09
MoorDyn: Coupled Pinned Bodies and bug fixes

RyanDavies19:
This update adds in coupled pinned bodies to MoorDyn. These objects are free in their rotational degrees of freedom while coupled in their translational degrees of freedom. The intent is to allow for a bundle of rods attached to a body to rotate about a single coupled point. In the same way coupled pinned rods are set up, the glue code still sees the object as having a full 6 DoF however moments of zero are returned by the object and the rotational kinematics inputted into MoorDyn by the glue code are ignored. 

A flag to include/ignore inertial components of the forces passed to the glue code has been added as well. The flag belongs in the options list and is called “inertialF”. A value of 1 indicated include inertial components, a value of 0 indicated do not include inertial components. The default value is 0, which preserves the functionality of OpenFAST prior to this update. This flag is included because the inertial forces can cause instability due to transients in the accelerations. This issue should be resolved in OpenFAST v5 with the introduction of the tight coupling. The MoorDyn documentation will be updated accordingly shortly. 

This update also removes the deprecated echo file flag and presents an updated log file as it’s replacement (the log is an existing feature, just updated now to include more information). The requires a change to the MoorDyn driver input file format to remove the echo file flag from there. The general MoorDyn input file will ignore if an echo file is called for.

The option to run MoorDyn as a standalone code with the driver has been added, when NumTurbines = -1. This ignores all tower coordinate systems and can be useful when running a system with multiple free bodies using MD as a standalone code. 

Echo file removal and updated log file #1849 

MoorDyn

Coupled pinned bodies were added to support the USFLOWT project’s use of buoyancy cans. Through this development process various bug were discovered and resolved including the addition of the toggle for the inclusion of inertial forces. 

Opening this PR because r-tests are not running on my computer correctly. Using GitHub actions to run tests.

@RBergua has verified this new feature with analytical solutions and comparison to other codes.

- [ ] Update the documentation version in docs/conf.py
- [ ] Update the versions in docs/source/user/api_change.rst
- [ ] Verify readthedocs builds correctly
- [ ] Create a tag in OpenFAST
- [ ] Create a merge commit in r-test and add a corresponding tag
- [ ] Compile executables for Windows builds
    - [ ] FAST_SFunc.mexw64
    - [ ] OpenFAST-Simulink_x64.dll
    - [ ] openfast_x64.exe
    - [ ] DISCON.dll
-->


andrew-platt:
@mattEhall, do you want to take a quick look at this?

mattEhall:
Generally looks good to me. These are great additions. @RyanDavies19, what's the reason for moving some log file outputs from other modules into the init function?

modules/moordyn/src/MoorDyn_Misc.f90
```diff
@@ -1707,49 +1712,58 @@ SUBROUTINE gridAxisCoords(coordtype, entries, coordarray, n, ErrStat, ErrMsg)
...
-         end if
-         
-         print *, "Set water grid coordinates to :"
-         DO i=1,n
-            print *, " ", coordarray(i)
-         end do
-         
+
+         IF (len(trim(entries)) == len(entries)) THEN
+            print*, "Warning: Only 120 characters read from wave grid coordinates"
```
andrew-platt, line 1717: We generally discourage the use of `print*` statements as they do not work when coupled to Simulink.  Instead, use `call WrScr("text to print")`.

modules/moordyn/src/MoorDyn.f90
```diff
@@ -3885,27 +4068,39 @@ SUBROUTINE Init_Jacobian_x()
...
-         InitOut%LinNames_x(idx+6) = 'Body '//trim(num2lstr(m%FreeBodyIs(l)))//' omega_z, rad/s'
-         p%dxIdx_map2_xStateIdx(idx+1) = m%BodyStateIs1(l)+0         ! x%state index for Rx
-         p%dxIdx_map2_xStateIdx(idx+2) = m%BodyStateIs1(l)+1         ! x%state index for Ry
-         p%dxIdx_map2_xStateIdx(idx+3) = m%BodyStateIs1(l)+2         ! x%state index for Rz
-         p%dxIdx_map2_xStateIdx(idx+4) = m%BodyStateIs1(l)+3         ! x%state index for omega_x
-         p%dxIdx_map2_xStateIdx(idx+5) = m%BodyStateIs1(l)+4         ! x%state index for omega_y
-         p%dxIdx_map2_xStateIdx(idx+6) = m%BodyStateIs1(l)+5         ! x%state index for omega_z
-         idx = idx + 6
+         if (m%BodyList(m%FreeBodyIs(l))%typeNum == 2) then ! Coupled pinned body
+            ! corresponds to state indices: (m%BodyStateIs1(l):m%BodyStateIs1(l)+5)
```
andrew-platt, line 4072: This may break linearization.  Did you resize the `p%dx` `InitOut%LinNames_x` accordingly?

We don't have a linearization case involving bodies though, so I don't have a good way to check.

andrew-platt (changes requested):
Requesting changes to the use of `print*` statements.

I also need to think about the linearization impacts of the `p%dx` changes.

modules/moordyn/src/MoorDyn_Registry.txt
```diff
@@ -24,7 +24,8 @@ typedef   MoorDyn/MD  InitInputType    ReKi               g              -   -99
 typedef   ^           ^                ReKi               rhoW           -   -999.9 -   "sea density"        "[kg/m^3]"
 typedef   ^           ^                ReKi               WtrDepth       -   -999.9 -   "depth of water"     "[m]"
 typedef   ^           ^                ReKi               PtfmInit      {:}{:} -    -   "initial position of platform(s) shape: 6, nTurbines"     -
-typedef   ^           ^                IntKi              FarmSize       -    0     -   "Indicates normal FAST module mode if 0, FAST.Farm coupled mode and =nTurbines if >0"     -
+typedef   ^           ^                IntKi              FarmSize       -    0     -   "Indicates normal FAST module mode if 0, FAST.Farm coupled mode and =nTurbines if >0, standalone mode if -1"     -
+typedef   ^           ^                IntKi              Standalone     -    0     -   "Indicates MoorDyn run as standalone code if 1, coupled if 0"     -
```
bjonkman, line 28: Seems like the value of `FarmSize` is sufficient to determine `Standalone`, so this seems redundant.

modules/moordyn/src/MoorDyn.f90
```diff
@@ -1763,20 +1882,20 @@ SUBROUTINE MD_Init(InitInp, u, p, x, xd, z, other, y, m, DTcoupling, InitOut, Er
...
-            m%BodyList(m%CpldBodyIs(l,iTurb))%r6(4:6) = EulerExtract(MATMUL(OrMat, TRANSPOSE( EulerConstruct( rRef(4:6)))))     ! apply rotation from PtfmInit onto input file's body orientation to get its true initial orientation
+            IF (p%Standalone /= 1) THEN 
+               !TODO: >>> should also maybe set reference orientation (which might make part of a couple lines down redundant) <<<
+               OrMat2 = MATMUL(OrMat, ( EulerConstruct( rRef(4:6))))  ! combine the Body's relative orientation with the turbine's initial orientation
+               u%CoupledKinematics(iTurb)%Orientation(:,:,J) = OrMat2          ! set the result as the current orientation of the body <<<
+
+               ! calculate initial point relative position, adjusted due to initial platform translations
+               u%CoupledKinematics(iTurb)%TranslationDisp(:,J) = InitInp%PtfmInit(1:3,iTurb) - rRef(1:3)
+               m%BodyList(m%CpldBodyIs(l,iTurb))%r6(1:3) = u%CoupledKinematics(iTurb)%Position(:,J) + u%CoupledKinematics(iTurb)%TranslationDisp(:,J) + p%TurbineRefPos(:,iTurb)
```
bjonkman, line 1896: Are these variables initialized elsewhere when `p%Standalone==1`? Or did the driver not initialize some `InitInp` values that are used here? It seems unusual to me that this would be set differently based on how the module is called.

modules/moordyn/src/MoorDyn.f90
RyanDavies19: I believe so, those sizes are set by the size of the state vector which is determined when the input file is being read as the variable Nx: https://github.com/RyanDavies19/MoorDynF_ryan/blob/5ebcc5987650a7d6b1d7a53a8f942f32d1a41bd0/modules/moordyn/src/MoorDyn.f90#L881

For reading in the input file and allocating etc. I followed the exact same setup that was being used for the existing coupled pinned rods.

RyanDavies19:
> Generally looks good to me. These are great additions. @RyanDavies19, what's the reason for moving some log file outputs from other modules into the init function?

@mattEhall it seemed to me that it simplified the code, rather than having the log being written to in many different files it all happens in in the main MoorDyn file. Happy to change if back though if you feel otherwise.

modules/moordyn/src/MoorDyn.f90
andrew-platt: Figured out my confusion.  In setting up the states there are 6 states for the pinned (3 DOF, and 3 derivatives of DOFs), and 12 states for the free bodies (6 DOF, and 6 derivatives of DOFs).  In the linearization section, we place motions first, so 3 states for pinned and 6 for free bodies. The derivatives are added to this matrix later giving the same total number of states, but ordered differently.

modules/moordyn/src/MoorDyn_Registry.txt
RyanDavies19: Agreed, good catch. I am removing them in the next commit

modules/moordyn/src/MoorDyn_Driver.f90
```diff
@@ -520,7 +532,11 @@ PROGRAM MoorDyn_Driver
          ! any coupled points (type -1)
          DO l = 1, MD_p%nCpldPoints(iTurb)
             
-            MD_u(1)%CoupledKinematics(iTurb)%TranslationDisp(:,K) = r_in(i, J:J+2) - MD_u(1)%CoupledKinematics(iTurb)%Position(:,K) - MD_p%TurbineRefPos(:,iTurb)
+            IF (MD_InitInp%Standalone == 1) THEN
```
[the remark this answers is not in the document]
RyanDavies19: Agreed, changing in the next commit as well.

modules/moordyn/src/MoorDyn.f90
RyanDavies19: They are initialized regardless of the standalone flag. I put them in the if here because the standalone flag is supposed to keep the object positions and orientation in the same location was where they are defined in the input file, and somewhere in these adjustments they were causing different reference frames to be superimposed and messing up the locations of objects defined in the input file. Would it be better practice to leave the MD file as is and just change the driver?

andrew-platt (approved):

modules/moordyn/src/MoorDyn.f90
bjonkman: Thanks for the info. It sounds like the `InitInp` values may not all be set in the driver like they are when coupled with OpenFAST. I would recommend changing the driver so that this code works the same way regardless of how it is coupled.

andrew-platt:
Thanks @bjonkman for reviewing!

jjonkman:
I'm not sure I understand the comment about the inertial flag and tight coupling.  I was not aware of problems with the acceleration-load feedback in the existing MoorDyn coupling to OpenFAST, which I figured was because the mass of the small portion of the cables attached to the floater is much smaller than the floater mass itself.  Are there existing coupling problems associated with MoorDyn within OpenFAST that I'm not aware of?

The tight coupling implementation is ongoing, but at this time, we were not planning to tightly couple MoorDyn within OpenFAST; rather, MoorDyn was going to remain loosely coupled with explicit loose coupling (along with many other OpenFAST modules).  Tight coupling is focused on OpenFAST modules BeamDyn, ElastoDyn, and SubDyn.  In tight coupling, implicit loose coupling will be used for OpenFAST modules ExtPtfm, HydroDyn, Orca, Simplified ElastoDyn, and AeroDyn for MHK.  Explicit loose coupling will be used for AeroDyn (for wind turbines), AeroDisk, ExtInflow, ExtLd, FEAMooring, IceDyn, IceFloe, InflowWind, MAP++, MoorDyn, SoilDyn, ServoDyn, and SeaState.  If there are coupling problems associated with MoorDyn, we may need to change this.

modules/moordyn/src/MoorDyn.f90
RyanDavies19: I looked at things a bit more and I don't think that it's possible to have the standalone feature reside solely in the driver unless the user also defines the coupled objects location in the drive file. With the coupling equations in MD right now, the only way to have the final initialized positions of a coupled object match the MD input file is for the driver to know those positions. Orientations are fine because the driver could just pass the identity matrix as OrMat, but translational positions are determined by a position difference, so zeroing out terms in the driver would lead to negative positions.

https://github.com/RyanDavies19/MoorDynF_ryan/blob/7bdeeb35eb588aaee40c60e9dd08190f8b0d2653/modules/moordyn/src/MoorDyn.f90#L1891

RyanDavies19: We could add a logic statement that makes the positions negative if farm < 1 which should resolve that, but to me that seems more complex than leaving things as is.

RBergua:
@jjonkman: I have experienced quite some issues with the rods to model cans for the USFLOWT project. MoorDyn tends to crash even at the beginning of the simulation when it has to return bending moments (e.g., using Coupled attachment in MoorDyn). See below for reference:
image

This "extreme pitch moment" is quite surprising to me because at t = 0 s the platform, as well as the cans, don't have a platform pitch or roll angle yet (they are straight upwards). Accordingly, there shouldn't be any bending moments associated. So, I'm not sure this is related to the tight coupling. To me, it seems some kind of limitation of the glue-code to transfer bending moments between MoorDyn and EdlastoDyn or SubDyn. I may be wrong...

For reference, even the CoupledPinned attachment sometimes return similar messages when the angles rotated are significant. Note that this CoupledPinned attachment is equivalent to a ball joint (i.e., the 3 rotational DOFs are free), so it shouldn't transfer any moments. 

I have also observed an unexpected behavior when using these cans that I'm trying to document at the moment... Maybe there is bug in the code.

RyanDavies19:
> I'm not sure I understand the comment about the inertial flag and tight coupling. I was not aware of problems with the acceleration-load feedback in the existing MoorDyn coupling to OpenFAST, which I figured was because the mass of the small portion of the cables attached to the floater is much smaller than the floater mass itself. Are there existing coupling problems associated with MoorDyn within OpenFAST that I'm not aware of?
> 
> The tight coupling implementation is ongoing, but at this time, we were not planning to tightly couple MoorDyn within OpenFAST; rather, MoorDyn was going to remain loosely coupled with explicit loose coupling (along with many other OpenFAST modules). Tight coupling is focused on OpenFAST modules BeamDyn, ElastoDyn, and SubDyn. In tight coupling, implicit loose coupling will be used for OpenFAST modules ExtPtfm, HydroDyn, Orca, Simplified ElastoDyn, and AeroDyn for MHK. Explicit loose coupling will be used for AeroDyn (for wind turbines), AeroDisk, ExtInflow, ExtLd, FEAMooring, IceDyn, IceFloe, InflowWind, MAP++, MoorDyn, SoilDyn, ServoDyn, and SeaState. If there are coupling problems associated with MoorDyn, we may need to change this.

The acceleration-load feedback issues only arise with the 6 DOF coupled objects in MoorDyn, rods and bodies. Because they have larger masses, the transients in acceleration cause large non-physical forces on the objects. Prior to this PR, the inertial responses of these objects had just been set to 0 as a workaround. As for the tight coupling plans, that was based on conversations I had with @andrew-platt

RyanDavies19:
> @jjonkman: I have experienced quite some issues with the rods to model cans for the USFLOWT project. MoorDyn tends to crash even at the beginning of the simulation when it has to return bending moments (e.g., using Coupled attachment in MoorDyn). See below for reference: image
> 
> This "extreme pitch moment" is quite surprising to me because at t = 0 s the platform, as well as the cans, don't have a platform pitch or roll angle yet (they are straight upwards). Accordingly, there shouldn't be any bending moments associated. So, I'm not sure this is related to the tight coupling. To me, it seems some kind of limitation of the glue-code to transfer bending moments between MoorDyn and EdlastoDyn or SubDyn. I may be wrong...
> 
> For reference, even the CoupledPinned attachment sometimes return similar messages when the angles rotated are significant. Note that this CoupledPinned attachment is equivalent to a ball joint (i.e., the 3 rotational DOFs are free), so it shouldn't transfer any moments.
> 
> I have also observed an unexpected behavior when using these cans that I'm trying to document at the moment... Maybe there is bug in the code.

@jjonkman I'll add that the acceleration-load feedback issues only arise with the 6 DOF coupled objects in MoorDyn, rods and bodies. We think that they have larger masses, the transients in acceleration cause large non-physical forces on the objects and cause OpenFAST to crash. Prior to this PR, the inertial responses of these objects had just been set to 0 as a workaround because they typically were small. With the USFLOWT design that's not the case and it was causing issues. @RBergua and I thought this flag was the best option because it kept the previous OpenFAST functionality but allowed for users to include the inertial terms if they didn't cause problems.

andrew-platt:
@jjonkman, I meant to bring this up in our tight coupling meeting, but forgot.  This does suggest that we may need to include MoorDyn in the tight coupling scheme. at least for models that include large buoyancy cans.

andrew-platt:
@RBergua, is it possible to reduce the numerical instability using correction steps, or does it still blow up when the moments are enabled?

I'm in favor of the short term solution with the option to disable the moments and merging the code.

RBergua:
I tried with several correction iterations, but couldn't make it work @andrew-platt. Often, the solver blows up at the first time step.

modules/moordyn/src/MoorDyn.f90
andrew-platt: If I remember correctly, MD has an unusual initialization of positions that was in part dictated by the initial relaxation of the mooring lines.  I vaguely remember exploring how to refactor that code and in the end concluded it was simpler to leave it as it is.  Given that, I can see how this `standalone` flag would be necessary.  I'm ok with leaving this in the module instead of moving it to the driver.

PR #1973, opened 2024-01-11
Use AD tower diameter for VTK visualization + minor improvements

bjonkman:

- If AeroDyn has tower diameter information, that is used for VTK visualization of the tower surface. Otherwise, it will use a tower diameter based on the NREL 5MW turbine (like it did before).
- The Registry now prints an error message before ending if a data type that should only contain reals finds an invalid type.
- Removed unnecessary dependency on SeaState in HydroDyn.
- Fixed some typos:
   - The word "Boeing" was misspelled in a few places.
   - The NWTC Library was potentially using the wrong number of bytes in some of its error messages if an array couldn't be allocated.
   - A line computing an index was repeated twice in FAST_Lin (likely a result of a merge).
   - I changed the case of the variable "l" to "L" for readability (it's easy to confuse lower-case l with the number 1 in some fonts)

None

- Tower surface VTK visualization
- Some error messages throughout the code

This does not change any test-case results. Only cases with AeroDyn tower diameter with VTK surfaces would look any different, but we don't have those kinds of tests.

### DIFF
--- a/modules/aerodyn/src/AirfoilInfo.f90
+++ b/modules/aerodyn/src/AirfoilInfo.f90
@@ -424,7 +424,7 @@ CONTAINS
             RETURN
          END IF
          
-         ! RelThickness, default is 0.2 if user doesn't know it, only used for Boing-Vertol UA model = 7
+         ! RelThickness, default is 0.2 if user doesn't know it, only used for Boeing-Vertol UA model = 7
       CALL ParseVarWDefault ( FileInfo, CurLine, 'RelThickness', p%RelThickness, 0.2_ReKi, ErrStat2, ErrMsg2, UnEc )
          if (ErrStat2 >= AbortErrLev) then ! if the line is missing, set RelThickness = -1 and move on...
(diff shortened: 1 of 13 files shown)


andrew-platt:
Thanks for adding this feature!  Do you have a picture you could share?

andrew-platt (approved):

bjonkman:
I ran the OpenFAST `IEA_LB_RWT-AeroAcoustics` r-test with VTK output both with and without the tower points defined in AD15. I chose this model only because it's not the NREL 5MW and it had blade surface data defined already. Here are the two plots (note, it seems like the blade airfoil coordinate files here have an error in their `x-y coordinate of airfoil reference` definition, causing the blades to look off [not centered] in these plots).

1. tower defined using 5MW proportions:
image

2. tower defined using `TwrElev` and `TwrDiam` from the AD15 file (note, I modified the tower elevation in the AD15 input file a little since ED defined the tower to be 110 m, but AD had it set to 140 m height).
image

PR #1986, opened 2024-01-16
Save registry structures to file without using a buffer

deslaughter:

This pull request changes the Registry and types files to pack/unpack data directly to/from a file without using temporary buffers; thereby reducing the memory used when generating a checkpoint file. This PR also updates the NWTC_Library so that the registry subroutines used by ModMeshMapping are automatically updated.

This PR also adds the ModVar module upon which the tight coupling changes are based.

NWTC_Library
*_Types.f90

The existing code generated by the OpenFAST registry to pack/unpack data structures to/from files used buffer arrays to temporarily store the data in memory before it was written. For models which require a large amount of data (InflowWind/SeaState); or which have a long runtime and use binary output; this could cause a significant amount of memory to be used. As these buffers grow, there is also the potential to overflow the indexing variable (typically an IntKi) which is only capable of representing a number up to 2,147,483,647. So if an array has more than 2 billion elements (which could happen when packing all the data into an array), the program will crash with a strange error. 

No test results changed. 5MW_OC3Mnpl_DLL_WTurb_WavesIrr_Restart, which checks the restart functionality, passed successfully.


andrew-platt:
I noticed that the file `Registry_NWTC_Library_mesh.txt` is used to generate `NWTC_Library_Subs.f90`.  I'm wondering if we should consider changing the filename `NWTC_Library_Subs.f90`.  I think all the other registry generated files currently end in `_Types.f90`, which kind of doesn't make sense for this file given it's contents.

Of course, if we do start auto-generating the params for each module through the registry (or similar) from csv files instead of the OutListParameters.xlsx, we'll have another set of auto-generated files using a file name other than `*_Types.f90`.

modules/openfast-registry/src/main.cpp
```diff
@@ -13,6 +13,7 @@ Usage: openfast_registry registryfile [options] -or-
     -h                this summary
     -I <dir>          look for usefrom files in directory "dir"
     -O <dir>          generate types files in directory "dir"
+    -inc              generate types file to be included in another file
```
andrew-platt, line 16: Should this be `subs` for the new flag?

andrew-platt (changes requested):
the help menu for the registry might have a typo

deslaughter:
I'd be fine with changing the name since this solution is a bit of a hack. The existing process requires two steps: 1) generate `NWTC_Library_Types.f90` from `Registry_NWTC_Library_mesh.txt` and copy just the subroutines to `ModMesh_Mapping.f90` 2) generate the actual `NWTC_Library_Types.f90` from `Registry_NWTC_Library_base.txt`. The new code generates just the subroutines from `Registry_NWTC_Library_mesh.txt` and stores it in `NWTC_Library_Subs.f90` which is `include`ed in `ModMesh_Mapping.f90`

andrew-platt:
I definitely like the new solution.  Could we do something like `_RegSubs.f90` for a name?  Maybe not the best name, but it makes it a bit more distinct.

modules/openfast-registry/src/main.cpp
deslaughter: changed to incsubs and updated documentation in new commit

andrew-platt (approved):

PR #1990, opened 2024-01-16
MD version update

RyanDavies19:
Updates the outputs to the command line to reflect most recent MoorDyn version and dates.

### DIFF
--- a/modules/moordyn/src/MoorDyn_Driver.f90
+++ b/modules/moordyn/src/MoorDyn_Driver.f90
@@ -114,7 +114,7 @@ PROGRAM MoorDyn_Driver
   
    CHARACTER(20)                         :: FlagArg              ! flag argument from command line
    CHARACTER(200)                        :: git_commit    ! String containing the current git commit hash
-   TYPE(ProgDesc), PARAMETER             :: version = ProgDesc( 'MoorDyn Driver', '', '' )
+   TYPE(ProgDesc), PARAMETER             :: version = ProgDesc( 'MoorDyn Driver', '', '2024-01-18' )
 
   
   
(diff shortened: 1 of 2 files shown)


andrew-platt (approved):

modules/moordyn/src/MoorDyn_Driver.f90
```diff
@@ -132,7 +132,7 @@ PROGRAM MoorDyn_Driver
    IF ( LEN( TRIM(FlagArg) ) > 0 ) CALL NormStop()
 
       ! Display the copyright notice
-   CALL DispCopyrightLicense( version%Name, 'Copyright (C) 2021 NREL, 2019 Matt Hall' )
+   CALL DispCopyrightLicense( version%Name, 'Copyright (C) 2024 NREL, 2019 Matt Hall' )
```
bjonkman, line 135: You could get rid of the '2024 NREL' copyright here since that's done automatically in `DispCopyrightLicense`.

modules/moordyn/src/MoorDyn_Driver.f90
```diff
@@ -144,7 +144,7 @@ PROGRAM MoorDyn_Driver
    CALL CPU_TIME ( ProgStrtCPU )                                    ! Initial time (this zeros the start time when used as a MATLAB function)
    
 
-   CALL WrScr( ' MD Driver updated 2022-01-12')
+   CALL WrScr( ' MD Driver updated 2024-01-16')
```
bjonkman, line 147: I'd recommend printing `TRIM(MD_ProgDesc%Date)` instead of hardcoding the date here.

modules/moordyn/src/MoorDyn_Driver.f90
RyanDavies19: Resolved with my most recent commit

modules/moordyn/src/MoorDyn_Driver.f90
RyanDavies19: I made this change in my most recent commit. I also removed a call to DispCopyrightLicense in the driver because MoorDyn.f90 displays the license when called by the driver.

andrew-platt (approved):

PR #1992, opened 2024-01-18
SeaState: fix grid size in wave surface visualization

andrew-platt:
This PR is ready to merge.

When the _SeaState_ module was introduced, the grid used for wave surface visualization was based on interpolated data from the _SeaState_ wave grid stored in the _WaveField_ data.  The wave surface size was hard-coded to be 4x the blade length of the turbine being modeled.  This resulted in clamping any parts of the wave surface outside the _WaveField_ data.  As result, the visualization looks a bit strange.

***Before***
Screenshot 2024-01-17 at 5 52 33 PM
Figure 1: before modification, wave surfaces beyond the _WaveField_ data would be set to the value of the nearest edge of the _WaveField_ data.  The _WaveField_ edges are shown as pink vertical boundaries.  In the original formulation, 25x25 data points were interpolated from the _WaveField_ data and stored.  For this visualization, the `5MW_OC4Semi_WSt_WavesWN` regression test was modified with `WaveHs=55`, `WaveTp=4`, `WaveDir=30`, `WaveDirMod=1`, `WaveDirSpread=1`, `WaveNDir=75`, `WaveDirRange=60`, and second order turned off.

***After***
Screenshot 2024-01-17 at 5 55 24 PM
Figure 2: after modification, the wave surfaces for visualization are bounded by the `WaveField` data set and use the native _WaveField_ grid resolution specified in the _SeaState_ input file (6 nodes in each half-width direction).  The same _WaveField_ is used here as in the original case above.

None.

Visualization of wave surface.

Before the addition of the _SeaState_ module, the grid had been set to 25x25 points.  The sea surface at each point was calculated from the FFT of the wave spectral data.  After the addition of the _SeaState_ module, each point requested was interpolated from the _WaveField_ data.  It would be nice to someday restore the ability to calculate an arbitrary number of points for the sea surface from the FFT of the wave spectral data for that location, but I don't have time right now to modify this.

None.

### DIFF
--- a/modules/openfast-library/src/FAST_Subs.f90
+++ b/modules/openfast-library/src/FAST_Subs.f90
@@ -777,20 +777,6 @@ SUBROUTINE FAST_InitializeAll( t_initial, p_FAST, y_FAST, m_FAST, ED, BD, SrvD, 
 
 
    ! ........................
-   ! set some VTK parameters required before SeaState init (so we can get wave elevations for visualization)
-   ! ........................
-
-      ! get wave elevation data for visualization
-   if ( p_FAST%WrVTK > VTK_None ) then
-      call SetVTKParameters_B4SeaSt(p_FAST, Init%OutData_ED, Init%InData_SeaSt, BD, ErrStat2, ErrMsg2)
-         CALL SetErrStat(ErrStat2,ErrMsg2,ErrStat,ErrMsg,RoutineName)
-         IF (ErrStat >= AbortErrLev) THEN
-            CALL Cleanup()
-            RETURN
-         END IF
-   end if
-
-   ! ........................
    ! initialize SeaStates
    ! ........................
    ALLOCATE( SeaSt%Input( p_FAST%InterpOrder+1 ), SeaSt%InputTimes( p_FAST%InterpOrder+1 ), STAT = ErrStat2 )
@@ -816,8 +802,11 @@ SUBROUTINE FAST_InitializeAll( t_initial, p_FAST, y_FAST, m_FAST, ED, BD, SrvD, 
       Init%InData_SeaSt%WaveFieldMod  = p_FAST%WaveFieldMod
       Init%InData_SeaSt%PtfmLocationX = p_FAST%TurbinePos(1)
       Init%InData_SeaSt%PtfmLocationY = p_FAST%TurbinePos(2)
-      
+
       Init%InData_SeaSt%TMax          = p_FAST%TMax
+      
+         ! wave field visualization
+      if (p_FAST%WrVTK == VTK_Animate .and. p_FAST%VTK_Type == VTK_Surf) Init%InData_SeaSt%SurfaceVis = .true.
       
       CALL SeaSt_Init( Init%InData_SeaSt, SeaSt%Input(1), SeaSt%p,  SeaSt%x(STATE_CURR), SeaSt%xd(STATE_CURR), SeaSt%z(STATE_CURR), &
                           SeaSt%OtherSt(STATE_CURR), SeaSt%y, SeaSt%m, p_FAST%dt_module( MODULE_SeaSt ), Init%OutData_SeaSt, ErrStat2, ErrMsg2 )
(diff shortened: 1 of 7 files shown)


andrew-platt:
@luwang00 and @RBergua, could you take a look at this?

luwang00:
@andrew-platt 
Thanks for fixing this! I agree that constraining the wavefield visualization to the SeaState wave grid is a good solution. This also helps the user in checking whether or not the structure stays within the SeaState wave grid throughout the simulation.

RBergua:
It looks good to me @andrew-platt. Thanks for the effort!

modules/openfast-library/src/FAST_Subs.f90
```diff
@@ -816,17 +802,28 @@ SUBROUTINE FAST_InitializeAll( t_initial, p_FAST, y_FAST, m_FAST, ED, BD, SrvD,
...
          CALL SetErrStat(ErrStat2,ErrMsg2,ErrStat,ErrMsg,RoutineName)
          
       p_FAST%ModuleInitialized(Module_SeaSt) = .TRUE.
       CALL SetModuleSubstepTime(Module_SeaSt, p_FAST, y_FAST, ErrStat2, ErrMsg2)
          CALL SetErrStat(ErrStat2,ErrMsg2,ErrStat,ErrMsg,RoutineName)
-         
+
+      if (allocated(Init%OutData_SeaSt%WaveElevVisGrid)) then
+         p_FAST%VTK_surface%NWaveElevPts(1) = size(Init%OutData_SeaSt%WaveElevVisX)
+         p_FAST%VTK_surface%NWaveElevPts(2) = size(Init%OutData_SeaSt%WaveElevVisX)
```
bjonkman, line 821: @andrew-platt : Should line 821 use `Init%OutData_SeaSt%WaveElevVisY` instead of `WaveElevVisX`?

modules/seastate/src/SeaState.txt
```diff
@@ -89,8 +91,11 @@ typedef   ^                  InitOutputType         CHARACTER(ChanLen)       Wri
 typedef   ^                  ^                      CHARACTER(ChanLen)       WriteOutputUnt    {:}     -      -  "The is the list of all HD-related output channel unit strings (includes all sub-module channels)"    -
 typedef   ^                  ^                      ProgDesc                 Ver                -      -      -  "Version of SeaState"
 typedef   ^                  ^                      LOGICAL                  InvalidWithSSExctn -      -      -  "Whether SeaState configuration is invalid with HydroDyn's state-space excitation (ExctnMod=2)" (-)
-typedef   ^                  ^                      SiKi                     WaveElevSeries    {:}{:}  -      -  "Wave elevation time-series at each of the points given by WaveElevXY.  First dimension is the timestep. Second dimension is XY point number corresponding to second dimension of WaveElevXY." (m)
+typedef   ^                  ^                      SiKi                     WaveElevVisX       {:}     -      -  "X locations of grid output"    "m,-"
+typedef   ^                  ^                      SiKi                     WaveElevVisY       {:}     -      -  "X locations of grid output"    "m,-"
```
bjonkman, line 95: `WaveElevVisY` should say `Y locations of grid output`, not `X` (was done correctly in FAST_Registry.txt).

modules/openfast-library/src/FAST_Subs.f90
```diff
@@ -4070,13 +4007,18 @@ SUBROUTINE SetVTKParameters(p_FAST, InitOutData_ED, InitOutData_AD, InitInData_S
    !.......................
 
    !bjj: interpolate here instead of each time step?
-   if ( allocated(InitOutData_SeaSt%WaveElevSeries) ) then
-      call move_alloc( InitInData_SeaSt%WaveElevXY, p_FAST%VTK_Surface%WaveElevXY )
-      call move_alloc( InitOutData_SeaSt%WaveElevSeries, p_FAST%VTK_Surface%WaveElev )
+   if ( allocated(InitOutData_SeaSt%WaveElevVisGrid) ) then
+print*,'Storing Wave surface visualization'
```
bjonkman, line 4011: Is this a leftover debugging statement?

modules/openfast-library/src/FAST_Subs.f90
andrew-platt: I think you are right.  They aren't necessarily the same size.

modules/seastate/src/SeaState.txt
andrew-platt: Good catch!

modules/openfast-library/src/FAST_Subs.f90
andrew-platt: Not sure how that got left in there.

I'll make a PR to fix the issues you pointed out.  Thanks @bjonkman!

andrew-platt: Addressed in PR #2071

PR #1999, opened 2024-01-19
Registry: add flag to allow integers in InputType

andrew-platt:
This is ready to merge.

PR #1973 modified the registry to check for invalid types in the `InputType`.  However, the `ExtLoads` registry includes integers in the `InputType`.  We really don't want integers in that derived type as it breaks with linearization.  

Rather than change the `ExtLoads` registry and potentially break downstream codes that depend on it (AMR-Wind for example), I propose we allow an exception here.  Unless I add such an exception, or force downstream changes, we cannot merge the `dev` branch into the `dev-unstable-pointers` branch.

I added a flag `-inputintallow` to the registry for use in such special cases.

Registry only.

None.

### DIFF
--- a/modules/openfast-registry/src/registry_gen_fortran.cpp
+++ b/modules/openfast-registry/src/registry_gen_fortran.cpp
@@ -133,11 +133,14 @@ void Registry::gen_fortran_module(const Module &mod, const std::string &out_dir)
 
         // If derived data type should only contain reals,
         // verify that it does, otherwise exit with error
-        if ((ddt.interface != nullptr) && ddt.interface->only_reals)
-            if (!ddt.only_contains_reals())
+        if (!this->input_int_allow)
             {
-                std::cerr << "Registry warning: Data type '" << dt_name << "' contains non-real values." << std::endl;
-                exit(EXIT_FAILURE);
+            if ((ddt.interface != nullptr) && ddt.interface->only_reals)
+                if (!ddt.only_contains_reals())
+                {
+                    std::cerr << "Registry warning: Data type '" << dt_name << "' contains non-real values." << std::endl;
+                    exit(EXIT_FAILURE);
+                }
             }
 
         // Write derived type header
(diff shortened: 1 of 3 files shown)


andrew-platt:
@bjonkman, do you have an opinion on this approach?

bjonkman:
One reason for not having integers in the Input type is because of the extrapolation/interpolation routines. Are you omitting those routines from this module, too?

bjonkman:
Also, that PR didn't ADD a check on the invalid type, it just printed the error message before ending. Otherwise, the Registry would end without an error message and without completing.

modules/openfast-registry/src/registry_gen_fortran.cpp
```diff
@@ -133,11 +133,14 @@ void Registry::gen_fortran_module(const Module &mod, const std::string &out_dir)
 
         // If derived data type should only contain reals,
         // verify that it does, otherwise exit with error
-        if ((ddt.interface != nullptr) && ddt.interface->only_reals)
-            if (!ddt.only_contains_reals())
+        if (!this->input_int_allow)
             {
-                std::cerr << "Registry warning: Data type '" << dt_name << "' contains non-real values." << std::endl;
-                exit(EXIT_FAILURE);
+            if ((ddt.interface != nullptr) && ddt.interface->only_reals)
```
deslaughter, line 138: Instead of adding a new flag, you could disable this check if the `-noextrap` flag has been set: `if ((ddt.interface != nullptr) && ddt.interface->only_reals && !this->no_extrap_interp)` and then add that flag to for this module. I didn't see extrap/interp being called for ExternalInflow.

bjonkman:
@andrew-platt , The integers that I see in that `InputType` for `ExtLoadsDX` seem like they should be parameters or initialization inputs.
image

I know it's more work to fix that module (and the interface to CFD), but I think in the long term, it will be better to make the ExtLoads module align with the OpenFAST framework as much as possible.

andrew-platt:
After looking at this a bit more, I agree with @bjonkman and @deslaughter we should simply not allow integers in the `InputType` regardless.  These particular values really are parameters and should be treated as such.  In the long run it will be better to fix this properly and align the _ExtLoads_ module with the framework.

There also appear to be some additional issues to resolve from PR #1932, so we will address this along those issues in a different PR.

So, I'm going to close this PR without merging.

andrew-platt:
Better solution than this PR: #2001

Issue #1978, opened 2024-01-13
MoorDyn: unexpected behavior at initialization


RBergua:
**Bug description**
I'm modeling a system in OpenFAST with buoyant cans as follows:
![image](https://github.com/OpenFAST/openfast/assets/33709140/1a5a51dd-12b0-4b32-95a7-4b163c344ff6)

The cans are modeled in MoorDyn by means of rods and I'm using the CoupledPinned attachment available in OpenFAST dev (4fafe68) to allow the 3 rotational DOFs at the connection with the legs (like a ball joint).

For these tests, I consider still water conditions and no inflow wind (i.e., it's a gravity-only condition). I run these tests with a rigid system. I only account for the rotation of the cans (CoupledPinned attachment in MoorDyn) and the 6 rigid DOFs of the platform.

When I use TmaxIC = 0 s (no initial computation in MoorDyn), I can observe that the platform pitch oscillates around -1.9 deg and the cans experience rotation amplitudes smaller than 1.5 deg. This behavior is the expected one.
![image](https://github.com/OpenFAST/openfast/assets/33709140/8aaf9916-c278-4031-8cd4-c71be23002d1)

However, if I define a given time to compute the initial condition (e.g., TmaxIC = 100), I get converged magnitudes for the fairlead tensions:
![image](https://github.com/OpenFAST/openfast/assets/33709140/5c25d965-979f-48a2-a802-efecbf400b38)

But the cans are all over the place (in the below image you can see that the cans are tilted around 85 deg at t = 0 s in OpenFAST, so the cans are almost horizontal instead of vertical) and the platform pitch behavior is erratic due to the very large motion of the cans:
![image](https://github.com/OpenFAST/openfast/assets/33709140/9657eb27-4077-4291-b4b7-12289e1934cf)

Summarizing: Using TmaxIC different than 0 seconds in MoorDyn results in unexpected responses when using rods and CoupledPinned attachments.

Additional note: Maybe this large initial transient is related to the crashes experienced in MoorDyn when using Coupled attachments ("Warning: extreme pitch moment from body-attached Rod") that we commented offline.

RyanDavies19:
@RBergua Have you tried with CdScaleIC = 1? Maybe the larger drag in initialization is causing issues. Also, the way that the IC generation is set up is based on line tension convergence, so while the line tensions converge that doesn't necessarily mean the bodies positions converge.

RBergua:
[...]

PR #2005, opened 2024-01-25
Removes MD driver standalone option, bug fix on initialization

RyanDavies19:
Fixes an issue where coupled bodies did not initialize at the correct location when coupled with OpenFAST, especially when platform offsets were provided. @mattEhall authored this bug fix.

Removes the standalone flag from inside of MoorDyn.f90 and MoorDyn_Driver.f90, previously triggered by numturbines < 0 in the driver input file. The standalone issue was a work around to the bug fixed in this PR and thus is no longer needed. 

#1978 
#1990 

MoorDyn

@andrew-platt can you check that I made the correct changes to the functions in MD_Types.f90 (Pack and unpack routines). I believe in looking at the git history you set that up. 


andrew-platt:
Looks good!  Thanks for revising how this works!

PR #2026, opened 2024-02-08
SeaState: combine SeaSt_Interp into SeaSt_WaveField, and bug fix

andrew-platt:
This is ready for merging

@luwang00, please review.
@RBergua, please check that this now works for your model.

There was a bug in the how the indices for time in the wave field data was calculated when `deltaT` was zero.  This resulted in nonsensical bounds for time, which caused a segmentation fault during interpolation.

Included in this PR:

1. `SeaState_Interp` has been combined into the `SeaSt_WaveField` module.  While attempting to find the root issue, it became obvious that there was no reason to keep these separated.  This includes revisions to how the misc vars and parameters are stored for WaveField.
2. Parts of HydroDyn were directly accessing the `SeaState_Interp` routines.  New interfaces were provided within `SeaSt_WaveField` so that the WaveField discretization information could be used in HydroDyn to find the indices and weighting during interpolation (HydroDyn shares the same grid discetization for the WAMIT forces).
3. `Waves.f90` was revised with some updated error handling so that it is more readable.  I couldn't figure out what was going on before cleaning and formatting a bit.
4. All direct calls to interp functions from `SeaState.f90` were converted to WaveField function calls instead.
5. A check on `deltaT` was added to the interpolation functions.

@RBergua ran into a segmentation fault with the HydroDyn driver when running linearization from the driver only.  The model is proprietary so it cannot be shared.

HydroDyn and SeaState

A new HD driver test should be added to exercise the driver linearization function.


modules/seastate/src/SeaSt_WaveField.f90
```diff
@@ -432,6 +423,442 @@ SUBROUTINE WaveField_GetWaveKin( WaveField, SeaSt_Interp_m, Time, pos, forceNode
...
+   isopc   = -1.0
+   Indx_Lo = 0
+   Indx_Hi = 0
+   if ( Time < 0.0_DbKi ) then
+      CALL SetErrStat(ErrID_Fatal,'Time value must be greater than or equal to zero!',ErrStat,ErrMsg,'SetTimeIndex') !error out if time is outside the lower bounds
+      RETURN
+   end if
+
+   ! if there are no timesteps, don't proceed
+   if (EqualRealNos(deltaT,0.0_ReKi) .or. deltaT < 0.0_ReKi)  return;
```
andrew-platt, line 591: @luwang00 & @RBergua, this check was the root cause of the seg-faults.

modules/seastate/src/SeaSt_WaveField.f90
```diff
@@ -432,6 +423,442 @@ SUBROUTINE WaveField_GetWaveKin( WaveField, SeaSt_Interp_m, Time, pos, forceNode
...
+   end function
+end subroutine WaveField_GetWaveKin
+
+
+!----------------------------------------------------------------------------------------------------
+! Interpolation related functions
+!----------------------------------------------------------------------------------------------------
+
+!> Set the WaveField 4D Params
+subroutine WaveField_SetParam( InitInp, p )
```
bjonkman, line 439: I'd be inclined to remove the `InitInp` type from the WaveField module. We set a lot of the WaveField parameters in different routines, and this one only sets values that were in the SeaState_Interp module. WaveField also didn't have any InitInp types before this.

andrew-platt: Thanks for looking at this PR!

I agree; `InitInp` really doesn't need to exist at all.  I debated removing this at the time, but in the end copied over the `InitInput` from the `SeaState_Interp` for simplicity.

While we could remove the `SetParam` routine as well, I'm inclined to leave the it to force future programmers to set all the parameters.

bjonkman: I'm not sure we can set all the parameters for WaveField in this routine without rewriting all of SeaState from scratch. :-) Most of the SeaState, Waves, and Waves2 init routine set different parameters in WaveField. It gets particularly tricky in the Waves modules where parameters get set differently based on which type of waves are being modeled. If you're going to keep this routine, maybe rename it?

modules/seastate/src/SeaState.f90
```diff
@@ -325,31 +255,22 @@ SUBROUTINE SeaSt_Init( InitInp, u, p, x, xd, z, OtherState, y, m, Interval, Init
...
-      SeaSt_Interp_InitInp%n        = (/p%WaveField%NStepWave,p%nGrid(1),p%nGrid(2),p%nGrid(3)/)
-      SeaSt_Interp_InitInp%delta    = (/real(p%WaveDT,ReKi),p%deltaGrid(1),p%deltaGrid(2),p%deltaGrid(3)/)
-      SeaSt_Interp_InitInp%pZero(1) = 0.0  !Time
-      SeaSt_Interp_InitInp%pZero(2) = -InputFileData%X_HalfWidth
-      SeaSt_Interp_InitInp%pZero(3) = -InputFileData%Y_HalfWidth
-      SeaSt_Interp_InitInp%pZero(4) = -InputFileData%Z_Depth  ! zi
-      SeaSt_Interp_InitInp%Z_Depth  =  InputFileData%Z_Depth
-      call SeaSt_Interp_Init(SeaSt_Interp_InitInp, p%WaveField%seast_interp_p,  ErrStat2, ErrMsg2)
-         CALL SetErrStat(ErrStat2,ErrMsg2,ErrStat,ErrMsg,RoutineName)
+      WaveField_InitInp%n        = (/p%WaveField%NStepWave,p%nGrid(1),p%nGrid(2),p%nGrid(3)/)
```
bjonkman, line 266: Related to my other comment... it seems like it would be easier to just set `p%WaveField%GridParams%*` here instead of making them `WaveField_InitInp%*` and then copying them over in the `WaveField_SetParam` routine. I've debated removing this with the SeaSt_Interp module many times in the past :-).

RBergua:
@andrew-platt I confirm that this PR runs with my OpenFAST model (Morison-only approach) and returns the expected results:
image

modules/seastate/src/SeaSt_WaveField.f90
luwang00: @andrew-platt Thanks for fixing this!

modules/seastate/src/SeaState.f90
andrew-platt: I agree. I'm going to remove the `InitInp` type entirely and remove the init routine.  They are just extra code that isn't needed.

andrew-platt: Done!

modules/seastate/src/SeaSt_WaveField.f90
andrew-platt: Removed!

deslaughter (approved):

luwang00:
Thanks @andrew-platt for the bug fix and the code cleanup. Looks great to me.

PR #2069, opened 2024-03-05
HD bug fixes to the inertial loads from marine growth and ballast water on strip-theory members

luwang00:
This is a bug fix to the Morison module of HydroDyn to address the following bugs:
* There should not be a parallel-axis correction to the axial component of the marine-growth/ballast-water moment of inertia.
* The matrix `Imat` used to rotate the moment of inertia to the earth-fixed system was not properly zeroed for each node.

HydroDyn

**Test results** 
Interestingly, all existing r-tests passed as is.


luwang00:
Hi @jjonkman and @andrew-platt, could you take a quick look at this and see if you agree with the changes. Thanks!

andrew-platt (approved):

jjonkman (approved):
Makes sense to me.  Thanks for tracking this down!

andrew-platt:
This probably also affects 3.5.2, so we should include it in 3.5.3

PR #2073, opened 2024-03-05
HD bug fix to the inertial load from ballast water on strip-theory members

luwang00:
The moment of inertia matrix for ballast water was not transformed to the earth-fixed coordinate system. With this PR, the ballast-water moment of inertia matrix is transformed similar to the marine-growth moment of inertia.

HydroDyn

Waiting for the r-tests to complete.


jjonkman (approved):
I agree with this fix, which matches the implementation plan.  Thanks!

PR #2078, opened 2024-03-07
Minor code cleanup

bjonkman:
This PR removes some unused variables and replaces non-standard Fortran 2003 code with code that uses F03. I also saved the updated Visual Studio FASTlib project with the folders in alphabetical order to prevent it looking like there are differences every time you save the Visual Studio solution file to build OpenFAST.

none.

There should be no changes to any results.


andrew-platt (approved):
Thanks for updating the `vfproj` file!  We'll try not to rearrange it in future PRs.

bjonkman:
> Thanks for updating the `vfproj` file! We'll try not to rearrange it in future PRs.

When you renamed the OpFM module, it made sense to not move it to alphabetical order so we could see the differences easier.

PR #2079, opened 2024-03-08
Handling of Paths for SrvD UserSubs Input Files

### LINKED ISSUE #2059: Handling of Paths for SrvD UserSubs Input Files
**Is your feature request related to a problem? Please describe.**

During Loads Analysis, it is convenient to leave Input and Parameter Files that do not change throughout the analysis in a "Template" or "Master" folder. This is the case, for example, of input files used by ServoDyn, such as `spd_trq.dat` and similar. 
Currently, `UserVSCont` is called with `p%RootName` as an argument, but that will seek `spd_trq.dat` in the "current" active folder of the `.fst`. Yet, the controller may not be changing across various DLC runs, and it would be efficient to leave it unaltered within a "template" folder thatcontains the main SrvD input file as well.

**Describe the solution you'd like**
 For this reason, `p%RootName` could be replaced by `p%PriPath` in the call to the UserSubs in SrvDyn. `PriPath`, path to the primary SrvD input file is already available, thus a small change to the registry would take care of this. I have implemented this change, and it seems to work out well. Alternatively, one would have to copy the `spd_trq` n_DLCs times in the various DLC folders...

andrew-platt:

In Issue #2059, @rdamiani pointed out that the existing method for handling file paths with `UserSubs` is difficult as it requires putting all the user input files in the same location as the ServoDyn file.  This PR gives a method that from @rdamiani for a method he has had good success with.  

This PR is entirely @rdamini's work merely copied into a PR by me.  

Co-authored-by:rdamiani <r.damiani@RRDengineering.com>

Closes #2059 

_ServoDyn_ controls where the `UserSubs` option is used (`PCMode`, `VSContrl`, `HSSBrMode`, or `YCMode` =3).

The usage of the affected routines is a very specialized case that we don't directly support, but certainly want to make life easier for users of these features.

None.

### DIFF
--- a/modules/servodyn/src/ServoDyn_Registry.txt
+++ b/modules/servodyn/src/ServoDyn_Registry.txt
@@ -439,6 +439,7 @@ typedef	^	ParameterType	IntKi	StCCmode	-	-	-	"Structural control control mode {0
 typedef	^	ParameterType	IntKi	NumOuts	-	-	-	"Number of parameters in the output list (number of outputs requested)"	-
 typedef	^	ParameterType	IntKi	NumOuts_DLL	-	-	-	"Number of logging channels output from the DLL (set at initialization)"	-
 typedef	^	ParameterType	CHARACTER(1024)	RootName	-	-	-	"RootName for writing output files"	-
+typedef	^	ParameterType	CHARACTER(1024)	PriPath  	-	-	-	"Path of the primary SD input file "	-
 typedef	^	ParameterType	OutParmType	OutParam	{:}	-	-	"Names and units (and other characteristics) of all requested output parameters"	-
 typedef	^	ParameterType	CHARACTER(1)	Delim	-	-	-	"Column delimiter for output text files"	-
 # parameters for Bladed Interface (dynamic-link library)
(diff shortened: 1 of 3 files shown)


modules/servodyn/src/ServoDyn_Registry.txt
```diff
@@ -439,6 +439,7 @@ typedef	^	ParameterType	IntKi	StCCmode	-	-	-	"Structural control control mode {0
 typedef	^	ParameterType	IntKi	NumOuts	-	-	-	"Number of parameters in the output list (number of outputs requested)"	-
 typedef	^	ParameterType	IntKi	NumOuts_DLL	-	-	-	"Number of logging channels output from the DLL (set at initialization)"	-
 typedef	^	ParameterType	CHARACTER(1024)	RootName	-	-	-	"RootName for writing output files"	-
+typedef	^	ParameterType	CHARACTER(1024)	PriPath  	-	-	-	"Path of the primary SD input file "	-
```
bjonkman, line 442: `SD` is `SubDyn`, so this should be changed to `SrvD`.

andrew-platt: Corrected directly on rc-3.5.3 with commit 7a2c43e8b

PR #1910, opened 2023-12-01
Unsteady Aero Driver:  adding 3 degrees of freedom for the motion of a genralized airfoil section

ebranlard:
This pull request is soon ready to be merged

**Feature or improvement description**
Adding 3 degrees of freedom fro the motion of a genralized airfoil section 

**Impacted areas of the software**
Unsteady Aerodynamics 

**Additional supporting information**
A paper will be released at the torque 2024 conference. A link is provided in the documentation. 

**Test results, if applicable**
Input file is backward compatible

- [x] Update the documentation  - Live version [here](https://ebranlard-openfast.readthedocs.io/en/f-ua-elast-dev/source/user/aerodyn/theory_ua.html#driver) 
- [x] Update r-test input files
- [x] Submit a new r-test for aeroelastic simulation will be done later after torque paper
- [ ] Quick review

Example input file:
[UA2.txt](https://github.com/OpenFAST/openfast/files/13519950/UA2.txt)

PR #2091, opened 2024-03-14
Fix bug in new UA driver

bjonkman:
The code was setting an array of size 4 equal to an array of size 3, causing an issue in the debugger.

This was introduced in https://github.com/OpenFAST/openfast/pull/1910

The new UA driver

This likely won't change results; I don't think this particular part of UA driver has a regression test.


ebranlard (approved):

ebranlard:
Thank you, this extra column was indeed not needed.

PR #2094, opened 2024-03-15
Support cross-compiling with Mingw 

pablo-benito:

_Support cross-compiling openfast in Linux (to build a Windows version)_

I'm using Mingw compiler on Linux to cross-compile openfast and build a Windows version (yes, geek but useful for me ...)  
Unfortunately the mingw windows header file is "windows.h". In a case-sensitive system (like linux) the build fails in this line, as it's trying to include the header "Windows.h" ('W' upper-case)

Changing the include line to `#include <windows.h>` solves my problem and I'm able to cross-compile openfast. The change should not affect the windows native build, as this is a case-insensitive system.

None 

None

### DIFF
--- a/modules/map/src/mapsys.h
+++ b/modules/map/src/mapsys.h
@@ -33,7 +33,7 @@
 
 
 #if defined(_WIN32) || defined(_WIN64)
-#  include <Windows.h>
+#  include <windows.h>
 #  include <tchar.h>
 #else
 #  include <unistd.h>


deslaughter (approved):
I think this change makes sense and is supported by https://stackoverflow.com/questions/15466613/lowercase-windows-h-and-uppercase-windows-h-difference.

andrew-platt:
@deslaughter, should we include this in 3.5.3?

deslaughter:
I think that’s a good idea.

Issue #2096, opened 2024-03-15
Inconsistent usage turbine indexing with CPP interfacing to OpenFAST library


andrew-platt:
**Bug description**
Within the `FAST_AllocateTurbines` routine, the `Turbine` data structure is allocated using the c/c++ convention of index 0 start (`Turbine(0:NumTurbines-1)`.  However, elsewhere in the _OpenFAST_ library code, the indexing to this array of derived types is assumed to start at 1 (`Turbine(1:NumTurbines)`).  This has caused issues with checkpoint files not getting closed when called from AMR-Wind (see https://github.com/Exawind/amr-wind/issues/984).

I propose we remedy this by:

1. Changing the name of `ExternInitData%TurbineID` to ExternInitData%TurbIDforName` to properly reflect the usage of this variable.  This variable is only ever used to set the name of the output file (_i.e._ `SimName.T0.out`).  This variable may be allowed to be 0 or larger.
2. All data storage for turbine data within OpenFAST will use the Fortran indexing starting at index 1.
3. Arguments passed from c/c++ for a turbine ID will be incremented by 1 to match the internal indexing used by OpenFAST.
4. Update any documentation about this indexing

Since the name is generated from `ExternInitData%TurbIDforName`, it is completely independent of the internal data storage array indexing method.

This should correct the issue with closing checkpoint files from AMR-Wind.

andrew-platt:
Fixed with #2097

PR #2097, opened 2024-03-15
Consistent use of turbine indexing when coupled to c/c++

andrew-platt:
This is ready to merge, pending discussion.

There has been an ambiguous usage of turbine indexing.  In C and C++ codes, arrays start with index 0, but Fortran starts with index 1.  In coupling _OpenFAST_ to _AMR-Wind_, an array of `Turbine` data structures was created spanning from index `0:NumTurbines-1`.  Indexing into this array has been handled by the `iTurb` variable passed into all routines in  `FAST_Library.f90` that include a c-bindings interface.  However, inside the `FAST_Subs.f90` file, this turbine index is assumed to run from `1:NumTurbines` for compatibility with _FAST.Farm_.

This PR corrects the following:

1. Change the variable `ExternInitData%TurbineID` to` ExternInitData%TurbIDforName` to more accurately reflect the usage of this variable.  This variable is only ever used to set the name of the output file (_i.e._ `SimName.T0.out`). This variable must be 0 or larger (new error handling added).
2. `FAST_AllocateTurbines` now allocates the `Turbine` data structure to the range `(1:NumTurbines)`.
3. All routines in `FAST_Library.f90` that include a c-bindings interface now take the argument `iTurb_c` instead of `iTurb`.  This is the index to the turbine using the c-style array indexing starting at 0.  Internally this shifted by 1 and stored as `iTurb`, a Fortran index starting at 1.

Since the output names are set using the `TurbIDforName` variable, the name is completely independent of the internal numbering.

https://github.com/Exawind/amr-wind/issues/984
#2096

This will affect any codes calling _OpenFAST_ using the `FAST_Library` interface using 1 as starting index for the first turbine -- these codes will need updating to use 0 as the index to the first turbine.  This _might_ affect NALU-Wind, and potentially SOWFA (I'm uncertain what interfaces those codes use, or what assumptions on indexing are made).

The only place this ambiguity of indexing appeared is when a checkpoint file was written or read.  It doesn't appear that anything else was affected.

This will require testing by @marchdf and possibly others.

### DIFF
--- a/glue-codes/fast-farm/src/FASTWrapper.f90
+++ b/glue-codes/fast-farm/src/FASTWrapper.f90
@@ -139,7 +139,7 @@ SUBROUTINE FWrap_Init( InitInp, u, p, x, xd, z, OtherState, y, m, Interval, Init
       
       end if
          !.... multi-turbine options ....
-      ExternInitData%TurbineID = InitInp%TurbNum
+      ExternInitData%TurbIDforName = InitInp%TurbNum
       ExternInitData%TurbinePos = InitInp%p_ref_Turbine
       ExternInitData%WaveFieldMod = InitInp%WaveFieldMod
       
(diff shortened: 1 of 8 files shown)


deslaughter (approved):

marchdf:
I can confirm that the files are closed with these changes and interfacing with amr-wind. Thank you!

Issue #1851, opened 2023-11-02
Docker images


andrew-platt:
**Is your feature request related to a problem? Please describe.**
It would be nice to add docker images of our releases and host them on the NREL dockerhub (https://hub.docker.com/u/nrel).

**Describe the solution you'd like**
Ideally these would be built using a really stripped down linux version that is as small as possible (a stripped down Alpine linux for example). Also ideally build with the OneAPI compiler set.  This should of course be automated (@mayankchetan has done some automation of this already).

**Describe alternatives you've considered**
None. 

**Additional context**
I'm not entirely certain how many users would be interested in this, so perhaps this isn't a really high priority.

cortadocodes:
I'm very much interested in this. The use case is to allow creating scalable openfast instances in the cloud so we can run any number of openfast analyses in parallel for a low cost. Docker is essential for this.

So far, I created a working Dockerfile [here](https://github.com/octue/openfast-docker) whose image is uploaded to Docker Hub. It was quite hard to get working as I had difficulties getting conda to give me a specific version of openfast (getting the latest version works fine) which is important for allowing other engineers to create tools that use openfast without breaking them each time a new version of openfast is uploaded to the conda repo. 

I think working together on this would be beneficial!

andrew-platt:
Thanks for your interest in this idea!  I definitely agree that we should collaborate on this as well as the openfast_toolbox.

One thing I haven't thought through is where to host containerization recipes.  I had been previously thinking they would just be added into the OpenFAST repository, but perhaps a separate repository as you have done is a more logical method.

@deslaughter, @mayankchetan, @rafmudaf -- do any of you have thoughts on this, or the best way to work with @cortadocodes  and @Octue on this?

rafmudaf:
If there's any need to have the Dockerfiles match a particular OpenFAST version, I recommend to keep them in the OpenFAST repo. There are already some [here](https://github.com/OpenFAST/openfast/tree/main/share/docker). Being in the same repository as the code, you could automatically update them on a release or at least have a release checklist to remind you to update them.

mayankchetan:
@cortadocodes, thank you for your interest in Dockerizing OpenFAST! 

[...]

PR #2124, opened 2024-03-27
Update dockerfile and facilitate `nrel/openfast` Docker Hub registry

cortadocodes:
## Summary
This pull request updates the existing `Dockerfile` to facilitate the building of production-ready docker images of OpenFAST which can be pushed to a Docker Hub image repository and used by the community.

## Changes
- Add build args for the base image, CMake options, Make build cores, and timezone
- Minimise the final image size (down from multiple GB to ~200MB)
- Add a manually-triggered GitHub Actions workflow to build specific versions of OpenFAST for the `linux/amd64` and `linux/aarch64` architectures and push them to a Docker Hub repository
- Add a readme explaining how the new `Dockerfile` can be used

## Related issue
https://github.com/OpenFAST/openfast/issues/1851

## Impacted areas of the software
Only the `Dockerfile` in `share/docker/openfast_ubuntu/Dockerfile` and the GitHub Actions workflows

## Test results
- I've locally built OpenFAST images for versions:
  - `3.1.0` - I then ran it as a container, successfully ran the `openfast` CLI command with no parameters, and ran a successful analysis with some test input files
  - `3.5.2` - I then ran it as a container and successfully ran the `openfast` CLI command with no parameters
- I've used the new GitHub workflow to build and push `3.1.0` and `3.5.2` images to the `octue/openfast` Docker Hub repository

## Pre-release to-dos
If the maintainers are happy with this pull request, we need to do the following things before merging:
- NREL team:
  - [ ] Create the `nrel/openfast` docker repo
  - [ ] Add docker secrets for `nrel/openfast` access to this repository's GitHub actions secrets
  - [ ] Check the updated copyright notice in the `Dockerfile` is correct

- @cortadocodes
  - [ ] Update `octue` to `nrel` in docker workflow (docker repo and location of dockerfile)
  - [ ] Move `Dockerfile` up one level (this will remove the diff that's currently visible for the file, which is why I haven't moved it yet)
  - [x] Document the Dockerfile and docker images outside of the readme I've added


cortadocodes:
> Also, should we add an automated test to GH actions for building docker (if that's possible)?

@andrew-platt that's no problem. I can write one that will automatically build and push to the docker registry on merge of a pull request into the `main` branch. Did you want it to do anything else apart from that?

deslaughter:
@andrew-platt I'm good with including this in 3.5.3. When you say add a test to GH actions, are you talking about seeing if it fails to build the image? I think @cortadocodes is talking about triggering an image build when pushing to main, but GH won't check to see if it builds.

@cortadocodes how will we know if there are docker image build failures?

andrew-platt:
@deslaughter, correct.  I think it would be good to do some automated build tests to make sure images build as expected.  That way we can catch issues before a final merge to main.

@cortadocodes, an automated image build for dockerhub on merges to main would be helpful. That way we wouldn't need to manually trigger that on releases.

cortadocodes:
@deslaughter we can do either or both

We'll know if the build fails because the workflow will show as a failure. We can even add a badge to the readme (or anywhere else on the internet) that shows if the latest build was successful. 

If you wanted to, we could automate GitHub releases of this repository and only release if the docker image builds on merge into `main` (and any other tests you want to run pass)

cortadocodes:
@deslaughter @andrew-platt

It sounds like this might be the set of workflows we want then:
- Pull request merge into `main`: build and push docker image for both architectures
- Every push to a release candidate branch: test building (but don't push) docker image for `linux/amd64`

For context, building from scratch for `linux/amd64` took a few minutes while `linux/aarch64` took 3 hours on GitHub actions. To avoid long waits and extra compute time/emissions for the docker build testing, I've suggested not testing building on non-release-candidate branches or multiple architectures.

andrew-platt:
that's a really long build time for `linux/aarch64`.  Testing just the `linux/amd64` would be sufficient quick verification that we probably didn't break anything.

share/docker/openfast_ubuntu/Dockerfile
```diff
@@ -13,42 +12,63 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+ARG BASE=ubuntu:mantic
```
andrew-platt, line 15: Should we stick with LTS releases (Jammy) instead of shorter support ones?

share/docker/openfast_ubuntu/README.md
```diff
@@ -0,0 +1,42 @@
+# OpenFAST docker images
```
andrew-platt, line 1: I wonder if we should put this into the `docs` (or link it into there) so that it shows up on readthedocs?

On the other hand, it is really nice to have right here with the dockerfiles.

share/docker/openfast_ubuntu/Dockerfile
cortadocodes: Yeah that makes sense. I'll switch to `jammy` tomorrow and verify that it builds

share/docker/openfast_ubuntu/README.md
cortadocodes: I think it's good to have it here next to the dockerfile so it's harder to forget to update when changes are made. I think linking from readthedocs is a good idea too as we only have to update it in one place.

deslaughter: I think linking to the README.md file from the docs is the way to go.

share/docker/openfast_ubuntu/Dockerfile
cortadocodes: The image builds fine with `jammy` although I've just noticed that the version `openfast` reports when I build from the `v3.5.2` tag is ` OpenFAST-v3.5.2-dirty` - any idea why it says it's dirty?

share/docker/openfast_ubuntu/README.md
cortadocodes: Docs updated and readme linked

cortadocodes:
I'm fixing a small issue with the test and automatic docker workflows; I'll confirm when it's done

cortadocodes:
> I'm fixing a small issue with the test and automatic docker workflows; I'll confirm when it's done

This is fixed now so I think that's everything ready!

share/docker/openfast_ubuntu/Dockerfile
andrew-platt: We haven't set the tag for 3.5.3 yet -- we do that after the merge to main.  I wonder if this will cause an issue for correctly setting the build info on automated builds after merges to main if the latest tag doesn't yet exist.

Not sure why it shows as dirty and doesn't include the number of commits since 3.5.2

cortadocodes: If the release candidate branches are always named `rc-major.minor.patch`, the tag for the docker image should be fine as I'm extracting the version from the branch name. However:
- The `openfast` CLI in the docker image won't show the version on the command line as (I think) it extracts it from the git ref
- The docker tag will be wrong if you decide to use a different git tag after merging

We could fix this by adding a `VERSION` file or something similar to `package.json` or `pyproject.toml`. You'd update it with the new version number for openfast before merging. The docker workflow would then use this for the docker tag and the openfast CLI would also need to to show the version.

Doing this would also open up the possibility for automating tags/releases with a workflow so the git tag is always created before the docker image is built and you don't have to manually release. You could either add the release notes as usual or use some nice automation to pull the release notes from somewhere else (like the pull request description or commits)

andrew-platt: Good thoughts!  We've typically been doing release merges directly from `dev`, but that would be simple to always do an `rc-x.y.z` branch first for the release.  We can probably tag directly from the last commit on the `rc` branch prior to merging to `main`.  This approach could in theory eliminate the issue of getting the correct tag into docker.

cortadocodes: Yeah that would work I think! As long as you're happy with the tag being on the commit before the merge commit on `main`

andrew-platt: I think it is acceptable.  We'll do a merge from `main` into the `rc-x.y.z` branch right before the final merge to `main`.  I suspect we'll end up with the same git commit hash on `main` that way.

share/docker/openfast_ubuntu/Dockerfile
```diff
@@ -13,42 +12,63 @@
...
+    apt-get install --no-install-recommends -y \
+        software-properties-common \
+        build-essential \
+        cmake  \
+        cmake-curses-gui \
+        gcc \
+        gfortran \
+        git \
+        libblas-dev \
+        liblapack-dev \
```
deslaughter, line 37: The OpenBLAS library has better performance than the native libraries in my experience.

```suggestion
        libopenblas-dev \
```

share/docker/openfast_ubuntu/Dockerfile
```diff
@@ -13,42 +12,63 @@
...
 WORKDIR /openfast/build
 
 # NOTE: building with optimizations on (RELEASE or RELWITHDEBINFO), the virtual machine
-# will require about 6GB of memoery. Otherwise, the gfortran compiler will exit with an
+# will require about 6GB of memory. Otherwise, the gfortran compiler will exit with an
 # "internal error"
-RUN cmake .. -DBUILD_TESTING=ON -DDOUBLE_PRECISION=ON -DCMAKE_BUILD_TYPE=RELWITHDEBINFO
-RUN make -j4 install
+ENV FC=/usr/bin/gfortran
+ARG CMAKE_OPTIONS="-DBUILD_TESTING=OFF -DDOUBLE_PRECISION=ON -DCMAKE_BUILD_TYPE=RELEASE"
```
deslaughter, line 53: It might be worthwhile to go ahead and add FAST.Farm, though then we might need to pull in the OpenMP libraries.

.github/workflows/build-docker-image-manual.yml
```diff
@@ -0,0 +1,58 @@
...
+  build-and-push:
+    runs-on: ubuntu-latest
+    timeout-minutes: 300
+    env:
+      DOCKERFILE_PATH: share/docker/openfast_ubuntu/Dockerfile
+      DOCKERFILE_PERMALINK: https://raw.githubusercontent.com/octue/openfast/update-dockerfiles/share/docker/openfast_ubuntu/Dockerfile
+      DOCKERHUB_REPOSITORY: octue/openfast
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
andrew-platt, line 27: Should this be updated to `@v5`?

share/docker/openfast_ubuntu/Dockerfile
andrew-platt: After chatting with @deslaughter, I think it makes sense to put FAST.Farm in this container as well.  It wouldn't have OpenMP so it would be of limited value for production use, but could be very useful for troubleshooting a setup before sending it to a cluster to run (FAST.Farm really should be compiled with Intel for production usage using native libraries on the cluster).

andrew-platt:
Whoops!  I didn't mean to merge this just yet (and didn't intend to do it as a rebase merge).  We can make some PR's to resolve the last questions:

- [ ] Include FAST.Farm
- [ ] openblas library (comment above from @deslaughter)

share/docker/openfast_ubuntu/Dockerfile
cortadocodes: I'll update this now

cortadocodes: Fixed by adding `libgomp1` to the final image!

cortadocodes:
@andrew-platt are you able to re-open this pull request so the new commits show up?

andrew-platt:
@cortadocodes, unfortunately there isn't a good way to re-open the PR.  We'll need to do a second PR with the updates (thanks for figuring those out!).

Again, my apologies for accidentally merging this too early.

cortadocodes:
No problem!

cortadocodes:
@andrew-platt here's the new PR: https://github.com/OpenFAST/openfast/pull/2139

PR #2133, opened 2024-04-01
Remove linking of implicit Fortran libraries

jrood-nrel:
This causes problems when linking to OpenFAST downstream as it adds numerous `-l` arguments when linking to OpenFAST which are unable to resolve. Using `-l` is prone to issues in general. It does not appear to be necessary, unless someone can explain why it's necessary and how a project should robustly handle the effects of this downstream.


andrew-platt:
@deslaughter, let's merge this to 3.5.3, then to dev from there.

deslaughter (approved):
I think this should be removed if it's causing linker issues with other software. I'm pretty sure that I added this to resolve an issue with building the Simulink library, but that's being test by Github Actions and appears to be working, so it seems these libraries are no longer needed.

I agree with @andrew-platt that this should be merged into the `rc-3.5.3` branch first, and then into `dev`

PR #2145, opened 2024-04-05
Docs/update for 3.5.3 release

andrew-platt:
Ready to merge

- Additional language regarding Docker containers from the GitHub container repository
- update changelog for previous couple of PRs

Follow up documentation related to #2141

Documentation only.

### DIFF
--- a/docs/source/install/index.rst
+++ b/docs/source/install/index.rst
@@ -157,26 +157,57 @@ containing the executables, and running a simple test command:
 
 .. _use_docker:
 
-Use a docker image
-~~~~~~~~~~~~~~~~~~
-Multiple versions of OpenFAST are available as docker images from our `docker registry <https://hub.docker.com/r/nrel/openfast>`_.
+Running OpenFAST with docker
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+OpenFAST is avilable to be run on docker starting with version 3.5.3. Three approaches are shared below.
+
+Using a docker image from Docker hub
+------------------------------------
+Multiple versions of OpenFAST are also available as docker images from our `docker registry <https://hub.docker.com/r/nrel/openfast>`_.
 To pull and run one with files from your local machine available, run:
 
 .. code-block:: shell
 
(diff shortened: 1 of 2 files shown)


docs/source/install/index.rst
```diff
@@ -157,27 +157,58 @@ containing the executables, and running a simple test command:
...
 .. _use_docker:
 
-Use a docker image
-~~~~~~~~~~~~~~~~~~
-Multiple versions of OpenFAST are available as docker images from our `docker registry <https://hub.docker.com/r/nrel/openfast>`_.
+Running OpenFAST with docker
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+OpenFAST is avilable to be run on docker starting with version 3.5.3. Three approaches are shared below.
+
+Using a docker image from Docker hub
```
mayankchetan, line 164: Do we add that it'll be available soon? and not in this specific release?

docs/source/install/index.rst
```diff
@@ -157,27 +157,58 @@ containing the executables, and running a simple test command:
...
+local directory as the input files. 
+
+You can also run commands interactively inside the container with:
 
 .. code-block:: shell
 
-    docker run --rm -it --volume=/path/to/files:/files nrel/openfast:3.5.2 /bin/bash
+    docker run --rm -it --volume=/path/to/files:/files nrel/openfast:3.5.3 /bin/bash
+
+To pull the latest container, substitute `latest` for `3.5.3` in the above
```
mayankchetan, line 182: I think it would be better to use `latest` in the docker command examples,  so that we don't have to update the docs at every release. Could write it as:

> To pull a container corresponding to a specific version of OpenFAST, substitute the specific version (eg: `3.5.3`) for `latest` in the above commands.

deslaughter (approved):

docs/source/install/index.rst
andrew-platt: I thought about it, but I don't now if I'll remember we need to change that in a later release.

docs/source/install/index.rst
andrew-platt: done!

PR #2173, opened 2024-04-23
Fix crash in MAP_End when using Intel's new icx compiler and disable caching in setup-python GH action

deslaughter:

This PR provides two fixes:
1. Fixes a crash in MAP during module shutdown (`MAP_End`) caused by double freeing memory related to strings. This crash only occurred when using the new Intel icx compiler.
2. Adjusting the settings in the `setup-python` Github Action to disable caching of Python modules. This should fix the issue where regression tests fail because `setup-python` can't find `requirements.txt`. 

PR #2129 was a previous attempt to fix the issue with `setup-python` and the regression tests which didn't succeed. 

- MAP
- Github Actions

The issue with MAP trying to free the memory twice looks like it has been around since the module was added. The problem stemmed from several nodes being added to the `OutputList` by directly copying the data. These nodes contained pointers to strings which were allocated. These pointers were shared with nodes in other lists which were freed earlier in `MAP_End`. When the `OutputList` nodes were freed, the code tried to free the same pointers again, resulting in a crash. The issue was resolved by creating new nodes in `OutputList` with a copy of the data instead of reusing the existing pointers. The amount of data copied is very small so this was a simpler solution than trying to track the pointer usage.

This fix will need to be migrated to the other branches once approved.

### DIFF
--- a/.github/workflows/automated-dev-tests.yml
+++ b/.github/workflows/automated-dev-tests.yml
@@ -43,7 +43,7 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: '3.11'
-          cache: 'pip'
+          # cache: 'pip'
       - name: Install dependencies
         run: |
           pip install -r requirements.txt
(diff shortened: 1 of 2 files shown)


andrew-platt (approved):
Thanks for fixing this!

PR #2182, opened 2024-04-30
Move test seastate_wavemod5 to after test seastate_wr_kin1

andrew-platt:
Ready to merge 

- seastate_wr_kin1 outputs are used by seastate_wavemod5, so seastate_wavemod5 would often fail in GH actions if it appeared earlier in the list
- Also set to only run 4 seastate tests in parrallel to give more time between these two tests

Testing for _SeaState_ module tests only


deslaughter (approved):
This must have gotten added after reducing the number of parallel CTest runs, 7 is a crazy number

Issue #2119, opened 2024-03-26
Misnamed NWTC_Library Registry files in Visual Studio project files in dev-unstable-branch 


bjonkman:
The Visual Studio project files that contain NWTC_Library need to be updated with the new names of the Registry input files in the `dev-unstable-pointers` branch. They currently reference `..\..\modules\nwtc-library\src\Registry_NWTC_Library_typedef_nomesh.txt` and `..\..\modules\nwtc-library\src\Registry_NWTC_Library.txt`, but these files have been renamed in the recent history. When doing a rebuild, where it deletes the `*_Types.f90` files, it fails because the registry is trying to recreate `NWTC_Library_Types.f90` from one of these missing files.

See for example: https://github.com/OpenFAST/openfast/blob/3e49c6797be33526f6414018d4fd22841047aa5c/vs-build/FASTlib/FASTlib.vfproj#L2038

andrew-platt:
I believe this was fixed with #2189

PR #2189, opened 2024-05-07
More fixes for Visual Studio

bjonkman:
Fixes a few more issues with Visual Studio in the dev-unstable-pointers branch, particularly with running the registry on the pre-build step. A few dependencies were missing and files were improperly named.

This fixes a couple of places referred to in https://github.com/OpenFAST/openfast/issues/2119


andrew-platt (approved):
Thanks for fixing these!  It would have been a while before I discovered this was an issue.

PR #158, opened 2018-08-18
Reopening output ascii text file and appending on restart

gantech:

PR #2219, opened 2024-05-15
Turn off "#Restoring here" messages in .out files during linearization visualization sims

andrew-platt:
Ready to merge

When mode shape visualization is used during linearization, the `.out` file has lines with `#Restarting here` at each linearization.  This creates issues with some readers.

The `#Restarting here` comment was added with #158 for use with AMR-Wind.  Unfortunately this also affected the mode shape visualization restarts.

ASCII output files from mode shape visualization during linearization only.

### DIFF
--- a/modules/openfast-library/src/FAST_Subs.f90
+++ b/modules/openfast-library/src/FAST_Subs.f90
@@ -10212,8 +10212,8 @@ SUBROUTINE FAST_RestoreFromCheckpoint_T(t_initial, n_t_global, NumTurbines, Turb
    ! deal with files that were open:
    IF (Turbine%p_FAST%WrTxtOutFile) THEN
       CALL OpenFunkFileAppend ( Turbine%y_FAST%UnOu, TRIM(Turbine%p_FAST%OutFileRoot)//'.out', ErrStat2, ErrMsg2)
-      IF ( ErrStat2 >= AbortErrLev ) RETURN
       CALL SetErrStat(ErrStat2, ErrMsg2, ErrStat, ErrMsg, RoutineName )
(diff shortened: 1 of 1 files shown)


modules/openfast-library/src/FAST_Subs.f90
```diff
@@ -10212,10 +10218,14 @@ SUBROUTINE FAST_RestoreFromCheckpoint_T(t_initial, n_t_global, NumTurbines, Turb
...
    IF (Turbine%p_FAST%WrTxtOutFile) THEN
       CALL OpenFunkFileAppend ( Turbine%y_FAST%UnOu, TRIM(Turbine%p_FAST%OutFileRoot)//'.out', ErrStat2, ErrMsg2)
-      IF ( ErrStat2 >= AbortErrLev ) RETURN
       CALL SetErrStat(ErrStat2, ErrMsg2, ErrStat, ErrMsg, RoutineName )
-      CALL WrFileNR ( Turbine%y_FAST%UnOu, '#Restarting here')
-      WRITE(Turbine%y_FAST%UnOu, '()')
+      IF ( ErrStat2 >= AbortErrLev ) RETURN
+      if (present(silent)) then
+         if (silent) then
+            CALL WrFileNR ( Turbine%y_FAST%UnOu, '#Restarting here')
```
bjonkman, line 10225: I was expecting this to be written when **not** silent. Seems counterintuitive to write extra text when silent is true.

andrew-platt: Whoops!  I shouldn't commit code in a hurry... Thanks for catching that.

PR #2223, opened 2024-05-16
nwtc lib: SetErrStat ref'd incorrectly in VTK

andrew-platt:
Ready to merge

The `SetErrStat` was moved to `NWTC_Base.f90` from `NWTC_IO.f90`.  This was not updated in `VTK.f90`.  Somehow GH actions and all the setups I was using for compiling did not error on this.

NA

Compilation on a few systems.

Thanks to @RyanDavies19 for finding this!

### DIFF
--- a/modules/nwtc-library/src/VTK.f90
+++ b/modules/nwtc-library/src/VTK.f90
@@ -5,10 +5,9 @@
 module VTK
 
    use Precision, only: IntKi, SiKi, ReKi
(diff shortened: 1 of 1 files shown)


deslaughter (approved):
It probably worked because `NWTC_IO` uses `SubSys` which uses `NWTC_Base` which has `SetErrStat`. gfortran must have allowed it to be exposed up through the chain. Yet another unintended consequence of the tight coupling changes.

PR #2226, opened 2024-05-19
Fix turbine indexing on ExtLoads Restart

ndevelder:
Ready to be merged...this bug was causing segfault on restart with fsi-connected Exawind restarts.

Simply takes the recipe from functions above and applies the new turbine indexing,

Original issue here: https://github.com/OpenFAST/openfast/issues/2096

Should only impact Exawind fsi cases.


andrew-platt (approved):
Thanks for fixing this!

andrew-platt:
I missed that routine when I wrote #2097.  Thanks for fixing it!

PR #2235, opened 2024-05-23
Checkpoint files possibly not created with correct name from fsi runs

ndevelder:
Exploring the weirdness with checkpoint files on non-precursor checkpoint creation. 

Problem:
During the precursor openfastcpp run, the code knows about all the turbines and so can properly write out turbine ids and openfast filenames, but this doesn't happen on subsequent restarts when each nalu instance only knows about a single turbine.


ndevelder:
Opened against wrong remote

andrew-platt:
> Opened against wrong remote

For future reference, it is possible to change the target branch on an open pull request.

PR #2236, opened 2024-05-23
Send actual filename to checkpoint creator

ndevelder:
Exploring the weirdness with checkpoint files on non-precursor checkpoint creation.

Problem:
During the precursor openfastcpp run, the code knows about all the turbines and so can properly write out turbine ids and openfast filenames, but this doesn't happen on subsequent restarts when each nalu instance only knows about a single turbine.


ndevelder:
It seems as though this is possibly also an issue with restart frequency...Lawrence is exploring. We'll update here when we find out. This is as-yet untested PR.

ndevelder:
Closing and will add new PR soon. I think this is actually an issue with the turbine index being passed to the write function in lieu of TurbID.

PR #2245, opened 2024-05-30
Fix restart file index and nc write

ndevelder:
Trying to fix the restart of fsi simulations from precursor when there are multiple separate turbines (and hoping this contributes to being able to restart multiple turbines)


andrew-platt:
Is this PR still needed?  @deslaughter, or is this covered in your fix for FSI coupling?

deslaughter:
@ndevelder PR #2365, merged into `dev` today, resolves the issue with the OpenFAST restart when using the C++ interface. Do you still want this change merged? 

@gantech I don't understand the change in this PR, does it look right to you?

ndevelder:
I'm pretty sure this is important for properly indexing the restart files, using "turb_id" as set in the input file as opposed to the index pulled out from the turbines glob. I do have these lines with this change in my current_fsi branch that we've been using.

deslaughter (approved):

PR #1429, opened 2023-02-01
AeroDyn14 removal

andrew-platt:
This PR is **not** ready for merging (regression tests must be dealt with first).

**Feature or improvement description**
The AeroDyn14 module was a carry-over from FAST v7 to FAST v8 as a stop-gap while AeroDyn15 was under development.  As such AeroDyn14 was never converted to the FAST Modular Framework; this created many code workarounds in ElastoDyn and in the openfast-library to get it to work.  At this point AeroDyn15 is far more comprehensive than AeroDyn14 so very few users ever use it. Therefore the plan for many years has been to remove it entirely, and now we finally got around to it.

**Related issue, if one exists**
#527 -- bug with AD14 tower
#93 -- deprecating AD14
https://github.com/OpenFAST/r-test/issues/43

**Impacted areas of the software**
There are some significant improvements to the glue code and ElastoDyn with the removal of AD14.

Several code development projects are depending on this PR.

- Blade resolved loads from CFD (alternative to AeroDyn, `CompAero==3`)
- AeroDisk (will take option `CompAero==1`)
- Tight coupling solver
- InflowWind data restructuring

**Additional supporting information**
The regression tests that previously used AeroDyn14 have been updated to AeroDyn15, but have not been tuned correctly for the DBEMT parameters (in particular the `Tau` time constants).  Some of these cases also show instabilities and had been turned off as a result.  We need to decide which cases we will simply remove, and which we will keep.

**Test results, if applicable**
Regression tests with AD14 converted to AD15, but not tuned correctly yet.

Issue #527, opened 2020-08-14
AeroDyn14 NewTower TwrCD interpolation


rwegnerGermany:
Hi,
I am using Simpack with AeroDyn v13 but looking at the current code the problem appears to also exist in AeroDyn v14.
I modeled a hybrid lattice tower with the NEWTOWER feature with 50 tower stations (NTwrHt) mapping to 23 different (NTwrCD) drag coefficients.
With a parked rotor at constant 4m/s with no wind shear I got very strange wind speeds at the blade elements:

- Tower dam model can produce highly negative x-direction wind speeds
- Tower dam model affects blade elements far outside of the tower center line and above the tower
- Wind speeds at the blade elements depend on height (for this test the 50 stations had the same diameter and used the same index in TwrCD vector)

This blade element is at the 6 o'clock moving slowly upwards vertically (rotor hangs on a crane). At 950s the elevation is constant.
![WindX](https://user-images.githubusercontent.com/69688683/90271284-d713a300-de5b-11ea-9ad6-2a6642461935.png)

**I think I found the bug:** The variable NTwrCDCol is basically never used after filling it. In AeroSubs.f90 line 2562
`TwrElCD = InterpStp( TwrElHt, p%TwrProps%TwrHtFr, p%TwrProps%TwrCD(1,:), N2, p%TwrProps%NTwrHt )`
the vectors TwrHtFr and TwrCD are of different length. To my understanding the procedure should be:

1. Use the tower station table to get a TwrCD **index** for the height in question
2. Use this index for NTwrCDCol and return the drag coefficient.

It appears to work if NTwrCD = NTwrHt and the TwrCD values are in the order of the tower table. Just like it is in AeroDyn v15.

andrew-platt:
Closing this issue as we won't be able to fix it before deprecating AD14.

Issue #93, opened 2018-02-20
Deprecating AeroDyn 14


rafmudaf:
It has been mentioned recently that AeroDyn 14 may be phased out in favor of AeroDyn 15. This issue is intended to start that conversation.

@jjonkman Is there anything that is in AeroDyn 14 but not in AeroDyn 15?

jjonkman:
The big things missing from AeroDyn v15 that are in AeroDyn v14 are (1) Generalized Dynamic Wake (GDW) and (2) support for Yujia's Dynamic Wake Meandering (DWM) code.

Regarding (1), we collaborated with Envision Energy to implement and test Dynamic Blade Element/Momentum Theory (DBEMT).  It was intended that DBEMT in AeroDyn v15 will replace the need for GDW.  My understanding is the Envision Energy has given NREL a branch of AeroDyn v15 with DBEMT that NREL needs to review and merge into OpenFAST.

Regarding (2), Yujia's DWM code will be replaced with FAST.Farm, which has been developed as a branch of OpenFAST.  We are currently validating FAST.Farm against SOWFA in preparation for the initial public release.

So, to me, once DBEMT and FAST.Farm have been included in OpenFAST, we can eliminate AeroDyn v14 from OpenFAST.

rafmudaf:
This is resolved by https://github.com/OpenFAST/openfast/pull/754/commits/7c35b9493b2929ed0d666d9d6e903ba155a767d3 in v3.0.0.

rafmudaf:
To clarify, https://github.com/OpenFAST/openfast/commit/7c35b9493b2929ed0d666d9d6e903ba155a767d3 adds a deprecation warning and notes that AD14 will be removed in a future release. Therefore, let's leave this issue open until AD14 is removed from OpenFAST.

andrew-platt:
Completed in 4.0.0 (#2586)

PR #2267, opened 2024-06-12
AeroDyn14 removal

andrew-platt:
This PR is **not** ready for merging (regression tests must be dealt with first).  This PR supercedes #1429 (that PR sat for too long).

The AeroDyn14 module was a carry-over from FAST v7 to FAST v8 as a stop-gap while AeroDyn15 was under development.  As such AeroDyn14 was never converted to the FAST Modular Framework; this created many code workarounds in ElastoDyn and in the openfast-library to get it to work.  At this point AeroDyn15 is far more comprehensive than AeroDyn14 so very few users ever use it. Therefore the plan for many years has been to remove it entirely, and now we finally got around to it.

Also includes minor fixes for AD15 input files:

- Updated all AD15 input files to include the `UA_Mod` comments instead of the deprecated `AFAeroMod` comments
- Updated all AD15 input files to include all lines (some lines are optional)

#527 -- bug with AD14 tower
#93 -- deprecating AD14
https://github.com/OpenFAST/r-test/issues/43

There are some significant improvements to the glue code and ElastoDyn with the removal of AD14.

The regression tests that previously used AeroDyn14 have been updated to AeroDyn15, but have not been tuned correctly for the DBEMT parameters (in particular the `Tau` time constants).  Some of these cases also show instabilities and had been turned off as a result.  We need to decide which cases we will simply remove, and which we will keep.

Regression tests with AD14 converted to AD15.


andrew-platt:
This PR also removes `PitchAxis` input column from the ED blade file (backwards compatible with older format).

PR #2280, opened 2024-06-24
MD: ramp inertial loads during startup transients

andrew-platt:
This is a draft PR so that I don't lose track of it.

When the inertial loads are used in the _MoorDyn_ loads and moments returned to OpenFAST, the simulation may fail immediately due to very large loads from MD during the initial transients.  One method to get around this is by adding a ramping of the inertial loads in MD.  This PR adds a 30 second ramp time (arbitrarily chosen ramp time, hard coded in the MD registry for now).

@RBergua is testing this feature.

_MoorDyn_

It would be useful to add the ramping time for inertial loads as an input option in MD rather than hard coding it

None


andrew-platt:
Ramp inputs added as optional inputs in the `Solver Options` section.  In the input file, add the following:
```
2 inertialF - Toggle to include inertial components in the returned forces from coupled bodies and rods. Transients in the acceleration passed into MoorDy-F by OpenFAST can result in large non-physical forces and moments which can cause instability in the model [0: no, 1: yes, 2: yes with ramp to inertialF_rampT] (default =0)

30.0 inertialF_rampT - Ramp time for inertial forces to reduce coupled object instability (s). This is ignored unless inertialF = 2 (default =30)
```

Documentation updated here: https://moordyn.readthedocs.io/en/latest/inputs.html#id4

RBergua:
We are working with the USFLOWT design that has large buoyant cans:
image

Often times, OpenFAST crashes at the beginning of the simulation due to a large acceleration passed to MoorDyn at the beginning of the simulation due to the initial transient.
image

Before, we were disregarding the inertial loading due to the platform acceleration in MoorDyn by defining inertialF = 0. However, this is less than ideal because it gets rid off part of the physics for the whole simulation. 

I confirm that by using inertialF = 2 and giving a ramp up time for the platform accelerations passed to MoorDyn, it's possible to avoid OpenFAST crashing at the beginning of the simulation and considering the proper accelerations after the ramp time.
image
image

RyanDavies19:
All looks good on my end, thanks for putting this together. Glad it is helping with the start-up transient problem.

PR #1295, opened 2022-10-21
Reduced order models: AeroDisk and Simplified-ElastoDyn (SED)

andrew-platt:
This PR is *NOT* ready for merging.

**Feature or improvement description**
Two new modules are added for low fidelity modeling of aerodynamics and structure.  These modules are intended for use as very low fidelity turbines in a FAST.Farm simulation in order to speed up the overall simulation.  In such a use case, these might be placed in the front few rows of a FAST.Farm simulation where only the turbines in rows further back are of interest, but the wake effects from the leading turbines are needed.

1. AeroDisk -- This module provides a very simple disk actuator model for the aerodynamics as an alternative to AeroDyn 15 and AeroDyn 14.
2. Simplified-ElastoDyn (SED) -- This module provides a very simple 2 DOF elastic model that only includes yaw and generator speed.  This module is not compatible with SubDyn, HydroDyn, or any mooring codes.

**Impacted areas of the software**

- glue code changes
    - new mesh mappings are introduced to map loads and motions from ADsk and SED to other modules
    - additional logic to accommodate the new modules
    - minor code restructuring as needed
    - checks to ensure SED and ADsk are not used with certain combinations that are not allowed
- FAST.Farm changes
    - Minor updates to FAST_Wrapper to pass correct Cp and Ct values

**Additional supporting information**

**Test results, if applicable**
Three new OpenFAST test cases are added as modifications of the `5MW_Land_DLL_WTurb` case:

- `5MW_Land_DLL_WTurb_SED` -- SED instead of ED
- `5MW_Land_DLL_WTurb_ADsk` -- AeroDisk module instead of AD15
- `5MW_Land_DLL_WTurb_ADsk_SED` -- AeroDisk instead of AD15, and SED instead of ED

New module tests are also included for SED and ADsk.

**To complete**

- [x] Verify correct SED and ADsk behaviour after merging dev
- [x] ~~Phase shift in GenPwr for `5MW_Land_DLL_WTurb_ADsk_SED` case (possibly mesh issue in glue code?)~~
        This was an input file issue
- [x] ~~SED + AD15 can result in tower strike event (possible mesh issue, or maybe setup -- not investigated yet)~~
        This was a setup issue
- [x] Disable blade flexibility in ElastoDyn when ADsk is used with ED
- [x] Add VS build projects for ADsk and SED drivers
- [x] Add a simple FAST.Farm case with a simplified turbine
- [x] add OpenMP directive in AWAE module (high res)
[...]

PR #2283, opened 2024-06-26
Aero modifications

bjonkman:
This pull request 
- makes some modifications to UAMod 5 for use in full 360-degree angles of attack.
- uses a weighted-average of the axial induction for DBEMT input instead of just a uniform average of axial induction at all the nodes.
- allows the user to change the `IntegrationMethod` used in UA (UA_Mod 4, 5, etc)
- allows the user to have the code calculate CantAngle (if `BlCrvAng` is not specified in the input file). These calculated values are printed in the summary file.
- caps the Ct, Cp, and Cq outputs if their denominators ever get too small
- implements the code around `BldNd_BlOutNd` in AeroDyn, allowing the user to specify which nodes they would like to output.
- fixes an issue where the AeroDyn echo file would not include data read from the blade property files.
- reports an error if one exists in the BEMT initialization of states (likely rare)
- updates some calculations of the UA parameters in AFI initialization
- removes non-standard tabs from MoorDyn
- changes `TwrAero` from a logical to integer internally (the input file still recognizes the logical value)
- simplifies some logic around the buoyancy calculations

AeroDyn: DBEMT, UA (UA_Mod 4, 5, or 6)

Failing cases:
- rtest-module-drivers (1 failing test)
  - `ad_BAR_SineMotion_UA4_DBEMT3`
      - There are very small difference in the two failing channels (the Fx and Fy loads on a single node):
image
image

- rtest-OF (1 failing test)
   - `MHK_RM1_Fixed`
      - The differences in the two failing channels are minor:
image
image


modules/aerodyn/src/FVW.f90
```diff
@@ -1572,8 +1572,9 @@ subroutine UA_Init_Wrapper(AFInfo, InitInp, interval, p, x, xd, OtherState, m, E
    type(FVW_MiscVarType),           intent(inout)  :: m           !< Initial misc/optimization variables
    integer(IntKi),                  intent(  out)  :: ErrStat     !< Error status of the operation
    character(*),                    intent(  out)  :: ErrMsg      !< Error message if ErrStat /= ErrID_None
+   
+   integer(IntKi), parameter :: NumBladesPerWing = 1 !bjj why are we using this number?
```
andrew-platt, line 1576: Good question.  I don't know why we are allocating these arrays as 2D.

Maybe there was a reason early in the development that is now irrelevant?

andrew-platt:
Thanks for adding the `BldNd_BlOutNd` code updates!  That will be very useful!

andrew-platt:
To do (thinking about this because I'm dealing with AD15 input files for AD14 removal PR):

- [x] update AD15 input files with new nodal output comments
- [ ] update AD15 input files with Solver and other new lines (I know these are optional, but want to get a fully consistent set of input files for the release) -- @andrew-platt

modules/aerodyn/src/FVW.f90
ebranlard: This is introduced because UA is implemented for many blades at once, whereas OLAF has a list of wings, and we call UA for each of the wing.  The parameter is introduced instead of putting "1" in the source code. It's supposed to confuse less the reader, but it seems it was not too clear :) 

Maybe we can add in the comment: 
```fortran
NumBladesPerWing !< UA is for multiple blades, FVW calls one instance of UA per wing. 
```

modules/aerodyn/src/AeroDyn_IO.f90
```diff
@@ -344,15 +353,14 @@ subroutine Calc_WriteOutput_AD()
       end do  ! k=blades
    
          ! rotor outputs
-      if ( EqualRealNos( real(m%V_dot_x,SiKi), 0.0_SiKi ) ) then
-         m%AllOuts( RtTSR )    = 0.0_ReKi
+      if ( abs( m%V_dot_x ) < 0.04_ReKi .or. p%NumBlades == 0 ) then ! < 0.018 is close to "v_dot_x**3 not equal to 0" (cubed because of Cp denominator)
```
ebranlard, line 356: It seems like `0.04` could be a parameter of AeroDyn if it's used elsewhere as a limit.

modules/aerodyn/src/AeroDyn_IO.f90
```diff
@@ -1499,8 +1539,130 @@ SUBROUTINE Cleanup()
...
+   END IF
+
+   IF (ANY(IOS /= 0)) THEN
+      CALL SetErrStat(ErrID_Fatal, "Unable to read numeric data from all columns in the table on row "//trim(num2lstr(i))//".", ErrStat, ErrMsg, RoutineName)
+      RETURN
+   END IF
+      
+END SUBROUTINE ConvertLineToCols
+!----------------------------------------------------------------------------------------------------------------------------------
+SUBROUTINE GetInputColumnIndex(HeaderLine, Indx, ErrStat, ErrMsg)
```
ebranlard, line 1611: In the future, it'll be great to move these to NWTC_IO

modules/aerodyn/src/UnsteadyAero.f90
```diff
@@ -2723,37 +2748,47 @@ subroutine UA_CalcContStateDeriv( i, j, t, u_in, p, x, OtherState, AFInfo, m, dx
...
-      Clp = AFI_AlphaE%FullyAttached + pi * Tu * u%omega                                       ! Eq. 13 (this is really Cnp)
-      dxdt%x(3) = -1.0_R8Ki / BL_p%T_p                                  * x%x(3) +         1.0_ReKi / BL_p%T_p  * Clp
-         
-      dxdt%x(4) = -1.0_R8Ki / BL_p%T_f0                                 *    x4  +         1.0_ReKi / BL_p%T_f0 * AFI_AlphaF%f_st
+      TuOmega = Tu * u%omega
+      TuOmega = MIN( MAX(TuOmega, -MaxTuOmega), MaxTuOmega)
+
+      Clp = AFI_AlphaE%FullyAttached + pi * TuOmega                                        ! Eq. 13 (this is really Cnp)
+      dxdt%x(3) = ( Clp             - x%x(3) ) / BL_p%T_p
+      dxdt%x(4) = ( AFI_AlphaF%f_st - x4     ) / BL_p%T_f0
```
ebranlard, line 2762: Can these modified lines be copied to the UA_HGM as well?

ebranlard (approved):
Thanks a lot, these are great improvements!

modules/aerodyn/src/AeroDyn_IO.f90
bjonkman: The '0.04' m/s limit is only used in this IF statement for calculating output-to-file values for Cp, Cq, and Ct. I had originally used 0.018 m/s, which is similar (but faster) than checking that the denominator in the Cp equation is not zero, but it still left us with large instantaneous values for Cp, Cq, and Ct in some of the DLCs with large wind direction changes.

modules/aerodyn/src/AeroDyn_IO.f90
bjonkman: I moved `GetInputColumnIndex` to NWTC_IO, but it's a little more challenging to get `ConvertLineToCols` to be generic enough to move there. I'll save that for someone else later. :)

modules/aerodyn/src/UnsteadyAero.f90
bjonkman: Done. Good catch, @ebranlard.

bjonkman:
> To do (thinking about this because I'm dealing with AD15 input files for AD14 removal PR):
> 
>     * [x]  update AD15 input files with new nodal output comments
> 
>     * [ ]  update AD15 input files with Solver and other new lines (I know these are optional, but want to get a fully consistent set of input files for the release) -- @andrew-platt

@andrew-platt , As soon as you are okay with the AD15 input file changes you said you wanted to make, this should be good to merge. Though the r-test branch this points to should also be merged into dev.

andrew-platt:
I'm going to update the AD15 input files with the ROM pull request (#1295) and merge this one in first.

There are a couple of minor conflicts on r-test since I merged #2358 in first. I'll resolve those and merge this PR shortly.

PR #2287, opened 2024-06-27
Bug Fix: AeroDyn driver output files contained wrong channel names

bjonkman:
The AeroDyn driver code uses AeroDyn_Inflow (ADI). ADI concatenates the output headers from InflowWind and AeroDyn's first rotor into one array at initialization. However, at each time step, it makes the driver code concatenate the `WriteOutput` arrays. The AeroDyn driver concatenated AeroDyn's outputs from the first rotor and InflowWind output channels (the opposite order of ADI), so all of the AD and IfW channels were mislabeled if both modules were used. 

I changed the order of the `WriteOutput` channels in the AeroDyn driver so that they now match the order of the concatenated headers from ADI. It likely would have caused fewer issues with the regression tests if I had just relabeled the columns, but since ADI is used elsewhere (I think), this seemed like it was less likely to cause issues with other codes that use ADI.

It seems like ADI should handle the concatenation of arrays instead of the driver. If it does it at initialization, it should do it at each time step, too. This would make sure only one code (ADI) needs to be concerned about the internal order of the columns. However, I will leave this as an exercise for someone else to do later.

I also fixed a problem where AD driver was attempting to print the AD version in the text output file, but not initializing the variable it uses. This left a bunch of non-printable characters in the text file.

Finally, I am printing the git hash (version info) in the output files generated by the AeroDyn driver. This applies to both the text and binary files.

AeroDyn driver

All of the AeroDyn driver tests that contain outputs from both AD and InflowWind will fail due to the outputs being in a different order.

```
The following tests FAILED:
	 57 - ad_EllipticalWingInf_OLAF (Failed)
	 58 - ad_HelicalWakeInf_OLAF (Failed)
	 59 - ad_Kite_OLAF (Failed)
	 60 - ad_MultipleHAWT (Failed)
	 67 - ad_BAR_SineMotion (Failed)
	 68 - ad_BAR_SineMotion_UA4_DBEMT3 (Failed)
```

### DIFF
--- a/modules/aerodyn/src/AeroDyn_Driver_Registry.txt
+++ b/modules/aerodyn/src/AeroDyn_Driver_Registry.txt
@@ -15,7 +15,7 @@ usefrom    AeroDyn_Registry.txt
 usefrom    AeroDyn_Inflow_Registry.txt
 #
 # ..... Table of combined cases to run .......................................................................................................
-typedef   AeroDyn_Driver/AD_Dvr Dvr_Case   ReKi                   HWindSpeed       -      -     -   "Hub wind speed" "m/s"
+typedef   AeroDyn_Driver/AD_Dvr Dvr_Case          ReKi                   HWindSpeed       -      -     -   "Hub wind speed" "m/s"
 typedef   ^                         ^             ReKi                   PLExp            -      -     -   "Power law wind-shear exponent" "-"
 typedef   ^                         ^             ReKi                   rotSpeed         -      -     -   "Rotor speed" "rad/s"
 typedef   ^                         ^             ReKi                   bldPitch         -      -     -   "Pitch angle" "rad"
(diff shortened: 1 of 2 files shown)


andrew-platt (approved):
Looks good!  Thanks for doing this -- the order was rather confusing before IfW outputs following the first turbine.

I'll update the regression tests and merge.

andrew-platt:
I agree it would be better to move some of the concatenation into the ADI library instead of having it duplicated in the driver and in the c-bindings interface.  Sometime we might have spare time to condense that.

andrew-platt:
The first four cases appear to work just fine, but the `ad_BAR_*` cases have some very odd issues that look almost like an indexing issue.  I'm not sure how though -- the changes here are really straightforward.  I'm hesitant to merge this until I understand what happened (it is possible it is my comparison code and not a real issue).

andrew-platt:
Thanks for fixing this!

PR #2310, opened 2024-07-09
Update to rotation matrix comments in NWTC Library

RyanDavies19:
Updating the comments in the NWTC library for rotation matrices to reflect what is in the code. This caused some confusion in https://github.com/FloatingArrayDesign/MoorDyn/pull/233#issuecomment-2211313485

n/a

NWTC library


deslaughter (approved):
Based on the equations outlined in the comments, I agree that these are better described as Tait-Bryan angles

PR #2317, opened 2024-07-11
Initialize ED RtHS to zeros

andrew-platt:
Ready to merge

The `RtHS` derived type in _ElastoDyn_ contains many arrays, none of which were initialized to zero during init.  Some of these arrays end up getting used without initializing, so whatever leftover stuff was in memory would then be accessed as values in some of these arrays.

We discovered this issue while converting the parsing of the blade file to use the `FileInfo` data structure and parsing.  If a file of 64 lines long was loaded with the `ProcessCom` command then discarded without use during the `ReadBladeFile` routine, the `5MW_Land_Linear_Aero` case would give the expected results (tested with gcc 12.3.0, double precision release). However, if more lines were present, whatever was written in those memory locations would end up in an `RtHS` array.  We noticed this as incorrect acceleration values for blade 1 nodes in the `y_op` array with values upwards of `1e+294`, which caused issues with linearization.

If line 65 contained `r`, the resulting frequencies would be:
```
46:               Frequency (Hz)                        Damping (%)
46:            ----------------------------         ----------------------------
46:             Ref              New                 Ref              New
46:            0.321115263      0.377324016         0.004761374      0.935368246
46:            0.336468595      0.764882101         0.080286934      0.642368682
46:            0.724206648      0.791899953         0.651774323      0.092775252
46:            0.749265197      1.080146181         0.656903895      0.008118302
46:            0.749326080      1.904270051         0.646239398      0.062853626
46:            1.083362496      1.961998652         0.012309791      0.160466846
46:            1.103154529      1.980423436         0.014866192      0.180651456
46:            1.701439646      2.395306777         0.023216699     -0.521713667
46:            1.928532353      2.618982440         0.180245471      0.577067874
46:            1.945246335      3.155262530         0.177738889      0.021554264
```

If line 65 contained `rrrrrrrr`, the resulting frequencies would be off as follows:
```
46:  Case: 5MW_Land_Linear_Aero:1.lin: :
46:               Frequency (Hz)                        Damping (%)
46:            ----------------------------         ----------------------------
46:             Ref              New                 Ref              New
46:            0.321115263     83.209828452         0.004761374     -0.826506890
46:            0.336468595   8662506957.551628113         0.080286934     -0.000000256
46:            0.724206648   574739393926509749142693497935502516420608.000000000         0.651774323      0.103782929
46:            0.749265197   37471977929674705460798753326446458938499006464.000000000         0.656903895      0.044189924
46:            0.749326080   1596353572817605839899136123782313993697574780928.000000000         0.646239398     -0.221361357
46:            1.083362496   6454645867183894278226561596858301745340078685046177792.000000000         0.012309791     -0.058691704
46:            1.103154529   44775087308082026458361647148189744859863545122018820096.000000000         0.014866192      0.681102728
46:            1.701439646   74380807686665623026286669625369691668842911394732965888.000000000         0.023216699     -0.789905619
46:            1.928532353   301056255025183220231506577309715866922450175489959002112.000000000         0.180245471      0.453293460
46:            1.945246335   3180994479181899036819026195066465440447406087302753026048.000000000         0.177738889      0.017797110
``` 

Explicitly setting all `RtHS` arrays to `0.0_ReKi` during allocation would get rid of this issue.

@deslaughter also found an issue in the `ProcessCom` routine that caused an index issue in Intel compilations.

There may have been some phantom issues caused by this.

ED calculations of blade 1 acceleration values, and perhaps other issues.

### DIFF
--- a/modules/nwtc-library/src/NWTC_IO.f90
+++ b/modules/nwtc-library/src/NWTC_IO.f90
@@ -1626,7 +1626,7 @@ END SUBROUTINE CheckR8Var
 
    Ch = 0
 
-   DO
+   DO while (Ch + 1 < len(Line))
 
       NextWhite = SCAN( Line(Ch+1:) , ' ,;''"'//Tab )
       Ch        = Ch + NextWhite
@@ -4375,12 +4375,11 @@ END SUBROUTINE CheckR8Var
             ! Deallocate the linked list of file names.
 
           CurrFile => FirstFile
-          NextFile => CurrFile%Next
-          DO
+          DO WHILE(ASSOCIATED(currFile))
+              NextFile => CurrFile%Next
               DEALLOCATE(CurrFile)
               IF ( .NOT. ASSOCIATED( NextFile ) )  EXIT
               CurrFile => NextFile
-              NextFile => CurrFile%Next
           ENDDO
           
 !bjj: this needs to happen elsewhere...
@@ -7867,7 +7866,7 @@ end subroutine ReadR8AryWDefault
    INTEGER                      :: NewLineIndx                                  ! The string index where the NewLine character occurs
 
    CHARACTER(10)                :: Frm                                          ! Format specifier for the output.
-   CHARACTER(LEN(InStr))        :: Str                                          ! The next string to be processed
+   CHARACTER(Len=:),allocatable :: Str                                          ! The next string to be processed
 
 
 
--- a/modules/nwtc-library/src/NWTC_IO.f90
+++ b/modules/nwtc-library/src/NWTC_IO.f90
@@ -296,20 +296,16 @@ CONTAINS
 !! If the array is already allocated on entry to this routine, an error will be generated. \n
 !! Use AllocAry (nwtc_io::allocary) instead of directly calling a specific routine in the generic interface.   
    SUBROUTINE AllCAry1 ( Ary, AryDim1, Descr, ErrStat, ErrMsg )
+   ! This routine allocates a 1-D CHARACTER array.
 
-      ! This routine allocates a 1-D CHARACTER array.
-
-      ! Argument declarations.
-
+   ! Argument declarations.
    CHARACTER(*), ALLOCATABLE         :: Ary    (:)                                 !< Array to be allocated
    INTEGER,      INTENT(IN)          :: AryDim1                                    !< The size of the first dimension of the array.
    CHARACTER(*), INTENT(IN)          :: Descr                                      !< Brief array description (for error message).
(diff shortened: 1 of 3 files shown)


deslaughter (approved):

modules/nwtc-library/src/NWTC_IO.f90
```diff
@@ -736,20 +686,17 @@ SUBROUTINE AllRPAry2 (  Ary, AryDim1, AryDim2, Descr, ErrStat, ErrMsg )
    ELSE
       ErrStat = ErrID_None
       ErrMsg  = ''
+      Ary = 0
```
andrew-platt, line 689: I moved this as we could get a memory issue if allocation didn't occur, but we try and zero it anyhow.

bjonkman:
Do you know which arrays were using uninitialized values? FAST 7 used to be built with a compiler option to initialize all variables to 0 (or false), but we were trying to make sure we explicitly initialized variables that need it. This should also give some performance improvements on variables that don't need to be initialized before their use.

modules/nwtc-library/src/NWTC_IO.f90
```diff
@@ -4375,12 +4247,11 @@ SUBROUTINE Cleanup (  )
             ! Deallocate the linked list of file names.
 
           CurrFile => FirstFile
-          NextFile => CurrFile%Next
-          DO
+          DO WHILE(ASSOCIATED(currFile))
+              NextFile => CurrFile%Next
```
bjonkman, line 4251: I am hopeful that this fixes some of the weird behavior that was sometimes causing regression tests to fail on file reads when the OpenFAST input file reading was changed to use this routine!

andrew-platt:
I don't know exactly which arrays weren't getting initialized.  I had started looking into that, but then decided to just take brute force approach to make sure everything got initialized.

modules/nwtc-library/src/NWTC_IO.f90
deslaughter: I hope it improves the stability. This is definitely in the realm of undefined behavior, so it could have been doing anything when CurrFile was not associated

bjonkman:
> I don't know exactly which arrays weren't getting initialized. I had started looking into that, but then decided to just take brute force approach to make sure everything got initialized.

I'm just a little concerned that some of those variables may need to be explicitly zeroed out in the RHS routine so they don't retain values between time steps. If the issue is in an initialization routine that's a little different.

andrew-platt:
That's a very good point!  I had been assuming it was an initialization issue only, but since it is in the RHS they may need explicit zeroing at every timestep as well.

PR #2333, opened 2024-07-18
Cross merge rc-3.5.4 to dev

andrew-platt:
Ready to merge

Some improvements for the 3.5.4 are needed in `dev` for further work in dev.

the primary change is the addition of the `openfast_python` for file io conversions.

@mayankchetan would like to update the `openfast_python` converters for the current `dev` branch.

No test results should change


openfast_python/openfast_io/FAST_linearization_reader.py
```diff
@@ -0,0 +1,391 @@
...
+                if i>=n-1:
+                    break
+            OP=np.asarray(OP)
+            return OP, Var
+
+        def readMat(fid, n, m, name=''):
+
+            # Read rows from file, raise exception on failure
+            try:
+                vals = np.genfromtxt(fid, dtype=np.float64, max_rows=n)
```
deslaughter, line 115: @mayankchetan This doesn't necessarily need to be fixed here, but it turns out that genfromtxt uses a bunch of memory and it's more efficient to do  the following (https://stackoverflow.com/questions/49076386/why-does-np-genfromtxt-initially-use-up-a-large-amount-of-memory-for-large-dat) though you may want to check this code first:

```suggestion
                vals = np.empty([n,m], np.float64)
                for i in range(n):
                    vals[i,:] = fid.readline().split()
```

deslaughter (approved):

openfast_python/openfast_io/FAST_linearization_reader.py
mayankchetan: Thank you for the heads up! will make the changes.

PR #2334, opened 2024-07-18
Updating MD Body Kinematics 

RyanDavies19:
This adds centripetal and gyroscopic forces to bodies, rods, and points to account for differences between the object reference point and its COG. It also enables rod centripetal and gyroscopic force/moment calculations about end A that had been previously commented out. 

The following general code has been added. In bodies it accounts for discrepancies between the reference position and the user defined COG. In points and rods it accounts for the centripetal force they exert on bodies that they are attached to. Note that for rods, while kinematics and forces are handled about end A of the rod, centripetal forces act between the rod COG and the body reference point. 
image

Please review @luwang00 @mattEhall when you get a chance. 
https://github.com/FloatingArrayDesign/MoorDyn/issues/207

MoorDyn

This will likely cause reg tests to fail that use bodies and/or rods in MoorDyn. 

If the commented out changed code (will remove before merging) in the driver is uncommented and the attached input file set is run the following test is run:

This test has two bodies with rods attached. One body is located at [0,0,-2] with a rod attached to the body that has it's COG at the body ref point (this is body1 and rod1 in the md input file). The other body is located at [0,10,-2] with the rod COG located at [0,-10,0] relative to the body (body2 and rod2). This places both rods in the same place. 

Body 1 is then given an initial velocity of [0, 0, 0, 0.2, 0, 0] and body 2 is given an initial velocity of [0, 0, 10*0.2, 0.2, 0, 0]. In theory this would cause both rods to rotate about their COG in the y-z plane with body 2 'orbiting' around rod 2. However what is seen is that body 2 moves constantly along the y=10 line with the rod orbiting it. This is best visualized from the original bug report of this on the C side, which ran the same test: https://github.com/FloatingArrayDesign/MoorDyn/issues/207#issue-2244358237

This test gives the following results (note in before the body2 y location is steady, indicating the body is not 'orbiting' the rod, and in the visualization of the rods, rod 2 clearly is not spinning about its COG): 

Before:
channels_before
zy_plot_before

Now:
channels_after
zy_plot_after

[kinematics_test.txt](https://github.com/user-attachments/files/16289608/kinematics_test.txt)
[kinematics_test_driver.txt](https://github.com/user-attachments/files/16289611/kinematics_test_driver.txt)


modules/moordyn/src/MoorDyn_Rod.f90
```diff
@@ -951,20 +950,20 @@ SUBROUTINE Rod_DoRHS(Rod, m, p)
...
       ! now add centripetal and gyroscopic forces/moments, and that should be everything
       h_c = 0.5*Rod%UnstrLen          ! distance to center of mass
       r_c = h_c*Rod%q                 ! vector to center of mass
       
       ! note that Rod%v6(4:6) is the rotational velocity vector, omega   
-      Fcentripetal = 0.0_DbKi !<<<TEMP<<< -cross_product(Rod%v6(4:6), cross_product(Rod%v6(4:6), r_c ))*Rod%mass <<<
-      Mcentripetal = 0.0_DbKi !<<<TEMP<<< cross_product(r_c, Fcentripetal) - cross_product(Rod%v6(4:6), MATMUL(Imat,Rod%v6(4:6)))
-      
+      Fcentripetal = -cross_product(Rod%v6(4:6), cross_product(Rod%v6(4:6), r_c ))*Rod%mass
+      Mcentripetal = cross_product(r_c, Fcentripetal) - cross_product(Rod%v6(4:6), MATMUL(Rod%Imat,Rod%v6(4:6)))
```
luwang00, line 965: Should we leave a comment here explaining the extra `cross_product(r_c, Fcentripetal)` term coming from using the moment of inertia about the CoG? Just in case we forget again in the future.

luwang00:
Thanks @RyanDavies19 for the update. The equations and code changes look good to me.

modules/moordyn/src/MoorDyn_Rod.f90
RyanDavies19: Resolved with https://github.com/OpenFAST/openfast/pull/2334/commits/f164d946650f2207a0476cc697b542467c175c72

RBergua:
@RyanDavies19: I confirm that with the fix from an hour ago, the initial rotation is properly taken into account.
image

I'm going to re-run the free-decays for the buoyant cans from the USFLOWT project and see if there is any improvement.

RBergua:
I re-run the free-decay that I was commenting here: https://github.com/FloatingArrayDesign/MoorDyn/issues/207#issuecomment-2200701740

Unfortunately, I don't see any differences between this update and the previous MoorDyn version:
image

In case someone wants to run this in MoorDyn standalone, the input file is available in the link that I provided above.

RyanDavies19:
@RBergua thank you for checking this, I was hopeful that wouldn't be the case. Either way, it's good to make things more accurate on the core kinematics.

mattEhall:
Interesting results. So this suggests MoorDyn is underpredicting the hydrodynamic added mass in pitch relative to the other models, right?
Could be a mysterious difference in the transverse added mass. Or more likely a difference related to the added mass of the Rod end(s) and whether that's accounted for in pitch?

RBergua:
That's right, @mattEhall. This suggests that the added mass effect in MoorDyn is slightly smaller than in the other models. Just as a clarification: none of the models accounts for added mass in the axial direction. The added mass is only included in the radial direction.

RyanDavies19:
Note this PR now also has additions to lines, including handling partial submergence and fluid inertia forces. These are from existing code in MD-C

RyanDavies19:
After https://github.com/OpenFAST/r-test/pull/132 is merged, tests should be all good

PR #2339, opened 2024-07-19
Increase number of unit numbers available for OMP file opening

andrew-platt:
Needs testing and approval by @rthedin before merging

We were having problems with the `!OMP` directives around high resolution file reading in AWAE.f90. Since the file reading starts with a call to `GetNewUnit` before starting the opening, parallel calls to `GetNewUnit` could result in the same unit number handed out to two processes.  The first process would open the file and start reading, but then the second process would open a different file with the same unit number causing read errors for both processes as they attempted to read the same file at the same time.

To resolve this, an optional start unit number is added to `GtNewUnit`.  This optional argument is propagated up through to the calls within the `!OMP` loop where a `N*100` is used as the starting unit number for the search (where `N` is the turbine number).  Theoretically this will solve the parallel file reading issue from AWAE.

Routine changes:

`GetNewUnit`
 - Increased maximum unit from 1024 to 64k (likely won't be an issue since most times where more than 1024 files are needed will be on supercomputers where the unit number limit is really high)
 - optional argument for starting unit number.  Defaults to old behaviour when not present. `ReadLowResWindFile`
 - optional argument for starting unit number.  Defaults to old behaviour when not present. `ReadHighResWindFile`
 - optional argument for starting unit number.  Defaults to old behaviour when not present. AWAE `UpdateStates`
 - call to `ReadHighResWindFile` includes `nt*100` as the start unit number

There may be several _FAST.Farm_ issues related to this, but it is unknown at the moment.

_FAST.Farm_ `AWAE` module file reading of high-res VTK files inside the `!OMP` directive in `UpdateStates`

### DIFF
--- a/modules/awae/src/AWAE.f90
+++ b/modules/awae/src/AWAE.f90
@@ -1458,6 +1458,7 @@ subroutine AWAE_UpdateStates( t, n, u, p, x, xd, z, OtherState, m, errStat, errM
    character(*), parameter                      :: RoutineName = 'AWAE_UpdateStates'
 !   real(DbKi)          :: t1, t2
    integer(IntKi)                               :: n_high_low, nt, n_hl, i,j,k,c
+   integer(IntKi)                               :: UnStart        ! Start number for opening file for this turbine (to avoid two turbines using same number in parallel)
    
    errStat = ErrID_None
    errMsg  = ""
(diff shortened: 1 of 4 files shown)


andrew-platt:
Closing this in preference of the simpler solution in #2340

PR #2340, opened 2024-07-19
Add !$ OMP critical around file opening for VTK

andrew-platt:
This is the preferred alternate approach to PR #2339.  If this works, this PR will be merged in and the other closed.

We were having problems with the `!OMP` directives around high resolution file reading in AWAE.f90. Since the file reading starts with a call to `GetNewUnit` before starting the opening, parallel calls to `GetNewUnit` could result in the same unit number handed out to two processes.  The first process would open the file and start reading, but then the second process would open a different file with the same unit number causing read errors for both processes as they attempted to read the same file at the same time.

Adding `$OMP critical` around the `GetNewUnit` and following `OpenFile...` calls so that in theory these cannot be done in parallel.

#2339 

VTK file reading within `!$OMP` loops

@rthedin will report on testing with many turbines in a _FAST.Farm_ simulation.

### DIFF
--- a/modules/nwtc-library/src/VTK.f90
+++ b/modules/nwtc-library/src/VTK.f90
@@ -158,8 +158,10 @@ contains
          closeOnReturn = .FALSE.
       END IF
       
+      !$OMP critical
       CALL GetNewUnit( Un, ErrStat, ErrMsg )      
       CALL OpenFInpFile ( Un, TRIM(FileName), ErrStat, ErrMsg )
+      !$OMP end critical
          if (ErrStat >= AbortErrLev) return
       
        CALL ReadCom( Un, FileName, 'File header: Module Version (line 1)', ErrStat2, ErrMsg2, 0 )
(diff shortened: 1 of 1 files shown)


rthedin:
Tested on Kestrel on an old 5-turbine case that worked with 3.5.3 commenting out the `!OMP` directives as Andy mentioned above. It runs with no problem. 

The following header is printed:
```
 FAST.Farm-v2.5.0-3190-g0e7774a6
 Compile Info:
  - Compiler: Intel(R) Fortran Intel(R) 64 Compiler Classic for applications running on Intel(R)
  64, Version 2021.10.0 Build 20230609_000000
  - Architecture: 64 bit
  - Precision: double
  - OpenMP: Yes, number of threads: 104/104
  ```

It seems like there is an issue with the version number. Also, note that I have not noticed any performance difference between omp-commented-out 3.5.3 and this branch. CPU use patterns are the same, estimated time for completion is within 2 minutes of each other (total expected runtime ~6 hours).

rthedin:
I don't have a test case ready with many (100+) turbines. I will test it, but might take a few days.

andrew-platt:
With a debug run, there may not be a big performance increase given all the other single threaded parts of `AWAE`.  Thanks for testing @rthedin!

PR #2342, opened 2024-07-22
Linear damping for rod elements in MoorDyn

RBergua:
I believe this PR is ready to be merged.

This is a development done by Front Energies (Johyun Kyoung) as part of the USFLOWT project. The system uses large buoyant cans modeled as rods in MoorDyn.
image

Previous to this development, it was only possible to include viscous damping from drag forces (quadratic).

The proposed code allows to include transverse linear damping per unit of length [N/(m/s)/m] for the rod element.

The updated section of the rod element in the input file looks as follows:
image

MoorDyn


andrew-platt:
Is the `LinDamp` column in the input file optional, or required?

RBergua:
If the user doesn't define the `LinDamp` column, the code assumes that there is no damping. It should be backcompatible.

Screenshot from the debugging that we were using in the development:
image

PR #2357, opened 2024-08-02
Bug Fix: UA driver aero-elastic simulation had wrong sign

bjonkman:
When using the new aero-elastic capabilities in the UA driver, we noticed some simple cases that diverged. Further inspection led us to find a sign error in the LinDyn implementation (see matrix `BB`).

This was discussed in an email chain between Envision Energy and @ebranlard.
 
UA driver, SimMod=3

I added a new regression test for UA driver to expose this bug, but the existing cases for the UA driver aren't run in GitHub, so... until that is added, this won't change any regression tests.

### DIFF
--- a/modules/lindyn/src/LinDyn.f90
+++ b/modules/lindyn/src/LinDyn.f90
@@ -293,7 +293,7 @@ subroutine StateMatrices(MM, CC, KK, AA, BB, errStat, errMsg)
    LWORK=nx*nx ! Somehow LWORK = -1 does not work
    allocate(WORK(LWORk))
    call LAPACK_getri(nx, MinvX, IPIV, WORK, LWORK, errStat2, errMsg2); if(Failed()) return
-   BB(nx+1:nq, : ) = -MinvX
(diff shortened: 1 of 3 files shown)


ebranlard (approved):
Thanks a lot for doing that!

modules/lindyn/src/LinDyn.f90
```diff
@@ -251,6 +251,7 @@ subroutine StateMatrices(MM, CC, KK, AA, BB, errStat, errMsg)
    integer(IntKi)                          :: errStat2    ! Status of error message
```
ebranlard, line 251: The typo was also in the comment above (the minus should be a plus in B).... 
```
!! A = [   0        I       ]     B = [0      ]
!!     [-M^{-1}K   -M^{-1}C ]       = [ + M^{-1}]  
```

bjonkman: Thanks for the review. I've fixed that comment.

PR #2397, opened 2024-08-28
HD bug fix: prevent array index out-of-bound error in HDOut_MapOutputs when more than 9 potential-flow bodies are present

luwang00:

Currently, HydroDyn allows an arbitrary number of potential-flow bodies, but the relevant output channels are only available for the first 9 bodies. Having more than 9 bodies in HydroDyn will lead to a memory access exception during output. 

This pull request fixes this bug. Output is still limited to the first 9 bodies, but having more potential-flow bodies in the model should no longer cause problems.

HydroDyn

No change to existing tests since none involve more than 9 potential-flow bodies in HydroDyn.


jjonkman (approved):
Thanks for finding and fixing this bug!

FYI: This bug was originally reported in: https://forums.nrel.gov/t/hydrodyn-nbody-maximum-limit/7418.

andrew-platt (approved):

PR #2203, opened 2024-05-10
Support for large platform yaw offset in OpenFAST

luwang00:
**Feature or improvement description**
* The PR adds support for unrestricted platform rotation in ElastoDyn and SubDyn (for floating only).
* HydroDyn is currently limited to large yaw motion only with small roll and pitch offset.
* This PR also updates the general checking and handling of wave headings in the WAMIT and WAMIT2 modules for robustness. It should address some known issues when the range of wave headings crosses the +/-180-deg cut.
* The option to disable GuyanLoadCorrection in SubDyn is removed. GuyanLoadCorrection will always be used from now on.

**Impacted areas of the software**
Main impacted modules: ElastoDyn, HydroDyn, SubDyn
Minor updates for linearization: AeroDyn, MoorDyn, ServoDyn

**Related Issue**
For floating, the Guyan (rigid-body) modes of SubDyn models are made nonlinear and exact, consistent with ElastoDyn-only models. This change should address the discrepancies observed in Issue #2021.

**Test results, if applicable**
* r-test HydroDyn files are updated with the new inputs. 
* With ElastoDyn and SubDyn switching to fully nonlinear platform rotation instead of small rotation approximation, the results of many existing r-tests involving a floating structure also changed and need to be updated. 
* Some r-test results for the SubDyn module also changed because the option to disable `GuyanLoadCorrection` is removed, and all r-test results for SubDyn need to be regenerated with `GuyanLoadCorrection` set to true.
* HydroDyn r-test input files need to be updated because the input PRP motion file is now assumed to contain the time derivatives of the Tait-Bryan angles instead of the angular velocity and angular acceleration. HD driver will convert the input to angular velocity and angular acceleration internally.
* HydroDyn r-test output files also changed because some nonlinearity with rotational offset is introduced into the hydrostatic loads of potential-flow bodies.
* Several new r-tests with large yaw offset are added for the HydroDyn driver.
See [PR #133](https://github.com/OpenFAST/r-test/pull/133) to r-test.

Representative changes to r-test results are shown below:

**5MW_OC3Mnpl_Linear**
The eigenfrequencies from this test case show slight changes due to the changes made to the platform roll, pitch, and yaw DoF in ElastoDyn.

```
|             New            |     Reference (dev with     |
|                            |     GuyanLoadCorrection)    |
|  Freq (Hz)  |   Damp (%)   |   Freq (Hz)  |   Damp (%)   | 
|-------------|--------------|--------------|--------------|
[...]

PR #2401, opened 2024-08-30
SD: Remove the misleading CBMod input

luwang00:
This pull request slightly modifies the SubDyn input file by removing the `CBMod` T/F switch. This input can be misleading because irrespective of `CBMod`, SubDyn always transforms the equations of motion to be in terms of Guyan and Craig-Bampton modes. Setting `CBMod` to false simply lets SubDyn retain all Craig-Bampton modes.

To better reflect how SubDyn actually functions, the `CBMod` switch is removed. Instead, if `Nmodes`>=0, SubDyn will function as if `CBMod` is true in previous versions. Alternatively, setting `Nmodes`<0 will let SubDyn retain all Craig-Bampton modes, recovering the same behavior with `CBMod` set to false in previous versions.

The user documentation for SubDyn is updated to reflect the removal of `CBMod` from the primary input file. The removal of `GuyanLoadCorrection` from the input file is also briefly explained (see PR #2203). 

SubDyn, Docs

There is no change to existing test results. All r-test SubDyn input files are updated by removing `CBMod`.

### DIFF
--- a/docs/source/user/subdyn/input_files.rst
+++ b/docs/source/user/subdyn/input_files.rst
@@ -229,13 +229,6 @@ static gravity and buoyancy loads, and high-frequency loads transferred
 from the turbine. Recommended to set to True.
 
 
-**GuyanLoadCorrection** is a flag to specify whether the extra moment due to 
-the lever arm from the Guyan deflection of the structure is to be added to the loads
-passed to SubDyn, and, whether the FEM representation should be expressed in the rotating 
-frame in the floating case (the rotation is induced by the rigid body Guyan modes).
-See section :numref:`SD_Loads` for details. Recommended to set to True.
-
-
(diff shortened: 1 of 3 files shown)


andrew-platt:
Are there documentation changes required, or is the documentation for this part missing?

luwang00:
@andrew-platt I have not updated the documentation for SubDyn yet. We also need to address the removal of GuyanLoadCorrection in the input file. I can quickly address these two changes and add to this PR. (SD theory manual also needs to be updated, but that will have to wait for the future I think.)

luwang00:
Hi @andrew-platt, I updated the SD user manual regarding the removal of the two inputs `GuyanLoadCorrection` and `CBMod`.

andrew-platt (approved):

PR #2406, opened 2024-09-05
Correct issues with SED

andrew-platt:
Ready to merge after regression test updates

Somewhere during final updates of the reduced order models (_SED_ and _ADsk_), an incorrect merge resulted in missing logic in `FAST_Subs.f90` for the _SED_ + _AD_ combination, which resulted in segmentation faults.  This is now fixed.

Also, _SED_ was carrying version info while no other modules apart from _MoorDyn_ carry version info anymore (the module version is now the _OpenFAST_ version), so version info was dropped.

The following regression tests are now turned on in GH actions and have changed very slightly from the original baselines 18 months ago (changes are minimal and due entirely to other changes in the code).

-  `5MW_Land_DLL_WTurb_ADsk`
- `5MW_Land_DLL_WTurb_ADsk_SED`
- `5MW_Land_DLL_WTurb_SED`

PR #1295 

The new module _SED_ now works with _AD_.

@mayankchetan, please review.

### DIFF
--- a/modules/openfast-library/src/FAST_Subs.f90
+++ b/modules/openfast-library/src/FAST_Subs.f90
@@ -783,17 +783,28 @@ SUBROUTINE FAST_InitializeAll( t_initial, p_FAST, y_FAST, m_FAST, ED, SED, BD, S
       Init%InData_AD%MSL2SWL            = p_FAST%MSL2SWL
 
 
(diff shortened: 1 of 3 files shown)


mayankchetan (approved):
Looks good Andy! don't see the segfault when I run locally.

deslaughter (approved):

PR #2408, opened 2024-09-07
HD: Bug fix for potential-flow wave excitation with multiple bodies and large yaw offset

luwang00:
Two bugs were identified that led to incorrect potential-flow wave excitation when multiple bodies are present with large yaw offset (`PtfmRefY!=0` or `PtfmYMod=1`):

* When `NBodyMod=1` and `NBody>1`, only the wave excitation on the first body is transformed back to the inertial frame. This PR adds the transformation of the subsequent bodies. 

* When `PtfmRefxt!=0` or `PtfmRefyt!=0`, there is effectively a displacement of each potential-flow body due to platform yaw offset when precomputing the wave excitation. In other words, the effect of this baseline displacement is already baked into the precomputed wave excitation, so it needs to be subtracted from the total body displacement when interpolating the wave excitation if `ExctnDisp>0`. This correction is implemented in this PR.

HydroDyn

Existing tests are not impacted by the bugs identified, so no change to the test results.


andrew-platt (approved):
I don't know all the details in your transform, but this looks correct.

andrew-platt:
Related to #2203

PR #2412, opened 2024-09-11
ExtLd: require use of InflowWind

andrew-platt:
Ready to merge.  Requires testing with FSI CFD (after update to interface on that side).

Originally the _ExtLoad_ module was designed to provide a set of steady wind parameters to _OpenFAST_ from FSI CFD (not sure if this is _AMR-Wind_ or _NALU-Wind_).  These parameters were used to calculate a steady wind profile for _AeroDyn_ to use during the initial few rotations where _AeroDyn_ and _ExtLoad_ loads were blended (this was for initializing the CFD better).

However, after merging in the _InflowWind_ pointers change, there was a bug with how this worked. After some discussion with @gantech and others, it was decided that the best approach would be to now require _InflowWind_ with _ExtLoad_ (previously this was not allowed) and have _InflowWind_ handle all the wind for _AeroDyn_ throughout the simulation.  The following changes were made:

- Remove wind calculations inside _ExtLd_.  These variables removed:
  - vel_mean
  - wind_dir
  - z_ref
  - shear_exp
- Require use of InflowWind
- remove following from OF library interface (_NOTE: this requires revision of the FSI CFD routines calling `FAST_ExtLoads_Init`_):
  - vel_mean_c
  - wind_dir_c
  - z_ref_c
  - shear_exp_c

Issue brought up in discussions, not GH.

The C++ interface is affected.

We still don't have a test case for this...


gantech (approved):
Thanks for getting this done Andy. After going through the implementation, I checked the `5MW_Restart` regression test against the output of the fortran driver with `CompAero` set to 2. All channel diff norms are within 1e-12. This is good to go and a welcome reduction in complexity for blade-resolved FSI cases.

Issue #2413, opened 2024-09-11
Lidar module array not fully initialized


bjonkman:
Is there a reason that the `ParseAry` routine called for  variable `RotorApexOffsetPos` is reading only one element in the array instead of all 3?
https://github.com/OpenFAST/openfast/blob/2246befaacf88a9657deb0606e08b33fc7338a11/modules/inflowwind/src/InflowWind_Subs.f90#L543

https://github.com/OpenFAST/openfast/blob/2246befaacf88a9657deb0606e08b33fc7338a11/modules/inflowwind/src/InflowWind.txt#L71

All elements of the array are used, so it seems like the y- and z- components of this position are uninitialized.
https://github.com/OpenFAST/openfast/blob/2246befaacf88a9657deb0606e08b33fc7338a11/modules/inflowwind/src/Lidar.f90#L371

andrew-platt:
That's a qood question.  I checked the history and it looks like it may have been written that way.

This should be corrected.

bjonkman:
I have a fix in my `f/Envision_updates` branch. I need to update some input files in the r-test so they don't fail with this new requirement, though. 

Just wanted to make sure it wasn't supposed to be an offset in only the X direction. Thanks.

andrew-platt:
Thanks!

I suspect the intent was for all three directions, but probably never got tested for that.

andrew-platt:
Corrected with #2415

PR #2415, opened 2024-09-12
Lidar bug fix + other minor changes

bjonkman:
- Bug fixes:
   - The `RotorApexOffsetPos` input in the InflowWind input file was reading only the first (x) component of the 3-dimensional array, so the other two were likely uninitialized when used in the Lidar module. This change also required an update to one of the r-test InflowWind input files that didn't have all 3 values.
   - The Visual Studio build was not able to create ServoDyn_Types.f90. This has been fixed.

- Other updates:
   - I updated some spacing and comments in the code.
   - I reorganized small bits of code for better compatibility with Envision code base.
   - I removed unused variables.
   - I fixed some issues with single-/double-precision types in some of the new HydroDyn code.
   - I removed some redundant error handling (e.g., errors in reading the Lidar input parameters from IfW would result in 2 messages instead of 1).
   - I reorganized a bit of the FAST Visual Studio project, creating folders for CompAero modules, CompInflow, and CompSub modules.

https://github.com/OpenFAST/openfast/issues/2413

InflowWind, Lidar, HydroDyn, MAP++, glue code, NWTC_Library, ServoDyn

This doesn't affect any results.

### DIFF
--- a/glue-codes/simulink/src/FAST_SFunc.c
+++ b/glue-codes/simulink/src/FAST_SFunc.c
@@ -189,9 +189,6 @@ static void mdlInitializeSizes(SimStruct *S)
              InitInputAry[i] = AdditionalInitInputs[i + 1];
           }
        }
-       else{
-          InitInputAry[0] = SensorType_None; // tell it not to use lidar (shouldn't be necessary, but we'll cover our bases)
-       }
 
        // set this before possibility of error in Fortran library:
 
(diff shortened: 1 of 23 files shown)


andrew-platt (approved):
My only concern is with the lidar connection to the OpenFAST_Library.  Hopefully we don't break someones connection to another code with this change (I'm guessing we don't)

bjonkman:
The change to that interface should be only (1) removing the definition of `SensorType _None` for the lidar and (2) modifying the `FAST_ExternInitType` data structure to remove the lidar variables. I'd be surprised if any of the other interfaces to OpenFAST (other than the Simulink one) would be affected. But, I guess we'll find out. :)

andrew-platt:
That's what I was wondering about; does someone link it to SimuLink and I'm not aware of it.  If we get a complaint about that, we can add it back in then (and maybe fix a few things at that point).

bjonkman:
> That's what I was wondering about; does someone link it to SimuLink and I'm not aware of it. If we get a complaint about that, we can add it back in then (and maybe fix a few things at that point).

Well, even if they _were_ linking to Simulink another way, I didn't actually change any functionality that's been there since #1464 was merged in April 2023. The code was not doing anything with those variables that used to get passed from Simulink. They are read from the InflowWind input file. You can still pass them in that Simulink interface, they just get ignored earlier than they used to.

andrew-platt:
Thanks for the clarification!  Given that the variables weren't getting passed all the way, this change shouldn't affect anyone.

PR #2417, opened 2024-09-12
MAP: replace strncpy with strncpy_s

andrew-platt:
Ready to merge (need someone with actual C knowledge to review it though).

The intel c++ w_oneAPI_2021.1.1.99 compiler does not like the `strncpy` function and won't compile.  Changing to `strncpy_s`.

See comment here: https://github.com/OpenFAST/openfast/commit/93e2d2776a17146c505f6e6158dcad4995db5f99#r14664285  from PR #2394
Compilation of MAP with some compilers.

@sanguinariojoe, can you check this?

### DIFF
--- a/modules/map/src/mapapi.c
+++ b/modules/map/src/mapapi.c
@@ -1164,7 +1164,16 @@ MAP_EXTERNCALL void map_set_gravity(MAP_ParameterType_t* p_type, const double gr
 
 MAP_EXTERNCALL void map_set_input_text(MAP_InitInputType_t* init_type, const char* input_txt_line)
 {
-  strncpy_s(init_type->library_input_str, sizeof init_type->library_input_str, input_txt_line, (sizeof init_type->library_input_str)-1);
+  int i = 0;
+  for (i = 0; i < (sizeof init_type->library_input_str); ++i)
+  {
+    init_type->library_input_str[i] = input_txt_line[i];
+    if (input_txt_line[i] == 0) break;
+  }
+  if (init_type->library_input_str[i] != 0)
+  {
+    init_type->library_input_str[i] = 0;
+  }
   init_type->library_input_str[254] = '\0';
 }
 
--- a/modules/map/src/mapapi.c
+++ b/modules/map/src/mapapi.c
@@ -1164,7 +1164,7 @@ MAP_EXTERNCALL void map_set_gravity(MAP_ParameterType_t* p_type, const double gr
 
 MAP_EXTERNCALL void map_set_input_text(MAP_InitInputType_t* init_type, const char* input_txt_line)
 {
-  strncpy(init_type->library_input_str, input_txt_line, 254);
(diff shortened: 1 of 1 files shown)


andrew-platt:
Of course, changing to `strncpy_s` doesn't work with gcc...  back to the drawing board I guess.

bjonkman:
I had an issue with this a few days ago using the Visual Studio 2022 v143 C/C++ compiler. I added the `_CRT_NONSTDC_NO_WARNINGS` definition in the code when using MSVC: https://github.com/OpenFAST/openfast/blob/32ad558b93eb9221e59d7f9ef0652a2628123981/modules/map/src/mapapi.h#L39
 but Visual Studio had also turned on the compiling option `Additional Security Development Lifecycle (SDL) recommended checks`, [which made the warning an error](https://learn.microsoft.com/en-us/cpp/build/reference/sdl-enable-additional-security-checks). When I removed the `\SDL` option, it was able to build with the `strncpy` routine. 

image

This may be helpful: https://stackoverflow.com/questions/22450423/how-to-use-crt-secure-no-warnings

sanguinariojoe:
Microsoft is a truck load of shit... Anyway, I would say the smartest solution would be replacing the offending line and the following one by:

```
size_t n = strnlen(input_txt_line, sizeof(init_type->library_input_str)) - 1; 
*((char *)memcpy (init_type->library_input_str, input_txt_line, n) + n) = '\0';
```

Checkout the critics of Ulrich Drepper to `strlcpy`.

I wrote that code directly here, on GitHub. It is not event tested.

modules/map/src/mapapi.c
```diff
@@ -1164,7 +1164,7 @@ MAP_EXTERNCALL void map_set_gravity(MAP_ParameterType_t* p_type, const double gr
 
 MAP_EXTERNCALL void map_set_input_text(MAP_InitInputType_t* init_type, const char* input_txt_line)
 {
-  strncpy(init_type->library_input_str, input_txt_line, 254);
+  strncpy_s(init_type->library_input_str, sizeof init_type->library_input_str, input_txt_line, (sizeof init_type->library_input_str)-1);
```
deslaughter, line 1167: I propose the following, which doesn't doesn't use the `str*` functions and guarantees that the string will be null terminated. Which is essentially what `strncpy` should do.

```suggestion
  int i = 0;
  for (i = 0; i < (sizeof init_type->library_input_str); ++i)
  {
    init_type->library_input_str[i] = input_txt_line[i];
    if (input_txt_line[i] == 0) break;
  }
  if (init_type->library_input_str[i] != 0)
  {
    init_type->library_input_str[i] = 0;
  }
```

andrew-platt:
In trying out @deslaughter's suggestion, I discovered there is a `MAP_STRCPY` macro that already handles switching from `strcpy` and `strcpy_s` depending on compiler setup.  Going to switch over to that instead -- will replace this PR with a PR from a new branch.

andrew-platt:
Thanks @bjonkman for the workaround in VS.  I think the new solution in #2420 is a bit simpler for end users, so we'll pursue that route.

PR #2420, opened 2024-09-13
MAP: change strncpy to use macro MAP_STRNCPY

andrew-platt:
Ready to merge.

Some compilers do not support the `strncpy_s` routine.  Replacing this call with a call to a macro who's definition is based on which compiler is in use.  This new macro `MAP_STRNCPY` is based on the `MAP_STRCPY` macro and  points to either `strncpy` or `strncpy_s`.

Supersedes PR #2417
Original issue mentioned in comment here: https://github.com/OpenFAST/openfast/commit/93e2d2776a17146c505f6e6158dcad4995db5f99#r14664285 from PR https://github.com/OpenFAST/openfast/pull/2394

_MAP_ module compilation only

Evidently incompatibilities between MSVC and gcc were run into before.  I suspect this is why @mmasciol added the macro definitions for `MAP_STRCPY` during initial _MAP_ development in ~2011.

No tests exist for this, so nothing changes.

### DIFF
--- a/modules/map/src/mapapi.c
+++ b/modules/map/src/mapapi.c
@@ -1164,7 +1164,7 @@ MAP_EXTERNCALL void map_set_gravity(MAP_ParameterType_t* p_type, const double gr
 
 MAP_EXTERNCALL void map_set_input_text(MAP_InitInputType_t* init_type, const char* input_txt_line)
 {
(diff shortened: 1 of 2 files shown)


andrew-platt:
@sanguinariojoe, we think this may be a solution that is more consistent with the existing code.

deslaughter (approved):

PR #2425, opened 2024-09-16
AD: Bug Fix: Twist in blades 2 and 3 wrong when using Polar BEM

ebranlard:

In #2349 I introduced a bug in the Polar BEM when I introduced a wrapper function fo rthe calculation of the toe, twist cant angles. An index was missing, leading to only blade 1 having the correct twist. 

AeroDyn  - Polar BEM (currently untested)


andrew-platt:
Does line 3615 in AeroDyn.f90 also need this change?  https://github.com/ebranlard/openfast/blob/f/bladeTwistMissing23/modules/aerodyn/src/AeroDyn.f90#L3615

ebranlard:
Yes, thanks a lot for spotting this!

andrew-platt (approved):

Issue #2241, opened 2024-05-28
Run FAST.Farm on windows.


CaiZhiming-HHU:
Dear all,
When I run the IEA-15MW-Semi case on windows using FAST.Farm (compiled and installed using MinGW), it always terminating the process after "Running WakeDynamics." without any error messages and without generating any files. OpenFAST is working fine, I don't know what's going on, and I'd appreciate any advice.

CMD display when running FAST.Farm👇
![image](https://github.com/OpenFAST/openfast/assets/89171153/e418cfc6-4e3d-4192-bc3c-976358f0ce6e)
CMD display when running OpenFAST👇
![image](https://github.com/OpenFAST/openfast/assets/89171153/dec080e5-46d9-4689-9c97-8c022a32d4d9)
@NRELAdmin @jjonkman @bjonkman

jjonkman:
Dear @CaiZhiming-HHU,

Do you have any any issues running the precompiled version of FAST.Farm for Windows as provided with the release of OpenFAST v3.5.3?

Best regards,

CaiZhiming-HHU:
> Dear @CaiZhiming-HHU,
> 
> Do you have any any issues running the precompiled version of FAST.Farm for Windows as provided with the release of OpenFAST v3.5.3?
> 
> Best regards,

![image](https://github.com/OpenFAST/openfast/assets/89171153/e71a05bf-758f-482c-928e-6dbf45d40155)
This is what I encountered when I ran the precompiled version of FAST.Farm for Windows as provided with the release of OpenFAST v3.5.3. This problem only occurs when the IEA-15MW-Semi case calculation is run, and the NREL-5MW case calculation behaves normally.

jjonkman:
Dear @CaiZhiming-HHU,

The FAST.Farm TSinflow r-test uses OpenFAST models of the NREL 5-MW baseline wind turbine.  Did you change this r-test to use OpenFAST models of the IEA Wind 15-MW RWT atop the VolturnUS semisubmersible?  Can you share your updated FAST.Farm primary input (_.fstf_) file?

Best regards,

CaiZhiming-HHU:
> IEA Wind 15-MW RWT atop the VolturnUS semisubmersible

[...]

Issue #2053, opened 2024-02-25
FAST.Farm


ILDENIEL:
Hello everyone, 
I'm trying to run FAST.Farm but an error occurred.
Before downloading the newest version I've used version 3.4.1 and FF ran without any problem.
![image](https://github.com/OpenFAST/openfast/assets/125210312/36bbcee7-ec8f-4547-9b97-092cf24d1511)

OpenFAST v3.5.2 runs smoothly as always, what could be the problem? 
Thanks in advance

ILDENIEL:
For compliteness
https://github.com/OpenFAST/openfast/issues/2010#issuecomment-1924534167
At this link it is possible to find the problems with the previous version of FAST.Farm. For this reason I tried with FF v3.5.2.

jjonkman:
Dear @ILDENIEL,

I'm not sure I understand.  Are you saying a FAST.Farm model that runs without issue in OpenFAST v3.4.1 triggers a stack overflow error when running in OpenFAST v3.5.2?  If so, can you share the full set of input files for both versions?  If not, please clarify.

Best regards,

andrew-platt:
Thanks for the set of input files.  @jjonkman confirmed that it isn't an obvious configuration issue, so there is a bug in the code someplace causing the stack overflow.  I will take a look at it and report what I figure out.

ILDENIEL:
Thank you for your support @andrew-platt @jjonkman,
I will continue using FF v3.4.1 for now.
Best regards

yangpy88:
It seems like version 3.5.1 does not have this problem, but 3.5.0 and 3.5.2 have. Just FYI.

andrew-platt:
We think this is fixed with #2118.  Please reopen if you have issues with v3.5.3 (will be posted later this week).

andrew-platt:
This apparently was not fixed with #2118 as the issue still exists in the 3.5.3 release.

771217832:
My dear
I had the same stack overflow issue(https://forums.nrel.gov/t/fast-farm/1887/484) with a low wind speed wind farm calculation ,and i fixed the issue by rolling back the Fast.farm version to 3.4.1
Best regards

andrew-platt:
This should now be fixed with #2452.  Closing this issue for now (we can reopen again if necessary).

Issue #843, opened 2021-09-28
Bug report


Ry8icks:
**Bug description**
"Stack overflow" is issued when running with a jacket. In my case the jacket have 383 members.

**To Reproduce**
Run with a substructure with a lot of joint/members.

**Expected behavior**
It should work...
However, this is also an easy fix. Just increase the stack size for the linker. I've increased to 512 Mb (from I think 9999999.)
![image](https://user-images.githubusercontent.com/39679738/135050700-018fe0ed-657f-4214-9fe5-f298aa44ad8e.png)

**Screenshots, if applicable**
<Add screenshots to help explain your problem.>

**OpenFAST Version**
<Please provide as much detail as possible including git commit. The best information is a screenshot of the OpenFAST system description that prints when running OpenFAST:>

```
**************************************************************************************************
 OpenFAST

 Copyright (C)  National Renewable Energy Laboratory
 Copyright (C)  Envision Energy USA LTD

 This program is licensed under Apache License Version 2.0 and comes with ABSOLUTELY NO WARRANTY.
 See the "LICENSE" file distributed with this software for details.
 **************************************************************************************************

 OpenFAST-v2.0.0
 Compile Info:
  - Architecture: 64 bit
  - Precision: double
  - Date: Nov 27 2018
  - Time: 17:19:38
 Execution Info:
  - Date: 11/29/2018
  - Time: 10:52:28-0700
```

**System Information (please complete the following information):**
 - OS: <e.g. Ubuntu 14.04 or macOS 10.12>
 - Compiler: <e.g. GFortran 4.4>
 - Compiler settings: <e.g. CMake flags or other settings>

**Additional context**
<Add any other context about the problem here.>

ebranlard:
Hi @Ry8icks,
Thank you for your report. 
Is the system information you listed correct? It seems you are using visual studio on Windows, no? 
Did you run OpenFAST or the SubDyn driver?

Ry8icks:
naah , here's the correct info:
OpenFAST-v3.0.0-dirty
Compile Info:
 - Compiler: Intel(R) Fortran Compiler 1910
 - Architecture: 64 bit
 - Precision: single
 - OpenMP: No
 - Date: Sep 27 2021
 - Time: 16:55:23
Execution Info:
 - Date: 10/01/2021
 - Time: 16:12:12+0200

and yes, I'm using Visual Studio with the latest Intel Visual Fortran Compiler.
And, I was running a recompiled version of Openfast.

[...]

PR #2452, opened 2024-10-08
Fix for some stack overflow issues with FAST.Farm

andrew-platt:
Ready to merge

We have been having issues with stack overflows in FAST.Farm compiled with Intel and Visual Studio.  The root cause is due to copying of large wind grids within the _AWAE_ module.

Using the zip files attached to issue #2053, this was eventually tracked down to line 1078 in `AWAE.f90` which reads:
```
m%u_IfW_Low%PositionXYZ = p%Grid_low
```

The `p%Grid_low` is of unknown size at compile time, but can be extremely large (3x1643364 in the test case).  This copy involves a temporary array and would normally be handled on the stack, but could result in an overflow for some models.  By setting  `/heap-arrays:1000`, any operation resulting in a temporary array of unknown array size at compile time will use the heap for the temprary array, and as will any array known at compile time to be larger than 1000 kB.

See https://www.intel.com/content/www/us/en/docs/fortran-compiler/developer-guide-reference/2024-2/heap-arrays.html for reference.

This solves #2053, and may also solve #843 and #2241

Certain models that had very large arrays that were being copied.

Ideally we would find a different way of handling the copy in `AWAE.f90` line 1078 (maybe a pointer to the dataset or similar operation that doesn't require the copy at all -- I'm assuming this is a bottleneck).

No test results are affected.

### DIFF
--- a/cmake/OpenfastFortranOptions.cmake
+++ b/cmake/OpenfastFortranOptions.cmake
@@ -158,7 +158,9 @@ endmacro(set_fast_intel_fortran)
 # arch
 #
 macro(set_fast_intel_fortran_posix)
-  set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -fpic -fpp")
+  # Set size where temporary are stored on heap instead of stack
+  #   1000: size in kB (1 MB)
+  set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -fpic -fpp -heap-arrays 1000")
 
   # debug flags
   if(CMAKE_BUILD_TYPE MATCHES Debug)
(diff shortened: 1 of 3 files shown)


andrew-platt:
Notes:

1. I am a little unsure if I have the syntax correct in the `cmake` file.  The documentation linked above says the flag on windows should be `/heap-arrays:1000`, but the VS project produces the flag `/heap-arrays1000`.  I suspect either should work, but don't have a good way to check at the moment.
2. The 1 MB limit was arbitrarily chosen.  This probably isn't an issue since the compiler will put any temporary arrays of unknown size at compile time on the heap instead of stack with this new flag.
3. It may be worth adding this flag to the VS projects for drivers that may have large temporary arrays (I may add that before merging)

andrew-platt:
@bjonkman, could you verify that this is a sensible solution?

Issue #2447, opened 2024-10-04
CalculateUACoeffs fails when iLow2 == iHigh2


deslaughter:
**Bug description**
<A clear and concise description of the bug.>

The `CalculateUACoefficients` subroutine was modified in 030500ef2b402273c90853b01151e75b610b1e11 and the subroutine `Calculate_C_alpha` was added. Under some circumstances, the indices determined in this subroutine, `iLow2` and `iHigh2`, are set to the same value. This seems to occur if min(Cd) is not close to alpha = 0. The logic used to determine `alphaLower` and `alphaUpper` results in a single point being passed to `Calculate_C_alpha` which requires at least two points to be provided to perform a linear curve fit. In Debug mode this results in a bounds access error in `Calculate_C_alpha`. In Release mode, no warning is produced though I'm not sure how the curve fit is successful. 

**To Reproduce**
<Update the following list with your specific information.>
1. Compile the `dev` branch in Debug
2. Run any of the following regression tests:
  - `AWT_YFix_WSt`
  - `AWT_WSt_StartUp_HighSpShutDown`
  - `AWT_YFree_WTurb`
  - `UAE_Dnwind_YRamp_WSt`
  - `WP_VSP_WTurb_PitchFail`
  - `SWRT_YFree_VS_EDG01`
4. Program exits with the array bound mismatch error shown below

**Expected behavior**
<A clear and concise description of what you expected to happen.>

Simulation runs to completion.

**Screenshots, if applicable**
<Add screenshots to help explain your problem.>

This is the backtrace produced by gfortran 

```
Output:
----------------------------------------------------------
At line 1187 of file /projects/openfast.worktrees/dev/modules/aerodyn/src/AirfoilInfo.f90
Fortran runtime error: Array bound mismatch for dimension 1 of array 'b' (2/1)

Error termination. Backtrace:
#0  0xffffaec9d08b in ???
#1  0xffffaec9ddb3 in ???
#2  0xffffaec9e2e7 in ???
#3  0xaaaae177686f in calculate_c_alpha
	at /projects/openfast.worktrees/dev/modules/aerodyn/src/AirfoilInfo.f90:1187
#4  0xaaaae177d3b3 in calculateuacoeffs
	at /projects/openfast.worktrees/dev/modules/aerodyn/src/AirfoilInfo.f90:1065
#5  0xaaaae178aa6b in readaffile
[...]

PR #2456, opened 2024-10-09
Add safety checks in AirfoilInfo for ill-defined airfoils

bjonkman:
When calculating UA parameters, we need to check that there are sufficient points to compute the least-squares solution. 

This can be a problem when the airfoil doesn't have enough points between -20 and 20 degrees angle of attack and/or when the shape of the polar is somewhat unusual. This fix should get around the issues we have been seeing in some unusual airfoils, but a more robust fix will come later.

https://github.com/OpenFAST/openfast/issues/2447

AeroDyn: AirfoilInfo and UnsteadyAero

This does not appear to change results in the current regression tests. I don't think the UA parameters that failed in debug mode were used in the simulations (based on the UA models that are used).


deslaughter (approved):

deslaughter:
@bjonkman Thanks for getting this put together!

PR #2457, opened 2024-10-09
Bug: AeroDyn-Inflow WriteOutput Value Ordering

deslaughter:
This PR is ready to merge.

This PR fixes a bug in the AeroDyn-Inflow library (`AeroDyn_Inflow.f90`) where the order of `WriteOutput` did not match `WriteOutputHdr` when `InflowWind` outputs were requested. `InflowWind` outputs were put at the end of `WriteOutput`, but the header indicated that `InflowWind` values should appear before `AeroDyn` values.

`AeroDyn_Inflow.f90`

This issue only occurs when using AeroDyn-Inflow directly or AeroDyn-Inflow C Bindings when outputs have been requested from InflowWind and AeroDyn. The AeroDyn Driver output is not affected as that program assembles its output array directly.

### DIFF
--- a/modules/aerodyn/src/AeroDyn_Inflow.f90
+++ b/modules/aerodyn/src/AeroDyn_Inflow.f90
@@ -299,9 +299,10 @@ subroutine ADI_CalcOutput(t, u, p, x, xd, z, OtherState, y, m, errStat, errMsg)
 
    ! --- Set outputs
    !TODO: this assumes one rotor!!!
-   associate(AD_NumOuts => p%AD%rotors(1)%NumOuts + p%AD%rotors(1)%BldNd_TotNumOuts)
-      y%WriteOutput(1:AD_NumOuts) = y%AD%rotors(1)%WriteOutput(1:AD_NumOuts)
-      y%WriteOutput(AD_NumOuts+1:p%NumOuts) = y%IW_WriteOutput(1:m%IW%p%NumOuts)
+   associate(AD_NumOuts => p%AD%rotors(1)%NumOuts + p%AD%rotors(1)%BldNd_TotNumOuts, &
+             IW_NumOuts => m%IW%p%NumOuts)
+      y%WriteOutput(1:IW_NumOuts) = y%IW_WriteOutput(1:IW_NumOuts)
+      y%WriteOutput(IW_NumOuts+1:p%NumOuts) = y%AD%rotors(1)%WriteOutput(1:AD_NumOuts)
    end associate
 
    !----------------------------------------------------------------------------


modules/aerodyn/src/AeroDyn_Inflow.f90
```diff
@@ -299,9 +299,10 @@ subroutine ADI_CalcOutput(t, u, p, x, xd, z, OtherState, y, m, errStat, errMsg)
 
    ! --- Set outputs
    !TODO: this assumes one rotor!!!
-   associate(AD_NumOuts => p%AD%rotors(1)%NumOuts + p%AD%rotors(1)%BldNd_TotNumOuts)
-      y%WriteOutput(1:AD_NumOuts) = y%AD%rotors(1)%WriteOutput(1:AD_NumOuts)
-      y%WriteOutput(AD_NumOuts+1:p%NumOuts) = y%IW_WriteOutput(1:m%IW%p%NumOuts)
+   associate(AD_NumOuts => p%AD%rotors(1)%NumOuts + p%AD%rotors(1)%BldNd_TotNumOuts, &
+             IW_NumOuts => m%IW%p%NumOuts)
+      y%WriteOutput(1:IW_NumOuts) = y%IW_WriteOutput(1:m%IW%p%NumOuts)
```
andrew-platt, line 304: You could change the second `m%IW%p%NumOuts` to `IW_NumOuts` as well.

andrew-platt (commented):
Looks good.

One modification you could do if you want.

modules/aerodyn/src/AeroDyn_Inflow.f90
deslaughter: That's a good point, will do.

bjonkman:
This is interesting. The AeroDyn driver uses the headers that ADI assembles. I wonder why the AD driver assembles this array a second time instead of using this one from ADI. It was an issue in https://github.com/OpenFAST/openfast/pull/2287.

andrew-platt:
I knew this issue sounded familiar yesterday when were diagnosing it for ADI c-bindings -- I couldn't find the issue number at the time.

The evolution of the driver and ADI library was convoluted with both @ebranlard and myself working on them.  During the writing of the c-bindings library, I had duplicated some code that was in the driver.  At that point we decided to collect some of the common pieces between the c-bindings and driver into the AeroDyn_Inflow.f90 library.  However neither of us fully completed revisions in the driver or c-bindings (it ended up at the bottom of the list of code cleanup priorities to do when we have nothing else to do), so I'm not surprised that the driver is handling this differently than c-bindings.

andrew-platt (approved):

Issue #2451, opened 2024-10-07
SubDyn: Lumped mass issues


RBergua:
**Bug description**
SubDyn behaves in an unexpected way when accounting for lumped masses.

**To Reproduce**
The unexpected behavior is not fully understood.

We have one model in SubDyn with beams and lumped masses. The model looks as follows:
![image](https://github.com/user-attachments/assets/69274cac-cd21-491e-a06e-b4fe083da90a)

The information about the mass, center of mass and moments of inertia was used to create an equivalent rigid floater in ElastoDyn.
![image](https://github.com/user-attachments/assets/803c321e-9156-487c-9bfc-4233641e2eb3)

The results of these 2 models (ElastoDyn rigid and SubDyn defined as rigid by means of SttcSolve = False and Nmodes = 0) was consistent in time domain for still water conditions:
![image](https://github.com/user-attachments/assets/df15f9fe-a2d7-471d-ab32-282a87cc51bd)

The agreement is very good between models specially considering that the ElastoDyn model includes 1 potential flow body in HydroDyn and the SubDyn model includes 6 potential flow bodies in HydroDyn (discretized system). 

The above SubDyn model used lumped masses with eccentricities. However, it is known that the lumped masses should not include eccentricities in SubDyn (see for reference: https://github.com/OpenFAST/openfast/issues/1710#issue-1826821882).

Accordingly, new joints were created in the SubDyn and linked to existing joints through rigid links. The model studied accounts for 22 lumped masses, so 22 rigid links were used. When doing this, the properties (e.g., mass, center of mass and moments of inertia) of the 2 SubDyn models (lumped mass with eccentricity vs lumped mass with rigid link) were consistent:
![image](https://github.com/user-attachments/assets/d17b30c5-3e25-43b4-9bad-fc93500b587b)

However, when comparing the results in time domain, there was a large difference in the behavior between models:
![image](https://github.com/user-attachments/assets/1530bc1f-10ef-4eec-a1cd-23fb34f16e63)

In order to try to better understand this behavior, the 22 lumped masses with 22 rigid links were removed from the SubDyn model and only 1 lumped mass with 1 rigid link was included. This unique lumped mass is equivalent to the 22 lumped masses in terms of mass, center of mass and moments of inertia. Since the structural model in SubDyn in this case is rigid, the response should be the same:
![image](https://github.com/user-attachments/assets/7bf17af7-10f0-42e8-97c9-cf99e0b7ba03)
 
[...]

Issue #1956, opened 2023-12-27
SubDyn: Floating system issues


RBergua:
**Bug description**
I'm simulating a floating system in SubDyn and the behavior is not the expected one. The floater looks as follows:
![image](https://github.com/OpenFAST/openfast/assets/33709140/c9de777b-fae7-4d47-b750-a6090911fc40)

The problem seems to come from the elements that connect the vertical stem to each leg. I have tried several elements available in SubDyn (e.g., cables, rigid links, and slender beams) and all of them return different results and the behavior is not the expected one.

First, I simulated the platform in ElastoDyn only considering a lumped mass and inertia at the proper center of mass for the floater. I did the same exercise in SubDyn with excellent agreement. And this is also aligned with the output that I obtained analytically. For reference, these 3 approaches take the mass, center of mass, and moments of inertia from the SubDyn summary file for the "detailed" design. 

The problem is that when I try to simulate with the "detailed" **floater as rigid (i.e., SttcSolve = False, Nmodes = 0 in SubDyn)**, the output is all over the place. Below you can see a plot with the most interesting outputs:
i. Reference: output from ElastoDyn (lumped mass and inertia), SubDyn (lumped mass and inertia), and analytical solution.
ii. SubDyn detailed with pretensioned cables (the cables have an initial pretension T0).
iii. SubDyn detailed with cables (the cables don't have initial pretension).
iv. SubDyn detailed with slender beams (the cables are removed and slender beams with equivalent axial stiffness and mass per unit of length are used).
v. SubDyn detailed with rigid link elements (the cables are removed and rigid links with the same mass per unit of length are used).
vi. SubDyn detailed without tilted elements connecting the vertical stem and the legs.

![image](https://github.com/OpenFAST/openfast/assets/33709140/80831d78-6ad0-4999-b628-f4c49e221bac)
 
As it can be observed, I can only recover the expected behavior if I get rid off all the elements connecting the vertical stem with the legs (24 elements in total).
 
Side note: in order to reduce possible differences coming from the glue code, I simply use a hydrostatic stiffness matrix (heave, roll, and pitch) in HydroDyn. There are no Morison elements. All the tests use the same hydrostatic approach.

Interestingly, if I look at the Guyan modes for these different SubDyn models, I can observe large differences. For this floating system, I would expect the Guyan modes being the six rigid body modes (0 Hz). However, this is not true for most SubDyn models with elements like rigid links, pretensioned cables... See below for reference:
i. Reference:
[...]

Issue #1710, opened 2023-07-28
SD: concentrated mass with excentricity cannot be fully accounted for in SubDyn


ebranlard:
Currently, SubDyn users can provide inertial properties of the concentrated mass such that the concentrated mass has an offset compared to the node location, using the three lasts inputs of the [concentrated mass table](https://openfast.readthedocs.io/en/dev/source/user/subdyn/input_files.html#joint-additional-concentrated-masses): `MCGX, MCGY, MCGZ`. 

The main issue is that the FEM model of SubDyn does not keep track of the center of mass offset, and therefore, cannot compute the Guyan load correction ("[extra moment](https://openfast.readthedocs.io/en/dev/source/user/subdyn/theory.html#corrections-to-the-baseline-formulation-guyanloadcorrection)") for this excentric mass. This will lead to inaccuracies in the results.

In the future, we will therefore discontinue this feature. 

The workaround is:
  - introduce a node at the location of the concentrated mass
  - connect this node with a rigid link to the closest node of the structure 

Small note: there is currently a small bug, where the gravity moment due to the excentricity of the concentrated mass is missing. It can be fixed with [this change](https://github.com/ebranlard/openfast/commit/f5abacd2066e177da46a56888dd5bf9e714e3237). But because in the future we will not support the excentricity any longer, there is no immediate need to fix it.

andrew-platt:
Fixed with #2458 and included in the 4.0.0 release.

Issue #2021, opened 2024-02-03
ElastoDyn vs SubDyn: Different behavior in floating systems


RBergua:
**Bug description**
I commented this issue in the past offline, but I leave it here for traceability.

When modeling a rigid platform in SubDyn (e.g., defining a rigid link and one lumped mass and inertia), the response is different than when a rigid platform in ElastoDyn is defined. This difference seems to show up when there is a modulation in the system response (e.g., two natural frequencies that are relatively close between them).

To illustrate this issue, the system has been simplified significantly:
![image](https://github.com/OpenFAST/openfast/assets/33709140/bb39d4ee-5ff6-4acf-8522-5acf9cc593d6)

Only three OpenFAST modules are used in this comparison: ElastoDyn, SubDyn and HydroDyn. HydroDyn is used to include the vertical force (equivalent to the buoyancy force) and a stiffness matrix (equivalent to the hydrostatic restoring). Below you can find the comparison between the platform being modeled in ElastoDyn vs the platform modeled in SubDyn vs the system simulated in HOTINT (EXUDYN).
![image](https://github.com/OpenFAST/openfast/assets/33709140/efe31153-943b-4a0c-8572-6eb93887fdb7)

As it can be observed, the agreement between ElastoDyn and HOTINT is excellent and SubDyn behaves differently.

The tests have been performed with the current OpenFAST dev branch ([e39453a](https://github.com/OpenFAST/openfast/commit/e39453af79726c5cb23dd6e2b2389da6b6dc2960)). 

This may be a good test case to run with the tight coupling ([#PR1850](https://github.com/OpenFAST/openfast/pull/1850/)) and check if the SubDyn response improves.

luwang00:
The discrepancy observed between the ElastoDyn model and the rigid SubDyn model is a consequence of the linearized equations of motion of the Guyan modes (rigid-body motion for floating) in SubDyn. This issue will be resolved as part of PR #2203.

PR #2458, opened 2024-10-09
SD: Bug fix for concentrated mass with CoG offset

luwang00:
This PR should be ready to merge after review.

This PR aims to fix a known bug with non-zero CoG offset/eccentricity of concentrated masses in SubDyn (Issue #1710) that was causing problems for some users (Issue #2451). The following fixes are implemented:
* The missing moments from the weight of offset concentrated masses are added during initialization when computing `p%FG` following the suggested fix of @ebranlard in Issue #1710.
* The node index, weight, and CoG offset of all concentrated masses are now saved in `p%CMassNode`. `p%CMassWeight`, and `p%CMassOffset` to support `GuyanLoadCorrection` for fixed-bottom structures and self-weight update for floating structures during the run.
* For fixed-bottom structures, an additional `GuyanLoadCorrection` term to the nodal moments is added. This term is in the form of `cross_product( matmul(p%CMassOffset(i,:),orientation)-p%CMassOffset(i,:), (/0.0,0.0,-p%CMassWeight(i)/) )`, where `orientation` is just the nodal orientation matrix due to the Guyan modes only. This additional correction term is zero if the offset/eccentricity is zero as expected.
* For floating structures, the additional force and moment from the weight of the concentrated masses are simply recomputed at each time step based on the Guyan (rigid-body) displacement of the structure. This follows naturally from the changes made in PR #2203.

This PR also adds a warning that SubDyn currently does not account for geometric stiffness from cable pretension if the user specifies one or more cables with non-zero pretension. See Issue #1956 and PR #2363.

Issues #1710, #2451, #1956

SubDyn

* There is no change to existing test results. All r-tests passed.
* An additional test was performed with a fixed-bottom structure having a concentrated mass with a large CoG offset. The results after the bug fix are effectively identical to those generated with the workaround proposed in Issue #1710, i.e., using a rigid link to connect the node for attachment to the concentrated mass CoG to avoid having non-zero eccentricity.
* A similar test with a floating structure also shows consistent results between the two approaches.

### DIFF
--- a/modules/subdyn/src/SD_FEM.f90
+++ b/modules/subdyn/src/SD_FEM.f90
@@ -1249,6 +1249,9 @@ SUBROUTINE AssembleKM(Init, p, ErrStat, ErrMsg)
    ENDDO
       
    ! Add concentrated mass to mass matrix
+   CALL AllocAry( p%CMassNode,   Init%nCMass,    'p%CMassNode',   ErrStat2, ErrMsg2); if(Failed()) return;
+   CALL AllocAry( p%CMassWeight, Init%nCMass,    'p%CMassWeight', ErrStat2, ErrMsg2); if(Failed()) return;
+   CALL AllocAry( p%CMassOffset, Init%nCMass, 3, 'p%CMassOffset', ErrStat2, ErrMsg2); if(Failed()) return;
    DO I = 1, Init%nCMass
       iNode = NINT(Init%CMass(I, 1)) ! Note index where concentrated mass is to be added
       ! Safety check (otherwise we might have more than 6 DOF)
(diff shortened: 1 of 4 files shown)


andrew-platt:
@RBergua, can you run a test of this?

jjonkman (approved):
I agree with the changes being implemented.  Please merge once you are happy with the tests.

RBergua:
I took advantage of the system documented here: https://github.com/OpenFAST/openfast/issues/2021, defined two lumped masses at (0, 0, 0) and defined two eccentricities. I'm able to get the same results as using the rigid link + lumped mass workaround. So, everything looks alright to me.

image

Maybe I would clarify in the documentation that when defining an eccentricity, the moments of inertia that the user provides are around the center of mass.

andrew-platt (approved):

PR #2470, opened 2024-10-17
Add blade distributed load output to AeroDyn-InflowWind C bindings

deslaughter:

This PR adds a flag, `PointLoadOutput`, to the `ADI_C_PreInit` subroutine in the ADI C Bindings library `AeroDyn_Inflow_C_Binding.f90` which allows the user to select between point load or distributed load output at the blade mesh points. Setting this value to `0` will cause distributed loads to be output and setting it to `1` will output point loads, which is the typical output from this library.

This PR also adds the parameter `HHVel_C`, a 3 element float array, to `ADI_C_GetRotorLoads` which will contain the hub height wind velocity components (X,Y,Z), if `storeHHVel` was set to `1` in `ADI_C_Init`; otherwise, it will contain zeros. Previously, the velocities would be stored, but were not accessible through the API.

The python library, `aerodyn_inflow_library.py`, was updated to reflect the changes in the API.

`AeroDyn_Inflow_C_Binding.f90` and `aerodyn_inflow_library.py`

Test results were unaffected.

### DIFF
--- a/modules/aerodyn/python-lib/aerodyn_inflow_library.py
+++ b/modules/aerodyn/python-lib/aerodyn_inflow_library.py
@@ -2,7 +2,7 @@
 # LICENSING
 # Copyright (C) 2021 National Renewable Energy Laboratory
 #
-# This file is part of InflowWind.
+# This file is part of AeroDyn.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


andrew-platt (approved):
Looks good!  Thanks for making that change!

PR #2478, opened 2024-10-23
[BugFix] Incorrect value of AirDens passed to ExtInfw_Init from FAST_Subs.f90

andrew-platt:
Ready to merge

Due to a logic error, the `AirDens` value passed to _ServoDyn_ and _External-Inflow_ was set to zero. The logic around zeroing out the `AirDens` value was corrected.

Also I noticed that _AeroDisk_ was not returning a value for `AirDens` during initialization.  Since _AeroDisk_ can use its own value instead of the value set in the main .fst file, this value should be returned so that _ServoDyn_ has the correct air density.

None (issue found by @marchdf and notified us internally)

This primarily affects the `ExtInfw` module used for actuator line coupling with Nalu-Wind, but could potentially show up with any controller that uses the air density (none of our test controllers use it).

None: our test controllers do not use the air density.


andrew-platt:
Error introduced by either PR #1932 or PR #2445

PR #2516, opened 2024-11-14
AD bugfix: Segmentation fault with ifx

andrew-platt:
Ready to merge, unless someone has a more elegant solution

When OLAF was used with the AeroDyn Driver compiled with ifx (IFX) 2023.2.0 20230622 (release only), there would be a segmentation fault when SetInputsForFVW was called.  The root issue was that passing an array of `u` as `(/u/)` doesn't work correctly with this compiler.  So to work around this, the SetInputsForFVW routine was reworked to only operate on a single `u`.

In my opinion this is a hack of a solution to accomodate a compiler bug.

none

AeroDyn driver, only when compiled with ifx

This would trigger a segmentation fault in the `ad_HelicalWakeInf_OLAF` in 3.5.5 with  ifx (IFX) 2023.2.0 20230622 in release.

### DIFF
--- a/modules/aerodyn/src/AeroDyn.f90
+++ b/modules/aerodyn/src/AeroDyn.f90
@@ -1623,8 +1623,10 @@ subroutine AD_UpdateStates( t, n, u, utimes, p, x, xd, z, OtherState, m, errStat
 
    else  ! Call the FVW sub module
          ! This needs to extract the inputs from the AD data types (mesh) and copy pieces for the FVW module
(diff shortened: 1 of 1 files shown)


deslaughter:
I think it’s actually a good fix as that syntax was dynamically creating an
array and copying u into it at every call, which is probably why it caused
a segfault. If something in u can’t be directly copied, then it could be
doing anything.

andrew-platt:
That's probably the issue: something in `u` couldn't be copied directly.

deslaughter (approved):

PR #2532, opened 2024-11-24
bugfix: IfW rotor points for disk average incorrect

andrew-platt:
Ready to merge

The calculations for the rotor disk average wind speed use a set of point projected on the disk perpendicular to the hub at ~0.7 rotor radius in the hub YZ plane.  However, the indices for this calculation were incorrect and projected the points onto disk in the hub XY plane.  This was discovered after one of the cases used in the curled wake paper failed to run correctly with 3.5.4 for a turbine very close to the back edge of the high resolution domain.

Also added some additional information about the location of points causing outside box bound errors.

- https://forums.nrel.gov/t/comparison-of-fast-farm-and-sowfa/7684/16
- This issue was introduced somewhere in PR #1240

FAST.Farm uses the disk average velocity for some calculations, so some results may change very slightly.

This was found while attempting to figure out why it was not possible to reproduce some of the curled wake results from a publication.

Some results may change.  Those will be documented below.

### DIFF
--- a/modules/inflowwind/src/InflowWind_Subs.f90
+++ b/modules/inflowwind/src/InflowWind_Subs.f90
@@ -1047,9 +1047,9 @@ SUBROUTINE InflowWind_SetParameters( InitInp, InputFileData, p, m, ErrStat, ErrM
    
    do i=1,IfW_NumPtsAvg
       theta = pi +(i-1)*TwoPi/IfW_NumPtsAvg
-      p%PositionAvg(1,i) = R*cos(theta)
-      p%PositionAvg(2,i) = R*sin(theta)
-      p%PositionAvg(3,i) = 0.0_ReKi
+      p%PositionAvg(1,i) = 0.0_ReKi          ! Hub X (perpindicular to rotor plane)
+      p%PositionAvg(2,i) = R*cos(theta)      ! Hub Y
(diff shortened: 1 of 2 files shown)


modules/inflowwind/src/InflowWind_Subs.f90
```diff
@@ -1047,9 +1047,9 @@ SUBROUTINE InflowWind_SetParameters( InitInp, InputFileData, p, m, ErrStat, ErrM
    
    do i=1,IfW_NumPtsAvg
       theta = pi +(i-1)*TwoPi/IfW_NumPtsAvg
-      p%PositionAvg(1,i) = R*cos(theta)
-      p%PositionAvg(2,i) = R*sin(theta)
-      p%PositionAvg(3,i) = 0.0_ReKi
+      p%PositionAvg(1,i) = 0.0_ReKi          ! Hub X (perpindicular to rotor plane)
+      p%PositionAvg(2,i) = R*cos(theta)      ! Hub Y
+      p%PositionAvg(3,i) = R*sin(theta)      ! Hub Z (in vertical plane when azimuth=0)
```
deslaughter, line 1052: The 'ifw_turbsimff` regression test is failing because it's using `R = max(1.0_ReKi, InputFileData%Uniform_RefLength)/2.0_ReKi` which is data from the wrong wind type. Maybe a rotor diameter field could be added the InflowWind driver? Or the driver could guess the rotor diameter to from the flow field size.

Issue #2510, opened 2024-11-12
Unable to run FAST.Farm simulation


j-killeen:
Dear @jjonkman 

I hope this message finds you well. I am trying to run a wind farm simulation that contains 2 5MW_OC3Spar_DLL_WTurb_WavesIrr offshore fst files. When I run the simulation, I am constantly getting the following error message: 

"forrtl: severe (47): write to READONLY file, unit 13, file C:\Users\jamie\Downloads\r-test-548bf2143633a7066f4582167fdf4180951fd7f6 (1)\r-test-548bf2143633a7066f4582167fdf4180951fd7f6\glue-codes\openfast\5MW_OC3Spar_DLL_WTurb_WavesIrr\fort.13
Image              PC                Routine            Line        Source
FAST.Farm_x64_OMP  00007FF724AF688F  Unknown               Unknown  Unknown
FAST.Farm_x64_OMP  00007FF724AB7993  Unknown               Unknown  Unknown
FAST.Farm_x64_OMP  00007FF7236A4A3C  Unknown               Unknown  Unknown
FAST.Farm_x64_OMP  00007FF7236A2AC6  Unknown               Unknown  Unknown
FAST.Farm_x64_OMP  00007FF7236A2A5B  Unknown               Unknown  Unknown
FAST.Farm_x64_OMP  00007FF723698219  Unknown               Unknown  Unknown
FAST.Farm_x64_OMP  00007FF72369672C  Unknown               Unknown  Unknown
FAST.Farm_x64_OMP  00007FF7231052AB  Unknown               Unknown  Unknown
FAST.Farm_x64_OMP  00007FF723263ECA  Unknown               Unknown  Unknown
libiomp5md.dll     00007FFCCEC2A853  Unknown               Unknown  Unknown
libiomp5md.dll     00007FFCCEB7A617  Unknown               Unknown  Unknown
libiomp5md.dll     00007FFCCEB7A0DC  Unknown               Unknown  Unknown
libiomp5md.dll     00007FFCCEC025C7  Unknown               Unknown  Unknown
KERNEL32.DLL       00007FFD275353E0  Unknown               Unknown  Unknown
ntdll.dll          00007FFD27EE485B  Unknown               Unknown  Unknown" 

I don't know if this is a bug issue as I'm using FAST.Farm version 3.5.1. The simulation runs normally without any errors if I remove the second wind turbine, making it only one turbine in the simulation. Any advice or guidance on how to fix this issue would be appreciated. 

[...]

PR #2538, opened 2024-11-28
Move calls to GetNewUnit into file opening routines and wrap with !$OMP critical

andrew-platt:
This is ready to merge

We have had issues with unit numbers when opening files for many years when _OpenMP_ is used.  The root issue was that the `GetNewUnit` call was separate from the actual file opening.  This could lead to race conditions when multiple threads hit a `GetNewUnit` before the retrieved unit numbers were used to open the corresponding files.

To solve this, we moved all the `GetNewUnit` calls into the `Open*File` subroutines in `NWTC_IO.f90`.  Inside those routines we wrap the `GetNewUnit` and `open(*)` command pairs in a `!$OMP critical` block.  This should prevent the same unit number from being given out to two different _OpenMP_ threads at once. 

We debated if this should go into 3.5.5 or 4.0.0.  In the end we decided this would be better to address in 3.5.5, mostly so that I don't have to field any more questions or bug reports about it when users stay on the 3.5.x series longer.

#2510
There were many others over the years that we suspect may be related.

This change should be completely transparent to the end users, but should fix issues with conflicting unit numbers when _OpenMP_ is used.

We originally thought about wrapping the original `GetNewUnit` calls and the following `Open*File` calls, but decided that would complicate the code considerably.  Instead we decided it was simpler to move the `GetNewUnit` calls and handle the `!$OMP critical` block in the library so future programmers would not need to think about it.


modules/nwtc-library/src/VTK.f90
```diff
@@ -360,10 +356,7 @@ SUBROUTINE WrVTK_SP_header( FileName, descr, Un, ErrStat, ErrMsg )
       INTEGER(IntKi)  , INTENT(  OUT)        :: ErrStat              !< error level/status of OpenFOutFile operation
       CHARACTER(*)    , INTENT(  OUT)        :: ErrMsg               !< message when error occurs
    
-      !$OMP critical
-      CALL GetNewUnit( Un, ErrStat, ErrMsg )
```
bjonkman, line 364: For backward compatibility, could you wrap this call to `GetNewUnit` (here and the other places it's now called) in an `IF` statement like `IF (Un < 0) call GetNewUnit()`?

Also note that the `ErrStat` and `ErrMsg` variables from the calls to `GetNewUnit` are now ignored.

modules/nwtc-library/src/NWTC_IO.f90
```diff
@@ -2337,18 +2331,21 @@ SUBROUTINE OpenBInpFile ( Un, InFile, ErrStat, ErrMsg )
...
-   IF ( ErrStat /= 0 ) THEN
-      ErrStat = ErrID_Fatal
-      ErrMsg  = 'OpenBInpFile:Cannot open file "'//TRIM( InFile )//'" for reading. Another program may have locked it.'
-   ELSE
-      ErrStat = ErrID_None
-      ErrMsg  = ''
-   END IF
-
+   !$OMP critical
+   call GetNewUnit(Un,ErrStat,ErrMsg)
```
bjonkman, line 2336: For backward compatibility, could you wrap this call to GetNewUnit (here and the other places it's now called) in an IF statement like IF (Un < 0) call GetNewUnit()?

bjonkman: Alternatively,  `GetNewUnit` could be changed to start looking at units at the given `Un` value if it's >= 0. Otherwise, `Un` is really `intent(out)` only.

andrew-platt: I'll wrap those calls in an `if` statement.  I think we want to maintain that backwards compatibility for now.

andrew-platt:
After discussing this with @deslaughter, we concluded that there are to many code changes going into this PR to make sense for a 3.5.5 release.

I am closing this PR and will open a new one to `dev` with a reworked version of these changes.

PR #2569, opened 2024-12-18
Disable LiDAR in IfW at FAST.Farm level

andrew-platt:
Ready to merge

The _LiDAR_ sub-module inside _InflowWind_ requires turbine specific information such as hub location.  However, this information is not available at the _FAST.Farm_ level instance of _InflowWind_.  This was resulting in segmentation faults when someone attempted to use it at this level.

This PR disables the _LiDAR_ submodule from use with the _InflowWind_ module used at the _FAST.Farm_ level.

#2316 
#2324 

_InflowWind_ module at the _FAST.Farm_ level only.

A few users have attempted to use the _LiDAR_ sub-module at the _FAST.Farm_ level and received segmentation faults.  Since this is not something the code can really do, we are simply disabling it for now.

No results change.

### DIFF
--- a/modules/inflowwind/src/InflowWind.f90
+++ b/modules/inflowwind/src/InflowWind.f90
@@ -224,7 +224,16 @@ SUBROUTINE InflowWind_Init( InitInp, InputGuess, p, ContStates, DiscStates, Cons
    p%lidar%PulseSpacing = InputFileData%PulseSpacing
    p%lidar%URefLid = InputFileData%URefLid
    p%lidar%ConsiderHubMotion = InputFileData%ConsiderHubMotion  
-         
+
+   ! Disable Lidar if not allowed (FAST.Farm doesn't allow this)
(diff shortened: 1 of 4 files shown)


deslaughter (commented):
I probably would have reversed the logic and done `LidarEnable = .true.` in the registry file and set it to `.false.` where it's initialized in FAST.Farm. Otherwise, the flag will have to be set to false to achieve normal operation. For example, does this flag need to be set false in the ADI library or the InflowWind C Bindings?

andrew-platt:
At present the Lidar doesn't output anything to file, but simply populates data to pass to the controller.  In the ADI library and InflowWind c-bindings library, there are no provisions for a controller so Lidar shouldn't be used there.

I can change the logic on that, but I do thing we should default it to disabled.

deslaughter (approved):

PR #2575, opened 2024-12-19
ADsk: correction to disk average velocity equations

andrew-platt:
Merge after 3.5.5 cross merge (rebase then).

_AeroDisk_ calculations were using an incorrect set of points for the `DiskAvgVel` calculations. The set of points used in the previous calculation was aligned with the hub axis instead of perpendicular to the hub axis.

See PR #2532 for details

_AeroDisk_ only

Merge this after cross-merging `main` into `dev` after the 3.5.5 release.

Test results will change slightly with this change.

- [ ] r-test merge required


andrew-platt:
Some regression test changed significantly:

## 5MW_Land_DLL_WTurb_ADsk
Screenshot 2024-12-22 at 1 32 22 PM
Figure1: Disk average wind velocity has changed significantly as the sampling area is now in the rotor disk plane.  This gives slightly lower, but smoother, average wind speeds than when the disk averaging points were located in the wrong plane (circle lying in the rotating hub X-Y plane, so it was spinning like a circular egg beater in front of the rotor which gave the periodic spikes).

Screenshot 2024-12-22 at 2 10 19 PM
Figure 2: blade pitch changes as rotor loads change.  The new results track slightly better than the original.


Screenshot 2024-12-22 at 2 11 03 PM
Figure 3: GenPwr comparison with `5MW_Land_DLL_WTurb` case.  This is closer to the expected full OpenFAST simulation with AeroDyn.

deslaughter (approved):

PR #2361, opened 2024-08-07
Updates to openfast_io to match current dev branch API

mayankchetan:
This PR is NOT ready to merge until TODOs below are cleared.

**Feature or improvement description**

Updates to the `openfast_io` library to match the current `dev` branch API. Once merged, will provide OpenFAST users with an up-to-date library that they can import to read/write OpenFAST files. This follows the `fst_vt` nested dictionary structure developed under `WEIS` and heavily borrows code from there. 

This library is currently limited to `OpenFAST` and does not work with `FAST.Farm` or any of the module drivers.

Related PRs: https://github.com/OpenFAST/openfast/pull/2333, https://github.com/OpenFAST/openfast/pull/2188

**Related issue, if one exists**

https://github.com/OpenFAST/openfast/issues/2233, https://github.com/OpenFAST/openfast/issues/2261, https://github.com/OpenFAST/openfast/issues/1851

**Impacted areas of the software**

openfast_io

**Additional supporting information**

**Test results, if applicable**

`pytest` added to span most OpenFAST input files (drivers not included). Based on @andrew-platt comments in https://github.com/OpenFAST/openfast/issues/2233, need to introduce automated testing.

TODO:

- [x] Add test automations & GH actions
- [ ] Create an OpenFAST [PyPI](https://pypi.org/) account so the openfast_io python package can be installed with pip
- [x] Set up a [trusted publisher](https://docs.pypi.org/trusted-publishers/) to securely upload the python package to PyPI
- [ ] Setup conga-forge for `openfast_io` since most NREL tools like `WEIS` and `WISDEM` prefer it.
- [x] Check if we can eliminate the `Poetry` file ~~We will continue using `Poetry`~~
- [x] Add additional tests based on recommendations in this PR if any
- [x] Address `old API` warning thrown by AD15
- [x] Add simple documentation to help users get started
- [ ] Add badges for PyPi and conda-forge

APIs to upgrade before merge:
- [ ] ~~https://github.com/OpenFAST/openfast/pull/986~~
- [x] https://github.com/OpenFAST/openfast/pull/1295
- [x] https://github.com/OpenFAST/openfast/pull/2203
- [x] https://github.com/OpenFAST/openfast/pull/2358
- [x] https://github.com/OpenFAST/openfast/issues/93
- [x] https://github.com/OpenFAST/openfast/pull/2401
- [x] https://github.com/OpenFAST/openfast/pull/2561

PR #2577, opened 2024-12-21
Refactor openfast/openfast_python/openfast_io/turbsim_file.py

Gjacquenot:

Refactor openfast/openfast_python/openfast_io/turbsim_file.py:

- [x] add an entrypoint to the script
- [x] simplify condition 

None

None

None

Tested manually on a bts file found in the submodule

### DIFF
--- a/openfast_python/openfast_io/turbsim_file.py
+++ b/openfast_python/openfast_io/turbsim_file.py
@@ -5,6 +5,7 @@ copied with love from the python-toolbox
 """
 import pandas as pd
 import numpy as np
+import argparse
 import os
 import struct
 import time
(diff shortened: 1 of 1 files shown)


Gjacquenot:
I have tested it with the wind bts file from [r-test repository](https://github.com/OpenFAST/r-test/blob/5368d3122a58d1b9ae8724d816e43c710b19bf90/modules/inflowwind/ifw_BoxExceed/wind.bts)

Then the command is the following, it sums up the content of the two Python pandas dictionaries

```bash
cd openfast_python 
python3 -m openfast_io.turbsim_file wind.bts         
```

andrew-platt:
@Gjacquenot, thanks for your reply.  I think we are missing some context here.  Could you explain why you want to do this as opposed to loading the `TurbSimFile` class and exploring the metadata that way?

mayankchetan (approved):
Looks good for now. Will update the summary in `main()` to use `__repr__` in #2361

PR #2584, opened 2024-12-23
F/md shared example

andrew-platt:
Ready to merge

This PR adds visualization of shared mooring lines from _MoorDyn_ at the _FAST.Farm_ level.  This is a feature that several have requested.

Visualization of shared moorings is at the _FAST.Farm_ timestep rate

NOTE: there are input files.  There is a new line in the _FAST.Farm_ input file for the `WrMooringVis` flag:
```
3                 Mod_SharedMooring  - Shared mooring system model (switch) {0: None, 3=MoorDyn}}
--- SUPER CONTROLLER --- [used only for UseSC=True]
"unused"          SC_FileName        - Name/location of the dynamic library {.dll [Windows] or .so [Linux]} containing the Super Controller algorithms (quoated string)
--- SHARED MOORING SYSTEM --- [used only for Mod_SharedMoor>0]
"SharedMooring2.dat" SharedMoorFile  - Name of file containing shared mooring system input parameters (quoted string) [used only when Mod_SharedMooring > 0]
0.05              DT_Mooring         - Time step for farm-level mooring coupling with each turbine (s) [used only when Mod_SharedMooring > 0]
true              WrMooringVis       -  Write shared mooring visualization, at the global FAST.Farm time step (-) [only used for Mod_SharedMooring=3]
--- AMBIENT WIND: PRECURSOR IN VTK FORMAT --- [used only for Mod_AmbWind=1]
```

None

Visualization from _FAST.Farm_ only

**Known issues with this PR**
- the width of the VTK file time code is hardcoded to 5 instead of calculated
- if visualization of OpenFAST platforms is also done, the numbering of the VTK files will not correspond.  This is really a question about changing how the coding of time of VTK files is done - that's a different discussion though.

New test case for two IEA 15 MW turbines in a shared mooring system.  This case highlights that there is an issue with the initialization of the mooring lines (one turbine is rotated by 180 degrees, but this rotation is not factored into the initial relaxation calculations).

Screenshot 2024-12-23 at 12 14 29 AM
Figure 1: Turbine 1, on the left, has a platform rotation of 180 degrees.  _MoorDyn_ is calculating the initial relaxation of these mooring lines with the platform rotation of 0 degrees, so the lines do not render well on the first timestep.

- [ ] r-test merge required


deslaughter (commented):
Looks like there's an issue with the file naming for the `MD_Shared` regression test. The test failed looking for `FAST.Farm.out`, but it generated a `FAST.Farm.FarmMD.MD.out` containing MoorDyn data.

andrew-platt:
Note: I added the `WrMooringVis` to the shared mooring section instead of the visualization section.  I think it is a little less confusing to end users this way.

docs/source/user/fast.farm/InputFiles.rst
```diff
@@ -139,7 +139,7 @@ documentation for details on the input file at the farm level.
 MoorDyn. 
 
 **WrMooringVis** [swithch]   Write shared mooring line visualization, at
-DT_Mooring timestep
+the globale FAST.Farm time step
```
deslaughter: `globale` -> `global`

andrew-platt: fixed in  [eaac346](https://github.com/OpenFAST/openfast/pull/2584/commits/eaac34606533b8f626e6429467196bca4b4b4736)

RyanDavies19:
@andrew-platt, lets chat next week about the initialization issue here. I have a suspicion about what's causing it

PR #2628, opened 2025-02-10
Fix bug in ExternalInflow Orientation indexing

deslaughter:

An issue was discovered with the indexing in ExternalInflow which populates the Orientation data that is sent to AMR-Wind. Effectively a `+1` was replaced with a `_1` and instead of throwing a compilation error, this was interpreted as a type specification. This lead to an off-by-one indexing error in the orientation matrices. This PR fixes this issue and adds a new regression test which exercises the ExternalInflow module.

ExternalInflow and regression tests.

New test was added and is passing.


andrew-platt (approved):
Thanks for adding a test as well!

PR #2632, opened 2025-02-11
Reduce optimization level for FAST*_Types.f90 with gcc

andrew-platt:
Ready to merge

Compiling in release with GCC takes a very long time due to unnecessary optimization of the `FAST_Types.f90` and `FAST_Farm_Types.f90` files.  These files contain a fair amount of code related to save/restore used in restarts, which really doesn't need to be all that optimized given the frequency of use. 

Some time ago we benchmarked the build process with `-O2` and `-O3` on `FAST_Types.f90` and found that the majority of the build time in that file was spent on the save/restore routines (tested with Intel OneAPI on linux).  Reducing to `-O2` reduced the compile time significantly in that case.

Can't find the first PR.

Build process with GCC.  May be slight performance hit on save/restore with GCC compiled executables, but no other effects expected.


deslaughter (approved):

andrew-platt:
Backport to 4.1.0 if we do such a release before 5.0

PR #2636, opened 2025-02-13
Add job to build Windows executables to deploy.yml

deslaughter:
This PR is ready merge.

This PR adds a job to the GitHub CI (`deploy.yml`) which will automatically build the full suite of Windows executables from the `vs-build` directory on release. This workflow can also be run manually which will allow for spot checking of the Visual Studio build without needing to build locally.

`deploy.yml` GitHub CI

This build uses the `ifort` compiler and Visual Studio. Once the `ifx` compiler is fully vetted, the build can switch to it. Until then, the build output will show warnings about the depreciation of `ifort`.


vs-build/OpenFAST-Simulink/OpenFAST-Simulink.vfproj
```diff
@@ -16,7 +16,7 @@
 				<Tool Name="VFManifestTool" SuppressStartupBanner="true"/></Configuration>
 		<Configuration Name="Release_Matlab|Win32" OutputDirectory="..\..\build\bin\" TargetName="$(ProjectName)_$(PlatformName)" ConfigurationType="typeDynamicLibrary">
 				<Tool Name="VFFortranCompilerTool" SuppressStartupBanner="true" Preprocess="preprocessYes" PreprocessorDefinitions="COMPILE_SIMULINK" StandardWarnings="standardWarningsF03" DisableSpecificDiagnostics="5268,5199"/>
-				<Tool Name="VFLinkerTool" SuppressStartupBanner="true" GenerateManifest="false" SubSystem="subSystemWindows" LinkDLL="true" AdditionalDependencies="&quot;C:\Program Files (x86)\MATLAB\R2018a\extern\lib\win32\microsoft\libmex.lib&quot;"/>
+				<Tool Name="VFLinkerTool" SuppressStartupBanner="true" GenerateManifest="false" SubSystem="subSystemWindows" LinkDLL="true" AdditionalDependencies="&quot;$(MATLAB_ROOT)\extern\lib\win32\microsoft\libmex.lib&quot;"/>
```
andrew-platt, line 19: That's a good way to do it.  I wasn't aware of that variable

deslaughter: Well, it's an environment variable that the user will need to define, but I still think this is better than hard coding the path into the Visual Studio project. I wonder what user's see if they don't have that path. It's not like everyone uses MATLAB 2018a so they should be used to setting it anyway.

.github/scripts/build_windows_executables.bat
```diff
@@ -0,0 +1,41 @@
+@call "C:\Program Files (x86)\Intel\oneAPI\setvars-vcvarsall.bat" %VS_VER%
+
+for /f "tokens=* usebackq" %%f in (`dir /b "C:\Program Files (x86)\Intel\oneAPI\compiler\" ^| findstr /V latest ^| sort`) do @set "LATEST_VERSION=%%f"
+@call "C:\Program Files (x86)\Intel\oneAPI\compiler\%LATEST_VERSION%\env\vars.bat"
+
+@REM Make git ignore changes to Types files (line endings) and 
+@REM vs-build directory so it doesn't append -dirty to version
+ECHO *_Types.f90>>".gitignore"
```
andrew-platt, line 8: Another option:
`git update-index --assume-unchanged vs-build */*/src/*_Types.f90`

vs-build/OpenFAST-Simulink/OpenFAST-Simulink.vfproj
andrew-platt: Right now if the user has a different version, it fails on the linking step.

.github/scripts/build_windows_executables.bat
deslaughter: I tried a variety of things here and finally just added a powershell command to remove `--dirty` from the `vsbuild/CreateGitVersion.bat` script before building the VS projects. I also added a section to `deploy.yml` to run `openfast_x64.exe -h` and output the version so we can see it in the logs. I think it looks correct now for this branch as `OpenFAST-v4.0.2-34-gb3410475` since it's not actually on the tagged commit. We'll see what it does on the next release.

andrew-platt: I take it the `git update-index` didn't work as expected (maybe it doesn't work the same way through bat scripts as in a bash/git shell).

The new approach should work.  The git hash should disappear when on a tagged commit.

andrew-platt (approved):
Looks good enough for now :)  Let's see what it does on the next release.

PR #2642, opened 2025-02-18
[BugFix] BD: variable passing in Calc_RR0mEta_rho

### LINKED ISSUE #2641: BeamDyn subroutine argument passing
While building the dev branch, today, my compiler warned

> BeamDyn.f90([2673](https://github.com/OpenFAST/openfast/blob/e85ee494d63b3aab6b45fcc6251f5b27ef949c34/modules/beamdyn/src/BeamDyn.f90#L2673)): warning #6843: A dummy argument with an explicit INTENT(OUT) declaration is not given an explicit value.   [RR0META]

The code looks a little unusual. The [calling code](https://github.com/OpenFAST/openfast/blob/e85ee494d63b3aab6b45fcc6251f5b27ef949c34/modules/beamdyn/src/BeamDyn.f90#L2665)
is sending `m%qp%RR0mEta(:,idx_qp,nelem)` as an intent(out) variable (which Fortran could fill with zeros):
https://github.com/OpenFAST/openfast/blob/e85ee494d63b3aab6b45fcc6251f5b27ef949c34/modules/beamdyn/src/BeamDyn.f90#L2675

But then instead of using the `Calc_RR0mEta_rho()` routine's argument, it sets that same portion of `m%qp` inside the routine
https://github.com/OpenFAST/openfast/blob/e85ee494d63b3aab6b45fcc6251f5b27ef949c34/modules/beamdyn/src/BeamDyn.f90#L2679

This should be changed to either (1) use the routine's `RR0META` variable instead of `m%qp%RR0mEta(:,idx_qp,nelem)` or (2) remove the routine's `RR0META` variable

andrew-platt:
Ready to merge

In the  `BD_Calc_RR0mEta_rho` routine, the `RR0mEta` output was not getting set, but instead the `m%qp%RR0mEta` was set directly.  However the `RR0mEta` value was passed out of the routine directly to `m%qp%RR0mEta`, so it becomes ambiguous if the compiler would set that value through the routine, or set it to the uninitialized value of `RR0mEta`.

Issue #2641 outlines the issue in more detail.

_BeamDyn_, but likely only with specific compilers (most seem to handle this odd code).

closes #2641

### DIFF
--- a/modules/beamdyn/src/BeamDyn.f90
+++ b/modules/beamdyn/src/BeamDyn.f90
@@ -2662,7 +2662,8 @@ SUBROUTINE BD_QPData_mEta_rho( p, m )
    do nelem = 1, p%elem_total
       qp_start = (nelem-1)*p%nqp
       do idx_qp = 1, p%nqp
-         call Calc_RR0mEta_rho(m%qp%RR0(:,:,idx_qp,nelem), &
+         call Calc_RR0mEta_rho(p%qp%mEta(:,idx_qp,nelem), &
+                               m%qp%RR0(:,:,idx_qp,nelem), &
                                p%Mass0_QP(:,:,qp_start+idx_qp), &
                                m%qp%RR0mEta(:,idx_qp,nelem), &
                                m%qp%rho(:,:,idx_qp,nelem))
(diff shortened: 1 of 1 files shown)


andrew-platt:
- [ ] Backport to 4.0.3 if we do such a release.

modules/beamdyn/src/BeamDyn.f90
```diff
@@ -2676,7 +2676,7 @@ subroutine Calc_RR0mEta_rho(RR0, Mass0, RR0mEta, rho)
 
          !> Calculate the new center of mass times mass at the deflected location
          !! as \f$ \left(\underline{\underline{R}}\underline{\underline{R}}_0\right) m \underline{\eta} \f$
-         m%qp%RR0mEta(:,idx_qp,nelem) = MATMUL(RR0, p%qp%mEta(:,idx_qp,nelem))
+         RR0mEta(:,idx_qp,nelem) = MATMUL(RR0, p%qp%mEta(:,idx_qp,nelem))
```
deslaughter, line 2679: Thanks for fixing this, looks like I forgot to remove the `m%qp` from `RR0mEta` inside the subroutine.

```suggestion
         RR0mEta(:) = MATMUL(RR0, p%qp%mEta(:,idx_qp,nelem))
```

PR #2650, opened 2025-02-25
MoorDyn: VIV and Rk4

RyanDavies19:
This PR in introduces a cross-flow VIV model based on the work of [Thorsen](https://ntnuopen.ntnu.no/ntnu-xmlui/handle/11250/2414408) in conjunction with FloatingArrayDesign/MoorDyn#290 (docs update there). A paper is in the works the details the theory. When released, it will be linked here. 

The semi-empirical model gives an oscillating lift force on each node. As part of this implementation, there is a synchronization model that looks at changes between the lift force phase and line motion phase. This model requires knowledge of the lines current acceleration (which we approximate by the previously calculated value) and an additional state variable that is integrated every time step. We validated this approach against the results in the original theory paper using a newly implemented rk4 and verified that it matches the implementation in MD-C.

`OpenFAST_IO` has also been updated to accommodate the new input file format. TODO's were added to OpenFAST IO for future work, it needs to accommodate the line viscoelastic capability, line failures section, and user specified external loads and damping section. 

Additional error checks have been added to the viscoelastic model.

MoorDyn
OpenFAST_IO


RyanDavies19:
The failing tests are a little perplexing, the changes here do not impact the MD features those tests use. In the StC_test_OC4Semi results, it looks like a very minor error. Not sure what's going on in the linearized testing, there's no nice HTML files to look at.

andrew-platt:
Changing to merge to `dev` branch.

andrew-platt:
I suspect the RK4 solve is changing the `StC_test_OC4Semi` very slightly due to the large motion transient when the StC forces zero out.

For the linearization case, the number of states has changed.  This will change the size of the linearization matrices which will automatically fail.  I'll check that the frequency information still makes sense and double check that the state names are handled correctly in the linearization code in MD (there might be something missing there).

andrew-platt:
@deslaughter: there are linearization changes here that will need incorporation into the tight-coupling when merging to `dev-TC`.

RyanDavies19:
> I suspect the RK4 solve is changing the `StC_test_OC4Semi` very slightly due to the large motion transient when the StC forces zero out.
> 
> For the linearization case, the number of states has changed. This will change the size of the linearization matrices which will automatically fail. I'll check that the frequency information still makes sense and double check that the state names are handled correctly in the linearization code in MD (there might be something missing there).

I thought about that, but RK2 is default solver unless the user says otherwise. See: https://github.com/RyanDavies19/MoorDynF_ryan/blob/1c883b541da4f91c84cd4d072dd8cc8926852ad9/modules/moordyn/src/MoorDyn.f90#L120

Will double check Rk2 is still used by adding MD log files.

RyanDavies19:
> @deslaughter: there are linearization changes here that will need incorporation into the tight-coupling when merging to `dev-TC`.

I will add an error check that throws fatal if users try to linearize with VIV. @deslaughter this conversation did highlight the optional viscoelastic state is not currently handled in the linearization (see: https://github.com/OpenFAST/openfast/blob/fc1110183bcc87b16d93129edabdce6d30e3a497/modules/moordyn/src/MoorDyn.f90#L1446). The easiest approach would probably be to throw another fatal error if users try to use that as well.

openfast_io/openfast_io/FAST_reader.py
```diff
@@ -2921,6 +2924,18 @@ def read_MoorDyn(self, moordyn_file):
...
                     self.fst_vt['MoorDyn']['CaAx'].append(float(data_line[9]))
+                    if len(data_line) == 10:
+                        self.fst_vt['MoorDyn']['Cl'].append(None)
+                        self.fst_vt['MoorDyn']['dF'].append(None)
+                        self.fst_vt['MoorDyn']['cF'].append(None)
+                    elif (len(data_line) == 11):
+                        self.fst_vt['MoorDyn']['Cl'].append(float(data_line[10]))
+                        self.fst_vt['MoorDyn']['dF'].append(None)
+                        self.fst_vt['MoorDyn']['cF'].append(None)
+                    elif (len(data_line) == 13):
```
mayankchetan, line 2935: Does MooeDyn throw an error when users only provide `Cl` & `dF` and NOT `cF` ?

RyanDavies19: Yes, it checks the line is 10, 11, or 13 entries. It doesn't check the headers though.

andrew-platt (approved):
I didn't check the physics of the VIV model or the RK4 algorithm.

There are a few `print*` statements in the new code, but since those are in `wordy` verbose sections, we'll allow them.

RyanDavies19:
Linking the theory and validation paper: https://doi.org/10.1016/j.jfluidstructs.2026.104651

Issue #2701, opened 2025-03-28
Error/warning messages during initialization is large, resulting in text being cut off from STDOUT


mayankchetan:
**Bug description**
<A clear and concise description of the bug.>
Difficult to debug errors in OpenFAST when there are a large number of warnings within a timestep/initialization. This causes users to miss the main error message since its not written to screen.

Example:
``` yaml
 **************************************************************************************************
 OpenFAST

 Copyright (C) 2025 National Renewable Energy Laboratory
 Copyright (C) 2025 Envision Energy USA LTD

 This program is licensed under Apache License Version 2.0 and comes with ABSOLUTELY NO WARRANTY.
 See the "LICENSE" file distributed with this software for details.
 **************************************************************************************************

 OpenFAST-v4.0.2-372-g058c6030
 Compile Info:
  - Compiler: GCC version 13.2.0
  - Architecture: 64 bit
  - Precision: double
  - OpenMP: No
  - Date: Mar 21 2025
  - Time: 10:26:05
 Execution Info:
  - Date: 03/27/2025
  - Time: 13:26:03-0600

 OpenFAST input file heading:
     Turbine Main Input File

  Running ElastoDyn.
 Nodal outputs section of ElastoDyn input file not found or improperly formatted.
  Running BeamDyn.
 Nodal outputs section of BeamDyn input file not found or improperly formatted.
  Running BeamDyn.
 Nodal outputs section of BeamDyn input file not found or improperly formatted.
  Running BeamDyn.
 Nodal outputs section of BeamDyn input file not found or improperly formatted.
  Running InflowWind.
  Running AeroDyn.
 AD15 Nodal Outputs: Nodal output section of AeroDyn input file not found or improperly formatted.
 Skipping nodal outputs.
    AeroDyn: projMod: 2
    Projection: Polar, BEM: polar (3D), MomentumCorrection, Sector Average
  Running ServoDyn.
  Running ServoDyn Interface for Bladed Controllers (using GNU Fortran for Linux).

 FAST_InitializeAll:AD_Init:ReadInputFiles:Blade1:ReadBladeInputs:BlCrvAng will be calculated and
 overwrite the values specified in blade file "./Turbine_AeroDyn15_blade_BDnodes.dat".
 ReadInputFiles:Blade2:ReadBladeInputs:BlCrvAng will be calculated and overwrite the values
 specified in blade file "./Turbine_AeroDyn15_blade_BDnodes.dat".
[...]

PR #2710, opened 2025-03-31
Increase length of error messages

andrew-platt:
Ready to merge (probably)

The error messages have been far too short for complex messages or messages from FAST.Farm.  Increasing it to ~2^16 characters (maybe a bit of an overkill)~ 2^13 characters (8192).

This will result in more memory required for error messages.  

In theory this should be enough (cue the "no-one will ever need more than 640k of memory..." quotes from the grey beards)...

#2701 

The few users that read error messages will see more of the error messages

Another option is to increase the `ErrMsgLen` by a more sane amount and change how we write errors to screen.

Long term we will want to go with @bjonkman's solution - for 5.0.

Output log text from test results may change slightly, but we don't check this data (timestamp always changes during testing).

### DIFF
--- a/modules/nwtc-library/src/NWTC_Base.f90
+++ b/modules/nwtc-library/src/NWTC_Base.f90
@@ -33,7 +33,7 @@ MODULE NWTC_Base
       ! General constants:
 
    INTEGER, PARAMETER            :: BITS_IN_ADDR  = C_INTPTR_T*8                  !< The number of bits in an address (32-bit or 64-bit).
-   INTEGER, PARAMETER            :: ErrMsgLen = 1024                              !< The maximum number of characters in an error message in the FAST framework
+   INTEGER, PARAMETER            :: ErrMsgLen = 65536                             !< The maximum number of characters in an error message in the FAST framework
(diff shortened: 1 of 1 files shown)


bjonkman:
@andrew-platt , I just pushed some code that I have used to capture some of the excessively long error messages in my own code base.  Not sure if you want to use this in addition to or instead of your very long error messages, but it's here: https://github.com/OpenFAST/openfast/commit/d73d81238fb9a79eaee7c3bdc23252ae680c0fc2

deslaughter (approved):

deslaughter:
I think we should go with `ErrMsgLen = 8196` which is 8 times as long as the current value. Though I also like @bjonkman's solution.

andrew-platt:
I agree with @deslaughter.  Let's put @bjonkman's solution into v5.0 (see issue #2712).  For v4.0.3 we'll do 8192

mayankchetan (approved):
Looks good to me, Thanks Andy! Let me know if you need help with the error file implementation for v6.0 :P

PR #2719, opened 2025-04-01
Python infrastructure improvements

rafmudaf:
# Python infrastructure improvements

This pull request makes some changes to the Python infrastructure in OpenFAST. It includes #2680 from @faisal-bhuiyan (merged into another branch) which brought a major refactoring of `aerodyn_inflow_library.py` and some of the Python infrastructure in r-test. Additionally, I've consolidated the existing Python interfaces (aerodyn_inflow, inflow wind, hydrodyn, moordyn, fastlib) into a pip-installable package called `OpynFAST` 😜 .

While this pull request is relatively small in scope, it sets up additional changes to the Python infrastructure in the future. Currently, there is a ton of boilerplate Python code in the OpenFAST repository. In an effort to make what exist more reusable, I wanted to start migrating it into a pip-installable package that can be used like below instead of having to manually include a variety of paths:

```python
from OpynFAST import fast

fast.init()
fast.sim()
fast.end()
```

The long term hope is that this package can evolve into a useful interface to drive OpenFAST modules. The scope here is distinct from `openfast_io` though closely related.

PR #2720, opened 2025-04-01
Adds first pass SeaState C Binding interface

rafmudaf:
# Adds first pass SeaState C Binding + NTWC C Binding module + MD C support for input file paths
This pull request adds a C Binding interface to the SeaState module. It is closely modeled on the InflowWind C Binding interface. A Python interface is also included along with a test case driven through the Python interface. Also, I've added an entry point for coupling ADI, MoorDyn, and SeaState for use in an upcoming wave tank test.

This PR also includes an update to the MoorDyn C Binding interface to support initializing from MoorDyn input files directly. That is, you can now pass the file path to `MD_C_Init` rather than requiring passing the input file strings.

Additionally, I've started putting together a NWTC C Binding module in the NWTC Library. While basic yet, it will include infrastructure used in the various C Binding modules to reduce the amount of boilerplate code.

Note that this includes #2719, so that should be merged prior to this one.


rafmudaf:
cc @hkross

andrew-platt:
Merge after #2179 is merged to `dev` and `dev` cross-merged to `dev-cbind`

modules/moordyn/src/MoorDyn_C_Binding.f90
```diff
@@ -194,8 +176,15 @@ SUBROUTINE MD_C_Init(InputFileString_C, InputFileStringLength_C, DT_C, G_C, RHO_
    ! Get fortran pointer to C_NULL_CHAR deliniated input file as a string 
    CALL C_F_pointer(InputFileString_C, InputFileString)
 
-   ! Convert string inputs to FileInfoType
-   CALL InitFileInfo(InputFileString, InitInp%PassedPrimaryInputData, ErrStat2, ErrMsg2); if (Failed()) return
+   ! Format input file contents
+   if (InputFilePassed==1_c_int) then
+      InitInp%UsePrimaryInputFile   = .FALSE.            ! Don't try to read an input -- use passed data instead (blades and AF tables not passed)
+      InitInp%FileName             = "passed_ad_file"   ! not actually used
```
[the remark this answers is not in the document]
rafmudaf: Yup - I removed that and just used an empty string. Thanks for catching that.

RyanDavies19 (approved):

glue-codes/labview/src/WaveTank.f90
```diff
@@ -0,0 +1,228 @@
...
+    IMPLICIT NONE
+    SAVE
+
+    PUBLIC :: WaveTank_Init
+
+    REAL(C_DOUBLE) :: dt_c = 0.01  ! 100 hertz
+    REAL(C_FLOAT) :: g_c = 9.8065
+    REAL(C_FLOAT) :: rho_c = 1025
+    REAL(C_FLOAT) :: depth_c = 200
+    REAL(C_FLOAT), DIMENSION(6) :: ptfminit_c = (/ 0.0, 0.0, 0.0, 0.0, 0.0, 0.0/)
```
deslaughter, line 20:
```suggestion
    REAL(C_DOUBLE) :: dt_c = 0.01_C_DOUBLE  ! 100 hertz
    REAL(C_FLOAT) :: g_c = 9.8065_C_FLOAT
    REAL(C_FLOAT) :: rho_c = 1025.0_C_FLOAT
    REAL(C_FLOAT) :: depth_c = 200.0_C_FLOAT
    REAL(C_FLOAT), DIMENSION(6) :: ptfminit_c = 0.0_C_FLOAT
```

Type specifiers needed on floating point constants; otherwise, they may have they may default to the wrong type when compiling in single or double precision.

glue-codes/labview/src/WaveTank.f90
```diff
@@ -0,0 +1,228 @@
...
+    INTEGER(C_INT) :: interporder_c = 2     ! 1: linear (uses two time steps) or 2: quadratic (uses three time steps)
+    
+    INTEGER(C_INT) :: N_CAMERA_POINTS
+    
+    INTEGER(C_INT) :: load_period = 20 ! seconds
+
+CONTAINS
+
+SUBROUTINE SetErrStat_C(ErrStatLocal, ErrMessLocal, ErrStatGlobal, ErrMessGlobal, RoutineName)
```
deslaughter, line 29: This seems like a useful function to have in `NWTC_Library` if one doesn't already exist. Can you see if there's already a version if this in OpenFAST, and if not, put it in the NWTC library?

glue-codes/labview/src/WaveTank.f90
```diff
@@ -0,0 +1,228 @@
...
+integer(c_int) :: numchannels_c
+character(kind=c_char) :: outputchannelnames_c(100000)
+character(kind=c_char) :: outputchannelunits_c(100000)
+integer(c_int) :: input_file_passed = 0                         ! We're passing paths to input files rather than input strings for all modules
+! character(kind=c_char), pointer :: filestring_c(IntfStrLen)     ! Point to input file path input argument
+
+print *, MD_InputFile_c
+print *, SS_InputFile_c
+print *, AD_InputFile_c
+print *, IfW_InputFile_c
```
deslaughter, line 79: Is this output desired or were these for debugging?

glue-codes/labview/src/WaveTank.f90
```diff
@@ -0,0 +1,228 @@
...
+!     g_c,                        &
+!     rho_c,                      &
+!     depth_c,                    &
+!     ptfminit_c,                 &
+!     interporder_c,              &
+!     numchannels_c,              &
+!     outputchannelnames_c,       &
+!     outputchannelunits_c,       &
+!     ErrStat_C, ErrMsg_C         &
+! )
```
deslaughter, line 98: Should this be removed?

glue-codes/labview/src/WaveTank.f90
rafmudaf: It was for debugging, and they've been removed in subsequent changes. See https://github.com/OpenFAST/openfast/pull/2720#discussion_r2027372503.

glue-codes/labview/src/WaveTank.f90
rafmudaf: Yea I can add it to the new `NWTC_C_Binding` module. I included it here to start because it's actually slightly different than the other C Binding utilities. Here, this is dealing with ErrStat and ErrMsg that are all C types rather than the `NWTC_C_Binding.SetErr` function that converts a local Fortran ErrStat/ErrMsg to a global C ErrStat/ErrMsg.
For the sake of keeping this appropriately scoped, I suggest to keep this subroutine here in this PR. In an upcoming change, I can build a better API in the `NWTC_C_Binding` to somehow communicate the nuances of C vs Fortran data exchanges. See https://github.com/OpenFAST/openfast/pull/2720#discussion_r2027372503.

glue-codes/labview/src/WaveTank.f90
rafmudaf: Right good catch, I'll make this change

glue-codes/labview/src/WaveTank.f90
deslaughter: It's fine to leave it in since it will be used in the future, just asking so it could be removed if it wasn't going to be used.

deslaughter (approved):

reg_tests/executePythonRegressionCase.py
```diff
@@ -37,7 +36,7 @@
 import openfastDrivers
 import pass_fail
 from errorPlotting import exportCaseSummary
-import openfast_library
+from OpynFAST import fast
```
andrew-platt, line 39: Is this `OpynFAST` or `PyOpenFAST`?

rafmudaf: This is resolved after merging `dev-cbind` which contains #2719.

PR #2726, opened 2025-04-02
generate BD files if fst_vt dictionary BD keys are not empty

ptrbortolotti:
In openfast_io, write BeamDyn files if data structure is available in the fast_vt dictionary (and not only if CompElast=2)
@mayankchetan


mayankchetan:
I messed up the re-base, Will redo the PR to `rc4.0.3`. 

More things to learn :)

Issue #2539, opened 2024-11-28
Issue when compiling OpenFAST with VS Code 2022 and IFX Fortran Compiler


ajmlu:
Dear all,

I am trying to compile OpenFAST following the documentation provided here: https://openfast.readthedocs.io/en/dev/source/install/install_vs_windows.html. I am using VSCode Community 2022 and the IFX Intel Fortran Compiler (as IFORT as recently be deprecated). However I always facing the same error that I don't manage to solve alone (x64\Release\simclist.obj: not a COFF object, bitcode, archive, import library or resource file
FASTlib : error PRJ0019: A tool returned an error code). I attached below the entire build output. 

 Do you have any insights on how I can solve this? 

Thanks for you help. 

[Error_compiling_OpenFAST.txt](https://github.com/user-attachments/files/17949645/Error_compiling_OpenFAST.txt)

ajmlu:
The dev branch of OpenFAST 3.5.4

deslaughter:
I was able to reproduce the issue on my Windows system, though I'm not sure what's causing it. As a temporary workaround, I was able to get the system to build by opening `vs-build/MAPlib/MAP_dll.sln` and building that solution first, then building `FAST` from `FAST.sln`. Please give this a try and see if it works while we try to come up with a better solution.

andrew-platt:
@ajmlu, could you try compiling again with the latest `dev` branch?

ajmlu:
Thanks a lot for your help!

@andrew-platt I just tried to compile again from the latest dev branch and I still get the same error (
[Error_compiling_OpenFAST.txt](https://github.com/user-attachments/files/18042546/Error_compiling_OpenFAST.txt))... Could that be due to an other thing than using IFX?

@deslaughter Thanks for this workaround! Where can I find this file vs-build/MAPlib/MAP_dll.sln? I could not find it anywhere on the Github repository.

Thanks again for your time and support!

deslaughter:
@ajmlu I apologize, the `MAP_dll.sln` that I had was left over from a previous version of OpenFAST. However, I was able to build the full solution with the latest changes to the `dev` branch. Please try removing the build directory and cleaning the solution before rebuilding. Sometimes Visual Studio will try to reuse files that it thinks haven't changed so it's best to start clean when troubleshooting these issues.

abhinaygoga1:
Dear FAST Community,

[...]

PR #2742, opened 2025-04-08
Fix VS build for FAST when using IFX Complier

### LINKED ISSUE #2725: Visual Studio compilation issue with fast files
I get the following error when I do a build solution to the fast file, and I only get four files out in the -build-bin folder (last image)
Pls how to solve the error in the picture below, thank you for your help

Image

Image

Image

Image

Image

Image

deslaughter:
PR is ready to merge.

This PR corrects an issue when building the FAST VS Solution in Release_x86 which failed due to missing files from the MAP library when using the Intel IFX compiler. The issue seemed to be caused by including `MAP_*` as an additional dependency instead of a build dependency, which made the paths to the object files wrong.

Resolves #2725 and #2539

Visual Studio Project and Solution Files


andrew-platt (approved):
Checking with `ifort` now.

PR #2760, opened 2025-04-17
bug fix: Adjust steady-state solver small angle assumptions

bjonkman:
This PR fixes the AeroMap solver's perturbations of orientation matrices. It was using log maps, but other parts of the code had been changed to no longer use log maps, potentially causing differences, particularly with angles farther from 0.  

This affects AeroMaps only.

This does not change existing regression tests.


andrew-platt:
Potentially backport to a 4.0.4 (if we do one)

deslaughter (approved):
Thanks for putting this together!

andrew-platt (approved):

find631:
great

PR #2776, opened 2025-04-25
Fix FAST_ExtInfw_Restart API

marchdf:
The C++ function declaration had an extra argument InflowType, that is not present in the Fortran function definition.

### DIFF
--- a/glue-codes/openfast-cpp/src/OpenFAST.cpp
+++ b/glue-codes/openfast-cpp/src/OpenFAST.cpp
@@ -642,7 +642,6 @@ void fast::OpenFAST::init() {
                         tmpRstFileRoot,
                         &AbortErrLev,
                         &turbineData[iTurb].dt,
-                        &turbineData[iTurb].inflowType,
                         &turbineData[iTurb].numBlades,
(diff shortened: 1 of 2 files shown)


andrew-platt (approved):
thanks for catching this!  We'll get this into 4.1.0

andrew-platt:
Longer description from @marchdf:

> Short story: on calling that restart function, we got invalid returns in the C++. num_blades would be ridiculous, the error code nonsensical (like ierr = 538976256)

> We wrap OF function calls with a function that check the error id. It was returning a wild error ID and so we were aborting. But it didn't have a message attached (which it would if it error "properly"). So then I started printing the error code in the fortran code and in the c++ code and they didn't match. And then I noticed that we were somehow getting 19 blades returned in the c++ whereas the fortran was saying there were 3. So I thought it was some kind of memory corruption. I took another look at the function signature in the fortran and got really confused that amr-wind had the extra inflow argument in the amr-wind function call. I couldn't understand why we would have it if it wasn't in the function definition. So I removed it and then compilation broke. More confusion. And then I compared it to the c++ function declaration and noticed that the arguments didn't match in the declaration and definition. It was a process :upside_down_face:

PR #2782, opened 2025-04-25
SubDyn upgrade to support multiple transition pieces

luwang00:
This PR modifies SubDyn to support multiple transition pieces. This is a part of the necessary upgrades to OpenFAST to enable simulations of multirotor systems (either wind turbines or MHK turbines).

For floating structures with multiple transition pieces, SubDyn now also solves and tracks the 6DoF "rigid-body" motion of the structure internally. The "rigid-body" motion can be large, including large rotation. We only require the elastic deflection of the structure relative to the "rigid-body" configuration to be small. Users need to select a joint on the structure to serve as the rigid-body reference point about which the 6DoF equations of motion are constructed.

The interface load and motion output channels is expanded to support up to 9 transition pieces. (Internally, there is no limit to the number of transition pieces. Only the output is limited to 9.) These channels are `Intf?FXss`, `Intf?FYss`, etc., where `?` is a number between 1 to 9. If no number is provided, the values associated with the first transition piece is returned for backward compatibility.  

The modified SubDyn is now compatible with tight coupling of OpenFAST v5. The necessary changes were made with the help of @deslaughter.

More information to come.

**Additional changes**
* Previously, SubDyn driver only uses the input at one time step. As a result, time integration methods involving corrector steps do not function correctly. This is now corrected. (This issue does not affect SubDyn when coupled to OpenFAST.)
* Also enabled time step interpolation on user-provided transition-piece motion time series in SubDyn driver.

SubDyn, openfast_io, glue-code

Existing SubDyn r-test results changed slightly due to corrections to SubDyn driver and other modifications. SubDyn driver and primary input files are also updated with new inputs needed for multiple transition pieces.


andrew-platt:
Could you add some test results here?

modules/subdyn/src/FEM.f90
```diff
@@ -1586,7 +1586,7 @@ SUBROUTINE PseudoInverse(A, Ainv, ErrStat, ErrMsg)
    end do
    ! Compute Ainv = 1.0*V^t * U^t + 0.0*Ainv     V*(inv(S))*U' 
    !call DGEMM( 'T', 'T', N, M, K, 1.0, V, K, U, M, 0.0, Ainv, N)
-   print*,'8'
+   !print*,'8'
```
andrew-platt, line 1589: We might as well remove this line entirely. Not sure why it is here.

modules/subdyn/src/SD_FEM.f90
```diff
@@ -91,7 +91,7 @@ MODULE SD_FEM
   ! Implementation Flags
   LOGICAL, PARAMETER :: DEV_VERSION    = .false.
   LOGICAL, PARAMETER :: BC_Before_CB   = .true.
-  LOGICAL, PARAMETER :: ANALYTICAL_LIN = .true.
+  LOGICAL, PARAMETER :: ANALYTICAL_LIN = .false.   ! Analytical linearization is no longer valid in general
```
andrew-platt, line 94: If analytical linearization isn’t valid anymore, should it be removed (future PR perhaps).

luwang00: Yes, we should remove that, but the subroutine for analytical Jacobian is still used for writing the summary file, so we can't get rid of it entirely.

modules/subdyn/src/FEM.f90
luwang00: True. I'll get rid of it.

modules/subdyn/src/SubDyn.f90
```diff
@@ -2095,60 +2183,64 @@ SUBROUTINE SD_AM2( t, n, u, utimes, p, x, xd, z, OtherState, m, ErrStat, ErrMsg
    CHARACTER(*),                   INTENT(  OUT)   :: ErrMsg         !< Error message if ErrStat /= ErrID_None
    ! local variables
    TYPE(SD_InputType)                              :: u_interp       ! interpolated value of inputs 
-   REAL(ReKi)                                      :: xq(2*p%nDOFM) !temporary states (qm and qmdot only)
-   REAL(ReKi)                                      :: udotdot_TP2(6) ! temporary copy of udotdot_TP
+   TYPE(SD_ContinuousStateType)                    :: dxdt
+   TYPE(SD_OutputType)                             :: y
+   REAL(R8Ki)                                      :: xq(m%Jac%Nx,1)
+   REAL(R8Ki)                                      :: tmpAry(m%Jac%Nx)
+   REAL(R8Ki),allocatable                          :: AM2Jac(:,:)
```
deslaughter, line 2190: Could `AM2Jac` be moved to the registry so it's not being allocated and deallocated each time this is called?

modules/subdyn/src/SubDyn.f90
```diff
@@ -2095,60 +2183,64 @@ SUBROUTINE SD_AM2( t, n, u, utimes, p, x, xd, z, OtherState, m, ErrStat, ErrMsg
    CHARACTER(*),                   INTENT(  OUT)   :: ErrMsg         !< Error message if ErrStat /= ErrID_None
    ! local variables
    TYPE(SD_InputType)                              :: u_interp       ! interpolated value of inputs 
-   REAL(ReKi)                                      :: xq(2*p%nDOFM) !temporary states (qm and qmdot only)
-   REAL(ReKi)                                      :: udotdot_TP2(6) ! temporary copy of udotdot_TP
+   TYPE(SD_ContinuousStateType)                    :: dxdt
+   TYPE(SD_OutputType)                             :: y
+   REAL(R8Ki)                                      :: xq(m%Jac%Nx,1)
```
deslaughter, line 2188: Could this be moved to the registry so it's not being allocated and decallocated each call. Also, this type of variable declaration implicitly causes an allocation since the dimensions are only known at runtime, so it it should be avoided.

modules/subdyn/src/SubDyn.f90
```diff
@@ -2095,60 +2183,64 @@ SUBROUTINE SD_AM2( t, n, u, utimes, p, x, xd, z, OtherState, m, ErrStat, ErrMsg
...
-   x%qm    = x%qm    - xq(        1:  p%nDOFM)
-   x%qmdot = x%qmdot - xq(p%nDOFM+1:2*p%nDOFM)
+   ! J*( x_n - x_n+1 ) = dt * ( x_dot(u_n,x_n) + x_dot(u_n+1,x_n) )/2
+   ! J*( x_n - x_n+1 ) = dt * ( A*x_n +  B *(u_n + u_n+1)/2 + Fx) <- if linear but no longer the case
+   !....................................................
+   CALL LAPACK_gels( 'N', AM2Jac, xq, ErrStat2, ErrMsg2); CALL SetErrStat(ErrStat2,ErrMsg2,ErrStat,ErrMsg,'SD_AM2')
+   CALL SD_VarsUnpackContState(p%Vars, xq(:,1), dxdt)
+
+   ! Update states
+   if (p%TP1IsRBRefPt) then
```
deslaughter, line 2230: Could this update be done with the state data array `xq` before unpacking back into `x`?

luwang00: I think doing unpacking might be easier to maintain. This way, we don't have to assume how the states are ordered.

modules/subdyn/src/SubDyn.f90
luwang00: Moved `AM2Jac`, `xq`, and `tmpAry` to `MiscVarType`. These are now only allocated once at initialization. Also removed `dxdt` and reuse `m%dxdt_lin` instead.

modules/subdyn/src/SubDyn.f90
luwang00: See above. Resolved.

modules/subdyn/src/SubDyn.f90
deslaughter: Sounds good.

PR #2791, opened 2025-04-30
MD cbind: correction to description of mesh point mapping

andrew-platt:
Merge after #2735

The documentation in `MoorDyn_c-binding.f90` about the input/output mesh was incorrect. Only a single mesh point can be passed across the interface, but it is able to rigidly map over to multiple _MoorDyn_ mesh points

Question raised in #2735 review: https://github.com/OpenFAST/openfast/pull/2735#discussion_r2049554766

### DIFF
--- a/modules/moordyn/src/MoorDyn_C_Binding.f90
+++ b/modules/moordyn/src/MoorDyn_C_Binding.f90
@@ -106,10 +106,12 @@ INTEGER(IntKi),   PARAMETER             :: INPUT_PRED = 1      !< Index for pred
 !------------------------------
 !  Meshes for external nodes
 !     These point meshes are merely used to simplify the mapping of motions/loads
-!     to/from MD using the library mesh mapping routines.  These meshes may contain
-!     one or multiple points.
-!        - 1 point   -- rigid floating body assumption
-!        - N points  -- flexible structure (either floating or fixed bottom)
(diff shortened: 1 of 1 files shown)


modules/moordyn/src/MoorDyn_C_Binding.f90
```diff
@@ -106,10 +106,12 @@ MODULE MoorDyn_C
 !------------------------------
 !  Meshes for external nodes
 !     These point meshes are merely used to simplify the mapping of motions/loads
-!     to/from MD using the library mesh mapping routines.  These meshes may contain
-!     one or multiple points.
-!        - 1 point   -- rigid floating body assumption
-!        - N points  -- flexible structure (either floating or fixed bottom)
+!     to/from MD using the library mesh mapping routines.  The input mesh into the
+!     library can only contain one point at present and is rigidly mapped to multiple
+!     MoorDyn mesh points.
```
RyanDavies19, line 111: Add / revise `This means all MoorDyn coupled objects as defined in the input file will be rigidly attached to the input mesh`

andrew-platt: Good clarification.  Will do.

andrew-platt: resolved with db87f7d81

RyanDavies19 (approved):

Issue #2748, opened 2025-04-10
leftover MD-SS coupling tasks from #2705


andrew-platt:
From PR #2705 

**TODO before release to main OpenFAST dev branch**
These will be addressed in future PR's to the dev-cbind branch
- [x] Add in coupling for currents
  - [x] SeaState
- [x] Add wave directional spreading into MoorDyn for WaterKin = 2 (SeaState inputs to the MoorDyn grid)
- [x] C-bind pointer passing (@rafmudaf is leading this)
- [x] Test implementation for accuracy
- [x] Add regression test for water kinematics in MoorDyn, include example input files
- [x] Docs update with new input file formats

RyanDavies19:
@andrew-platt everything here is done except for the wave directional spreading

PR #2705, opened 2025-03-28
MoorDyn-SeaState Coupling

RyanDavies19:
**Feature or improvement description**
This is the beginning of coupling MoorDyn to SeaState. So far it only supports waves, no currents. MoorDyn now has three options for water kinematics:
1) The existing method, with user provided wave and current information and grid information (Old method, WaterKin = 1)
2) A user defined grid in MoorDyn that the SeaState wave and current input data is interpolated to (Hybrid method, WaterKin = 2)
3) SeaState grid and wave and current data passed directly to MoorDyn, accessed every time step (SeaState method, WaterKin = 3)

Options 1 and 2 are specified in the `WaterKin.dat` file that MoorDyn currently requires for water kinematics (there are modifications for specifying the current grid if using option 2. They follow the same approach as defining the Z grid point data for waves). The input variables `WaveKinMod` and `CurrentMod` trigger which method is used; they must be equal or one must be zero. Option 3 is triggered by the keyword `SeaState` (case insensitive) in the `WaterKin` flag in the main MoorDyn input file. 

~~This is currently only set up to support waves, currents are on the todo list.~~

Much of the approach follows how the SeaState `WaveField` pointer is passed into Hydrodyn. The MoorDyn driver calls SeaState Init and then passes MoorDyn the pointer. Otherwise, `FAST_subs` and `FAST_Farm_subs` will pass it in. In the case of FAST Farm, option 3 (the full SeaState coupling) is not allowed because there is not a farm level SeaState instance. Instead (if using option 2), `FAST_Farm_subs` will pass MoorDyn the `WaveField` pointer from the first SeaState instance, because it contains the same seed information as all the other instances. This will change in future development once the farm level SeaState is implemented. 

**Impacted areas of the software**
MoorDyn
OpenFAST
FAST.Farm
C-Bindings

**Test results, if applicable**
Tested for no seg faults when running the three SeaState - MoorDyn coupling options in OpenFAST and FF. 
Tested for numerical accuracy by matching frequency of input waves with MD outputs, and matching current speeds w/ MD output current speeds. 

**TODO before release to main OpenFAST dev branch**
These will be addressed in future PR's to the dev-cbind branch
- [x] typo in FAST_Farms_subs
- [x] revert changes to Python wrapper of C-bind
- [x] Add in coupling for currents
  - [x] SeaState
- [ ] ~~Add wave directional spreading into MoorDyn for WaterKin = 2 (SeaState inputs to the MoorDyn grid)~~
[...]

PR #2794, opened 2025-05-02
MoorDyn wave directional spreading and custom current profile if no SeaState current 

RyanDavies19:
Enables wave spreading in MoorDyn in conjunction with the recently merged SeaState coupling. Wave directional spreading was previously only available when running MoorDy fully coupled with SeaState (waterkin = 3). Now wave directional spreading is enabled with the hybrid coupling (waterkin = 2), where input parameters are controlled by the SeaState input file but the grid size and discretization is controlled by MoorDyn. There's no input file changes, this just enables more SeaState features to be compatible with MoorDyn in the hybrid coupling.

This PR also enables a user defined current profile (CurrentMod 1) while using the hybrid coupling method if the SeaState currents are disabled. This allows users to calculate wave kinematics using SeaState but still provide their own current profile (which typically is based on buoy data). 

#2748 

MoorDyn

Final task from #2705 

To test wave directional spreading I ran MoorDyn fully coupled with SeaState with wave spreading enabled. I also ran MoorDyn with the hybrid SeaState coupling with SeaState wave spreading using an identical grid size as the SeaState grid in MoorDyn. For both of these, I checked submergence of a fixed surface piercing rod located at 5,5. Submergence matched exactly, which means the wave elevations pulled from the wave-field pointer match the wave elevations calculated by MoorDyn using the wave-field pointer frequency and direction information. Figure below shows the submergence results:

Screenshot 2025-05-02 at 2 45 08 PM

### DIFF
--- a/modules/moordyn/src/MoorDyn_Misc.f90
+++ b/modules/moordyn/src/MoorDyn_Misc.f90
@@ -1287,10 +1287,10 @@ CONTAINS
                END IF
                IF (i == 100) THEN
                   p%nzCurrent = 100
-                  CALL WrScr("WARNING: MD can handle a maximum of 100 current profile points")
                   IF (p%writeLog > 0) THEN
                      WRITE(p%UnLog, '(A)'        ) "    WARNING: MD can handle a maximum of 100 current profile points"
                   ENDIF      
+                  CALL SetErrStat(ErrID_Warn, "MD can handle a maximum of 100 current profile points", ErrStat, ErrMsg, RoutineName)
                   EXIT
                END IF
             END DO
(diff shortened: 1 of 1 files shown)


andrew-platt (approved):
Looks good!

I'm surprised that the `WaveDirArr` is stored in degrees instead of radians.  SeaState also has it stored in degrees.

RyanDavies19:
> Looks good!
> 
> I'm surprised that the `WaveDirArr` is stored in degrees instead of radians. SeaState also has it stored in degrees.

I noticed, figured it was set up that way on purpose. Lines 897 and 898 in Wave.f90 has the same as what I added for calculating CosWaveDir and SinWaveDir

PR #2798, opened 2025-05-06
Docs: correct Bladed Interface channel 63 description

andrew-platt:
Ready to merge

The description for the `avrSWAP` array record 63 was incorrect. This was reflected both in the documentation on readthedocs.org and in the summary file generated by _ServoDyn_.  The internal code was correct.

Original:
* `63      <--  Number logging channels`
Corrected:
* `63      -->  Record number for start of logging output (-) [set to ###]`

Reported internally

Documentation only.

Consider for backporting to 4.0.5


jjonkman (commented):
I agree with the wording change, but the arrow direction should still be "<--" as this data is passed to OpenFAST from the DISCON controller.

andrew-platt:
The code and the Bladed document both indicate record 63 is passed from SrvD to the DLL.  Am I misreading this?

jjonkman (approved):
My bad; I misread the code!  You are right.

PR #2808, opened 2025-05-13
[BugFix] TurbSim calculating grid bottom location

ptrbortolotti:
**Issue**
This PR tries to fix a long-standing issue with TurbSim, which at times errors out saying that the lowest grid point is underground, although that's not the case. 

To generate the error, try setting:

```
NumGrid_Z: 24
NumGrid_Y: 24
HubHt: 110
GridHeight: 218
GridWidth: 156
```

The lowest grid point should be at `110 - 218/2 = 1m` above ground

However the current TurbSim estimates the rotor diameter as the minimum between height and width, so `156 m`. Then, TurbSim estimates the top of the grid as `110 + 156/2 = 188`, and then errors out complaining that the bottom of the grid is underground at `188 - 218 =-30 m`

The PR fixes this issue by simply setting Zbottom as:
`p_grid%Zbottom = p_grid%Ztop - 0.5*p_grid%GridHeight`

TurbSim

[ws6.00_s0_TI0.06_shear0.20.txt](https://github.com/user-attachments/files/20196219/ws6.00_s0_TI0.06_shear0.20.txt)

### DIFF
--- a/modules/turbsim/src/TSsubs.f90
+++ b/modules/turbsim/src/TSsubs.f90
@@ -1347,8 +1347,7 @@ SUBROUTINE CreateGrid( p_grid, p_usr, UHub, AddTower, ErrStat, ErrMsg )
    p_grid%GridRes_Y = p_grid%GridWidth  / REAL( p_grid%NumGrid_Y - 1, ReKi )
    p_grid%GridRes_Z = p_grid%GridHeight / REAL( p_grid%NumGrid_Z - 1, ReKi )      
 
-   p_grid%Ztop = p_grid%HubHt + 0.5*p_grid%GridHeight ! height of the highest grid points
-   p_grid%Zbottom = p_grid%Ztop - p_grid%GridRes_Z * REAL(p_grid%NumGrid_Z - 1, ReKi)   ! height of the lowest grid points
+   p_grid%Zbottom = p_grid%HubHt - 0.5*p_grid%GridHeight ! height of the lowest grid points
 
    IF ( p_grid%Zbottom <= 0.0_ReKi ) THEN
       CALL SetErrStat(ErrID_Fatal,'The lowest grid point ('//TRIM(Num2LStr(p_grid%Zbottom))// ' m) must be above the ground. '//&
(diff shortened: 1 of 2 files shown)


bjonkman:
For backward compatibility and for other use cases, I'd recommend that you just make sure that Zbottom is always greater than zero instead of recalculating as the HubHt - half the grid height. I made a PR with my proposed changes: https://github.com/ptrbortolotti/openfast/pull/5

rthedin:
Bonnie- Unless I'm missing something, I don't see a reason to remove the error checking.

With your changes, the code will no longer crash (at least not at the same spot) with the following problematic input:
```
HubHt: 110
GridHeight: 230
GridWidth: 156
```

bjonkman:
> Bonnie- Unless I'm missing something, I don't see a reason to remove the error checking.
> 
> With your changes, the code will no longer crash (at least not at the same spot) with the following problematic input:
> 
> ```
> HubHt: 110
> GridHeight: 230
> GridWidth: 156
> ```

I set the grid bottom so that it can never be less than `Tolerance = 0.0001 meters`, so the `if` statement to check for errors there could never be true. That's why I removed it.

rthedin:
I understand that with the inputs I gave, the overall grid would still have the correct width and height, but should we change the behavior of handling problematic inputs? I think if the user requests `HubHt` 110 m and `GridHeight` 230 m, maybe it should still crash?

bjonkman:
It really doesn't matter to me if you want the user to change the inputs (i.e., cause a fatal error, or what you call "crash"), or if we just simulate more wind up to the height requested. I think it's easier for backward compatibility if we just simulate the extra wind. 

I don't want to just center the grid vertically because that makes it harder for someone with a single land-based turbine who wants to generate turbulence over the rotor and the full tower. They'd have to generate quite a few more points. Also, they wouldn't be able to re-generate existing wind files with the new code.

If you wanted to make a bigger change, we could modify the TurbSim input file so the user explicitly specifies the y-z bounds of the turbulence grid; the code would then calculate `GridHeight` and `GridWidth` from the user-specified boundaries.

andrew-platt:
I agree with Bonnie on calculating the extra height - we don't want to limit the grid height to only double the hub height.

There is a subtlety here with the hub height and grid height that should be mentioned (I spoke with @rthedin in person about this).  If the grid height requested is more than double the hub height, the grid is shifted from centering on the hub height to the bottom being just above ground with the hub height below the center of the grid.  This is a desired behavior as there are times we want the grid significantly taller than twice hub-height.

PR #2825, opened 2025-05-21
Simplify GitHub Action for regression tests (backport from dev-tc)

deslaughter:

This PR introduces a slightly more streamlined process for the regression tests run in Github Actions. It reduces the amount of caching used in the build process by compiling most of the executables in one step and then using them to run the tests so only one workspace is cached. The same regression tests are still being run, but some have been grouped into new jobs.

This change also separates out the Matlab Simulink build and test which is currently failing due to an outage in Matlab's servers. That issue is being tracked at matlab-actions/setup-matlab#146.

`automated-dev-tests.yml`

This PR also revealed an issue with the `5MW_OC4Jckt_DLL_WTurb_WavesIrr_MGrowth` regression test which has to be run with `OMP_NUM_THREADS` not specified to match the test baseline. This will be corrected in the future by setting `OMP_NUM_THREADS=1` and regenerating the baseline.

No tests were changed and the Simulink test cases will start working once the Matlab outage is resolved.

### DIFF
--- a/.github/workflows/automated-dev-tests.yml
+++ b/.github/workflows/automated-dev-tests.yml
@@ -21,15 +21,105 @@ env:
   C_COMPILER: gcc-12
   GCOV_EXE: gcov-12
   CMAKE_BUILD_PARALLEL_LEVEL: 8
-  CTEST_PARALLEL_LEVEL: 2
-
+  CTEST_PARALLEL_LEVEL: 4
 
 jobs:
 
   ### BUILD JOBS
 
+  build-all-debug-single:
+    # Tests compiling in debug mode with single precision.
+    # This workspace is not used by any other subtests, it checks type errors of the type ReKi/R8Ki
+    # Debug speeds up the build.
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Install dependencies
+        run: |
+          sudo apt-get update -y
+          sudo apt-get install -y libopenblas-dev libopenblas-openmp-dev libhdf5-dev libnetcdf-dev libopenmpi-dev libyaml-cpp-dev
+      - name: Setup workspace
+        run: cmake -E make_directory ${{runner.workspace}}/openfast/build
+      - name: Configure build
+        working-directory: ${{runner.workspace}}/openfast/build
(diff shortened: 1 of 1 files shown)


andrew-platt:
With 4 threads on OpenMP, the `5MW_OC4Jckt_DLL_WTurb_WavesIrr_MGrowth` test is currently failing.  So we might need to change the baseline to use only one thread now.

.github/workflows/automated-dev-tests.yml
```diff
@@ -602,145 +398,20 @@ jobs:
...
-          path: |
-            ${{runner.workspace}}/openfast/build/reg_tests/openfast_io
 
-
-  rtest-OF-simulink:
-    runs-on: ubuntu-22.04
-    needs: build-openfast-release
+  rtest-OF-offshore:
+    runs-on: ubuntu-24.04
+    needs: build-all-release
```
andrew-platt, line 404:
```suggestion
    needs: build-all-release
    OMP_NUM_THREADS: 2
```
Setting to 2 threads here to match old method.  Might get the `5MW_OC4Jckt_DLL_WTurb_WavesIrr_MGrowth` to match the baseline (we don't have any artifacts from this case passing on `main` or `.log` files in the repo to check against).

andrew-platt:
Regarding the `5MW_OC4Jckt_DLL_WTurb_WavesIrr_MGrowth` case:  the current suspicion is that something in the marine growth bit of code is causing different results in the simulation depending on the number of threads used (this is the only case using marine growth and showing this behavior).  The marine growth section was rewritten in `dev` (for v4.1.0), so we don't particularly want to address it here.

deslaughter:
There is something in the baseline for `5MW_OC4Jckt_DLL_WTurb_WavesIrr_MGrowth` that is specific to using `ubuntu-22.04`, `gfortran-12`, and leaving `OMP_NUM_THREADS` unset. For OpenFAST v4.1 we should update to `ubuntu-24.04`, `gfortran-14`, and set `OMP_NUM_THREADS=1` and redo the baseline for this case so it corresponds to a given environment.

Other than the issue with the Simulink test, which is caused by Matlab's mpm servers and beyond our control, the PR is ready to be merged.

Issue #2827, opened 2025-05-22
Catch or fix invalid VTK_fps


rafmudaf:
# Add functionality to either report or fix an invalid VTK_fps setting
As mentioned in #2826, I left the `VTK_fps` setting at 0 and got a runtime error since the Fortran string format setting included a negative number. It would be ideal to check for a value < 1 for `VTK_fps` and suggest to the user to change it. Alternatively, if a good guess at a reasonable frame rate is available, it might be good to set that when `VTK_fps` is 0.

andrew-platt:
Interestingly there is code in `FAST_Subs.f90` to handle the `VTK_fps = 0` case. However, I can reproduce exactly what you are seeing.

andrew-platt:
Fixed with #2831

PR #2831, opened 2025-05-22
[BugFix] WrVTK with VTK_fps fails

andrew-platt:
Ready to merge

If `WrVTK` is non-zero, setting `VTK_fps=0` causes a failure as follows:

```

  Time: 0 of 6 seconds.
At line 321 of file /Users/aplatt/software-development/GitHub/openfast-5/modules/nwtc-library/src/ModMesh.f90
Fortran runtime error: Nonnegative width required in format
(i-5.-5)
   ^

Error termination. Backtrace:
#0  0x103f22d2f
#1  0x103f238d7
#2  0x103f241a7
#3  0x10400a48b
#4  0x104019b0f
#5  0x102eac47b
#6  0x1022d301f
#7  0x1022d7acf
#8  0x1022ec2cf
#9  0x1022ec5f7
#10  0x1021c9e53
#11  0x102f6671f
```

The intent of `VTK_fps=0` was to output at `T=0` and at the last step of the simulation.  However, due to the way it was setup, this wasn't working due to a negative value for the field width for the time step in the VTK filenames.  Internally the math for calculating the width was as follows:

```
      if ( EqualRealNos(p%VTK_fps, 0.0_DbKi) ) then
         TmpTime = p%TMax + p%DT
...
      ... IF (TmpTime > p%TMax) THEN
         p%n_VTKTime = HUGE(p%n_VTKTime)
...
 p_FAST%VTK_tWidth = CEILING( log10( real(p_FAST%n_TMax_m1+1, ReKi) / p_FAST%n_VTKTime ) ) + 1
```
which evaluated to `CEILING( log10(         2000  /   2147483647 )) + 1  = -5 ` causing the format failure.

This has been corrected so that now `p%n_VTKTime = NINT( p%TMax  / p%DT )` giving `VTK_tWidth` will evaluate to `1` and triggering VTK writes at the last time step.

#2827 

VTK outputs only when `VTK_fps=0`

No results change

### DIFF
--- a/modules/openfast-library/src/FAST_Subs.f90
+++ b/modules/openfast-library/src/FAST_Subs.f90
@@ -3717,7 +3717,7 @@ END DO
       IF (p%WrVTK == VTK_ModeShapes) THEN
          p%n_VTKTime = 1
       ELSE IF (TmpTime > p%TMax) THEN
-         p%n_VTKTime = HUGE(p%n_VTKTime)
+         p%n_VTKTime = NINT( p%TMax  / p%DT )   ! write at init and last step only
       ELSE
          p%n_VTKTime = NINT( TmpTime / p%DT )
          ! I'll warn if p%n_VTKTime*p%DT is not TmpTime


rafmudaf:
I'm curious what the context was for the existing form. With the `HUGE` denominator, the argument to log10 will practically always be less than 1, so the result of that will practically always be negative. That seems intentional - what do you think @andrew-platt?

I agree that the NINT form is the fix for the problem I mentioned, but I'm wondering if this has a side effect if something else was expecting the other form for `p_FAST%n_VTKTime`.

PR #2883, opened 2025-06-24
Gh actions debug

deslaughter:

This PR fixes the issue with Python not being able to find the `requirements.txt` file in the regression test GitHub action. The issue was caused by not specifying or using the wrong variable for the working directory when running `pip install -r requirements.txt`. Different runners would interpret the working directory differently causing tests to sometimes pass and sometimes fail. This PR uses `${{github.workspace}}` instead of `${{runner.workspace}}` to specify the working directory.

In addition, the existing regression test GitHub action uses caches to transfer data and binaries between the jobs in a run. Based on https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/storing-and-sharing-data-from-a-workflow#comparing-artifacts-and-dependency-caching, artifacts should be used for sharing data within a run (across jobs) and caches should be used for sharing dependencies across workflow runs. This PR changes the action to use artifacts, which reduces storage size and makes it possible for the user to download the compilation output after the run, which may be useful.

`automated-dev-tests.yml`

### DIFF
--- a/.github/workflows/automated-dev-tests.yml
+++ b/.github/workflows/automated-dev-tests.yml
@@ -96,7 +96,7 @@ jobs:
             -DPython_ROOT_DIR:PATH=${{env.pythonLocation}} \
             -DBUILD_SHARED_LIBS:BOOL=OFF \
             -DBLA_VENDOR:STRING=OpenBLAS \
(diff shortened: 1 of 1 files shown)


andrew-platt (commented):
Looks good and tests ok.

PR #2891, opened 2025-06-29
Minor fixes for v4.1

gbarter:

Two small bug fixes:

- gfortran on windows in the conda environment is finicky that IMPLICIT NONE come before any ATTRIBUTE lines.  This is currently fixed in the conda build with a local patch but should be part of the source code.
- openfast_io functions when `FAST_directory` is not set results in errors since the default value is `None` (instead of an empty string).  I kept the `None` default value, but added a check for it.

Not sure if this PR should be directed to the `dev` branch, or the `rc4.1.1` branch.

### DIFF
--- a/glue-codes/labview/src/WaveTank.f90
+++ b/glue-codes/labview/src/WaveTank.f90
@@ -51,12 +51,13 @@ SUBROUTINE WaveTank_Init(   &
     ErrStat_c,              &
     ErrMsg_c                &
 ) BIND (C, NAME='WaveTank_Init')
+
+IMPLICIT NONE
+ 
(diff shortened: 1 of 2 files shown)


deslaughter:
@gbarter That's a really strange bug with the `implicit none` in MinGW. Having `implicit none` at the top of the module, is sufficient, so could you remove it from the functions? I think that will be a cleaner solution that putting it before the attribute lines.

gbarter:
Thanks, Derek.  That seems to have worked.  Relevant PR on the conda feedstock is [here](https://github.com/conda-forge/openfast-feedstock/pull/67).

andrew-platt (approved):

Issue #1081, opened 2022-04-06
Bug report - SubDyn bug when modeling rigid platforms


RBergua:
**Bug description**
When modeling a rigid platform in SubDyn (e.g., using rigid links and lumped masses), the buoyancy is not applied over the system and the structure sinks.
First, it's important to note that SubDyn somehow requests NPropSets > 0. So, at least one _dummy_ section that is not going to be used must be defined. 

**To Reproduce**
Any SubDyn model with only rigid links and lumped masses on it. For example, one SubDyn model with 5 joints conncted with rigid links between them: 1 joint for the interface, 1 joint for the lumped mass and 3 equally-spaced joints at the tip of three arms that would connect with the mooring lines. CBMod can be defined as False or True with Nmodes = 0. The system doesn't have internal DOFs, only the 6 DOFs corresponding to the rigid body motion.

Run one model with the above characteristics in OpenFAST v3.1.0 (compile already available from GitHub OpenFAST repo).

**Expected behavior**
The system to float. 
Workaround: Introduce at least one beam in the model. Then the system will behave as expected.

**Screenshots, if applicable**
![image](https://user-images.githubusercontent.com/33709140/162001422-464b723d-8e89-4d19-b15c-f8ee645f9107.png)

**OpenFAST Version**
![image](https://user-images.githubusercontent.com/33709140/162001519-b895f2a9-2253-4720-b6d4-64ca09076b2b.png)

jjonkman:
Presumably this issue is related to issue #854, whereby the "loads and displacements are wrong when rigid links are connected [to] the interface [joint]".  Until this issue is fixed, the workaround is to ensure that a beam element is connected to the interface joint.  (You can still set `Nmodes` = 0 if you want to model a rigid substructure in SubDyn.)

RBergua:
The issue #854 appears **sometimes** when there is a rigid link connected to the interface joint AND there are cables involved. The model that I'm working with doesn't have any cable. It only has rigid links and lumped masses.

Moreover, I have been using several models with rigid links at the interface joint and I have not encountered any problems. The issue here is that you must have, at least, one beam somewhere in your model. And this beam can be anywhere, it is not necessary to be connected to the interface joint.

jjonkman:
OK, thanks for clarifying.  I expect @ebranlard can comment on what he expects to be the cause of the issue.

andrew-platt:
@luwang00 and @RBergua, is this fixed in 4.0.0?

[...]

Issue #854, opened 2021-10-05
SubDyn issue when rigid links are connected to interface and cable


ebranlard:
**Bug description**
Loads and displacements are wrong when rigid links are connected between the interface node and nodes that have cables. 

**To Reproduce**
Use this test case. 
https://github.com/OpenFAST/r-test/main/f/SD-Loads/modules/subdyn/SD_Rigid2Interf_Cables

**Expected behavior**
IntfZss is 4.6e6N where a value of 2.23e7N would be expected.

**OpenFAST Version**
Dev version as of October 2021. Bug present ever since the rigid link feature was introduced (around OF 2.6).

rafmudaf:
The test case references above that exposes the bug is now in the `main` branch at https://github.com/OpenFAST/r-test/tree/main/modules/subdyn/SD_Rigid2Interf_Cables

RBergua:
I do believe the problem is related to defining one beam at the interface joint and, at least, one rigid link to that same interface joint. For example: If I define a rigid link that is used to connect a lumped mass to the interface joint, such a mass is somehow not taken into account. Despite SubDyn reporting the expected mass and inertia in the summary file.

luwang00:
Issue resolved with PR #2905.

PR #2905, opened 2025-07-08
SD: Fix a bug where interface DoF loads are incomplete if the interface joint is part of a rigid assembly

luwang00:

This PR fixes a bug where the external loads on SubDyn interface degrees of freedom are incomplete if the interface joint is part of a rigid assembly. Previously, only the loads on the interface joint itself are computed. This is incomplete. We instead need to include the external loads on all nodes belonging to the same rigid assembly due to DoF reduction.

Addresses both Issue #854 and Issue #1081 (verified by @RBergua below). Both issues can now be closed. The corresponding warning message in SubDyn is also deleted.

**Additional information**
This appears to be a very old bug present since rigid links were first added to SubDyn. We can consider backporting to v4.x.

SubDyn

No change to existing test results.

### DIFF
--- a/modules/subdyn/src/SD_FEM.f90
+++ b/modules/subdyn/src/SD_FEM.f90
@@ -386,11 +386,6 @@ SUBROUTINE SD_ReIndex_CreateNodesAndElems(Init,p, ErrStat, ErrMsg)
                CALL Fatal('All joints of a rigid link should be cantilever (not ball/pin/universal). The problematic member is MemberID='//TRIM(Num2LStr(mID))//' (which is a rigid link) involving joint JointID='// TRIM(Num2LStr(JointID))// ' (which is not a cantilever joint).')
                return
             endif
-            ! Check that rigid links are not connected to the interface
-            iInterf = FINDLOCI(p%Nodes_I(:,1), iJoint )
-            if (iInterf>=1) then
-               CALL WrScr('[WARNING] There might be a bug when one beam and one rigid link are connected to the interface nodes. The problematic member might be MemberID='//TRIM(Num2LStr(mID))//' (which is a rigid link) involving joint JointID='// TRIM(Num2LStr(JointID))// ' (which is in an interface joint).')
-            endif
          endif
       enddo
       ! Column 4-5: PropIndex 1-2 (instead of PropSetID1&2)
--- a/modules/subdyn/src/SubDyn.f90
+++ b/modules/subdyn/src/SubDyn.f90
@@ -670,7 +670,7 @@ SUBROUTINE SD_SolveEOM( t, u, p, x, xd, z, OtherState, m, ErrStat, ErrMsg )
 
          if (p%TP1IsRBRefPt) then ! More than one transition piece
 
-            call GetExtForceOnInterfaceDOF(p, m%Fext, F_I)
+            call GetExtForceOnInterfaceDOF(p, m, F_I)
 
             F_TP1 =   matmul(            F_I, p%TI(:,1:6) ) &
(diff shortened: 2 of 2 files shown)


andrew-platt:
Backport to 4.2

Edit: backporting this is not as trivial as initially thought.  We will leave this in 5.0

jjonkman (approved):

RBergua:
I performed an easy test to check if the behavior is the expected one. Similar to https://github.com/OpenFAST/openfast/issues/1081, I wanted to model a rigid platform in SubDyn simply as a rigid link with a lumped mass and inertia. It is known, that this test in the past resulted in issues. 

In this case, I took the model from r-test (https://github.com/OpenFAST/r-test/tree/main/glue-codes/openfast/5MW_OC4Semi_WSt_WavesWN). I simplified the system by disabling InflowWind and AeroDyn (i.e., no aero), disabling ServoDyn (i.e., no controller), and defined still water conditions in SeaState. I also defined all initial conditions as 0 in ElastoDyn and setup the `YawDOF` as False. 

The original model from r-test has the platform defined in ElastoDyn. I defined an equivalent system in SubDyn by defining two joints and the corresponding mass and inertia. Basically:
```
---- STRUCTURE JOINTS: joints connect structure members (~Hydrodyn Input File) --------
             2   NJoints     - Number of joints (-)
JointID          JointXss               JointYss               JointZss     JointType JointDirX  JointDirY JointDirZ JointStiff    ![Coordinates of Member joints in SS-Coordinate System][JointType={1:cantilever, 2:universal joint, 3:revolute joint, 4:spherical joint}]
  (-)               (m)                    (m)                    (m)         (-)        (-)        (-)       (-)     (Nm/rad) 
   1              0.00000                0.00000                0.00000        1         0.0        0.0       0.0       0.0    
   2              0.00000                0.00000               -8.65880        1         0.0        0.0       0.0       0.0    
```
```
----------------------- RIGID LINK PROPERTIES -----------------------------------------
             1   NRigidPropSets - Number of rigid link properties
PropSetID   MatDens   
  (-)       (kg/m)
   1           0    # Massless rigid link
```
```
------------------------ JOINT ADDITIONAL CONCENTRATED MASSES--------------------------
             1   NCmass      - Number of joints with concentrated masses; Global Coordinate System
CMJointID       JMass            JMXX             JMYY             JMZZ          JMXY        JMXZ         JMYZ        MCGX      MCGY        MCGZ
  (-)            (kg)          (kg*m^2)         (kg*m^2)         (kg*m^2)      (kg*m^2)    (kg*m^2)     (kg*m^2)       (m)      (m)          (m)
   2           3852180        2.56193E9         2.56193E9         4.24265E9       0            0           0            0        0            0
```
When using OpenFAST dev-tc without the fix (https://github.com/OpenFAST/openfast/tree/dev-tc), I can see that the behavior of the SubDyn approach is wrong:
image

When using OpenFAST dev-tc with the fix proposed here, the behavior is the proper one:
image

For reference, the output using ElastoDyn or SubDyn is not exactly the same. But the bug not accounting for the proper buoyancy or mass when a rigid link was connected to the interface joint in SubDyn has been fixed.

Next, I will test the issue https://github.com/OpenFAST/openfast/issues/854 and will check if this fix addresses that issue as well.

RBergua:
After compiling the SubDyn driver and running the standalone version, I can confirm that the issue #854 has been fixed as well. Yay!

For reference, the model included 2 rigid links and one beam connected to the interface joint. See below for reference:
image

The original issue was described as "rigid links connected to the interface and cable elements". However, the real problem were the rigid links connected to the interface joint.

The results are:
image

As can be observed, SubDyn with the fix returns the expected forces: IntfZss around 2.23e7 N.

andrew-platt:
Backport to 4.1.2

PR #2941, opened 2025-08-02
HD: Switched to an equivalent but numerically more reliable expression for the hydrostatic moment on (partially wetted) circular endplates

luwang00:

The previous expression for the hydrostatic moment on a circular member endplate can occasionally give inaccurate/incorrect results due to a numerical precision problem caused by dividing a number that is very close to zero. This PR replaces the old expression with an equivalent but more reliable one that completely avoids the possibility of division by zero. 

HydroDyn

No change to existing test results.


andrew-platt (approved):
Approving after our conversation about the method of derivation (I haven't verified it myself).

PR #2948, opened 2025-08-06
AeroAcoustics code cleanup

bjonkman:
This PR updates some of the code related to the AeroAcoustics module. It 
- removes unused code
- fixes some numerics and error handling:
   - adds checks that AoA and Re are entered in increasing order and the use the same values in each BL table (the way the code expects it)
   - adds checks so that `log10(0)` does not occur. I've added a parameter, `AA_EPSILON` set to 1E-16, to indicate the smallest value allowed in the log equation.
   - calculates LE_Location in both UpdateStates and CalcOutput instead of the previous method that used old values in UpdateStates
   - uses bi-linear interpolation in the BL_Param_Interp() routine, allowing nearest-neighbor extrapolation for AoA as well as Re (previously allowed only Re).
- calls AA_End in the AD_End routine, and in it, closes the text output files that AA had opened
- fixes an issue where ADI_End was called twice, potentially causing memory issues.
- updates the file output:
   - uses ".AA" instead of ".NN" as the module abbreviation used in the output file naming convention
   - puts the AA output file(s) in the main output directory (now relative to the _output_ file locations, not the AA _input_ file location)
   - makes sure the header in the "4.out" file contains the same number of lines as the other text output files (and follows the standard OpenFAST output format)
   - changes the formatting of the blade/node output channel names to match the AB#_N###_ChannelName format that is used in AeroDyn nodal outputs
   - adds a parameter, `AA_OutputToSeparateFile` that can be hardcoded to allow the outputs normally put in 4 separate text files (TRUE) to be sent to AeroDyn's WriteOutput array for output to the main OpenFAST output file (FALSE). It is set to TRUE now  
- allows the user to specify "default" as the AA output file name to use the `<RootName>.AD.AA#.out` naming convention
- makes sure echo file contains all the lines it reads

none

AeroDyn/AeroAcoustics

I don't believe that the current r-test framework checks the results of the AA text output files. I am aware of only one aeroacoustics r-test, and its output files are basically unchanged.

### DIFF
--- a/modules/aerodyn/src/AeroAcoustics.f90
+++ b/modules/aerodyn/src/AeroAcoustics.f90
@@ -784,6 +784,12 @@ subroutine AA_CalcOutput( t, u, p, x, xd, z, OtherState, y, m, ErrStat, ErrMsg)
     
 end subroutine AA_CalcOutput
 !----------------------------------------------------------------------------------------------------------------------------------
+REAL(ReKi) FUNCTION Log10AA(X) RESULT(F)
+   REAL(ReKi),INTENT(IN) :: X
+   
+    F = LOG10( MAX(AA_EPSILON, X) )
+    
+END FUNCTION Log10AA
 !----------------------------------------------------------------------------------------------------------------------------------!
 SUBROUTINE Calc_LE_Location_Array(p,m,u)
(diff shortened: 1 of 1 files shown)


andrew-platt (approved):
thanks for fixing these issues and reorganizing the code!

PR #2953, opened 2025-08-07
ED: Implemented the ability to add concentrated masses to the tower

luwang00:
This PR adds the ability to prescribe concentrated masses for the tower in ElastoDyn for convenience. Note that this is based on an approximated formulation where each concentrated mass is evenly spread over the discretized tower element it is located on. To more accurately localize the additional mass on the tower, finer tower discretization (increased `TwrNodes`) in the primary ElastoDyn input file is necessary. 

A list of concentrated masses for the tower can be specified at the end of the ElastoDyn tower input file in an optional section that comes after TOWER SIDE-TO-SIDE MODE SHAPES. An example with two concentrated masses (`NTwCMass` = 2) on the tower is shown below. The height fraction `HtFract` must be in the range of [0,1], and the mass `TwCMass` must be non-negative. The entries do not have to follow any particular order, and multiple entries can be assigned to the same `HtFract`.
```
---------------------- TOWER CONCENTRATED MASSES ------------------------------
          2   NTwCMass    - Number of tower concentrated masses
  HtFract       TwCMass
   (-)            (kg)
  0.324           6700
  0.727          10300
```
If this optional section is omitted, no concentrated mass is included, allowing existing ElastoDyn input files to work as normal.

ElastoDyn

No change to existing input files or test results. No verification was performed. However, through the ElastoDyn summary file, it is confirmed that the addition of concentrated masses does modify the total tower mass and the local tower node mass density correctly.

### DIFF
--- a/modules/elastodyn/src/ElastoDyn.f90
+++ b/modules/elastodyn/src/ElastoDyn.f90
@@ -3023,7 +3023,7 @@ SUBROUTINE SetTowerParameters( p, InputFileData, ErrStat, ErrMsg )
 
    ! Local variables:
    REAL(ReKi)                               :: x                            ! Fractional location between two points in linear interpolation
-   INTEGER(IntKi )                          :: J                            ! Index for the node arrays
+   INTEGER(IntKi)                           :: I, J                         ! Index for the node arrays
    INTEGER(IntKi)                           :: InterpInd                    ! Index for the interpolation routine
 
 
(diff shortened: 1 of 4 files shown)


modules/elastodyn/src/ElastoDyn.f90
```diff
@@ -3074,7 +3074,23 @@ SUBROUTINE SetTowerParameters( p, InputFileData, ErrStat, ErrMsg )
...
       p%StiffTFA  (J) = InterpStp( p%HNodesNorm(J), InputFileData%HtFract, InputFileData%TwFAStif, InterpInd, InputFileData%NTwInpSt )
       p%StiffTSS  (J) = InterpStp( p%HNodesNorm(J), InputFileData%HtFract, InputFileData%TwSSStif, InterpInd, InputFileData%NTwInpSt )
+
    END DO ! J
+
+   DO J=1,InputFileData%NTwCMass
+
+         ! Add contributions from concentrated masses. Find the tower element on which the concentrated mass is located.
+      DO I=1,p%TwrNodes
+         IF ( (p%HNodesNorm(I)+0.5*p%DHNodes(I)/p%TwrFlexL) > InputFileData%TwCMassHtFract(J) ) THEN
```
jjonkman: This test means that if HtFract equals the point half-way between two tower nodes, the higher node will receive the mass.  However, this test will never place a mass at HtFract = 1.  Suggest changing ">" to ">=" to capture this, in which case in the example above, the mass will go to the lower node.

jjonkman (commented):
Overall, I'm happy with this change.  Please find one small comment below.

modules/elastodyn/src/ElastoDyn.f90
luwang00: Thanks for the careful review! I think even though the test always returns false with `HtFract = 1`, `I` will still land on `p%TwrNodes`, and the extra mass will be put there. However, I do agree that changing `>` to `>=` makes the behavior more obvious. I will get this fixed!

jjonkman (approved):
Thanks for addressing my comment; looks good now!

andrew-platt (approved):

PR #2957, opened 2025-08-12
More MoorDyn Output Channels in openfast-io chans

dzalkind:
Enable the writing of more output channels to MoorDyn inputs when using openfast-io.

This probably should be redirected from `main`, but I'm not sure where it fits best.  Ideally, this would be included in an upcoming minor/bugfix release.  I can use a local branch in the meantime.

[This is where WEIS updates the outlist](https://github.com/WISDEM/WEIS/blob/8fc9b503b926c76d580cacf7ef078e014cfd7ba5/weis/aeroelasticse/runFAST_pywrapper.py#L170), and if there are more than 9 mooring line outputs, they are ignored.

openfast-io


mayankchetan (approved):
Looks good to me! Thanks Dan!

PR #2959, opened 2025-08-13
Add connection from ElastoDyn platform motion mesh to ServoDyn

dzalkind:
Add connection from ElastoDyn platform motion mesh to ServoDyn.
Minor update in StC file I/O.

When using `dev-tc` and extended avrSWAP, the platform motion channels are zeros.

FAST_Mapping
ServoDyn
BladedInterface_Ex

I'm not sure if there's a regression test that checks these values in the avrSWAP, but I can help make one, if desired.


andrew-platt:
This must have been lost in the tight coupling.  The connection exists in `dev`.

andrew-platt:
@deslaughter the new method of handling these mappings is really nice!!!

deslaughter (approved):

deslaughter:
This mesh mapping got lost in the new glue code since it wasn't used in linearization. I need to add some code so that this isn't included in linearization. I'm a little surprised that it didn't break any of the linearization cases.

andrew-platt:
In this case, this info is only passed to the controller.  The basic controllers don't use the info, so if they are included in linearization there wouldn't be any affect.

deslaughter:
@dzalkind I updated `dev-tc` with the latest code from `dev` and now there's a conflict in `FAST_reader.py`. Could you resolve the conflict and I'll get this merged?

dzalkind:
Done! Thanks!!

deslaughter (approved):

PR #3007, opened 2025-09-23
Cubic interpolation for SeaState and HydroDyn

luwang00:
This PR adds a new set of generic 3D and 4D uniform grid interpolation tools under nwtc-library. The new interpolation tools:

* Currently only supports 3D and 4D interpolation. Can add 2D interpolation in the future if needed.
* Supports interpolation of scalar data, 3-by-1 vector data, and 6-by-1 vector data.
* Will use C1 continuous cubic interpolation if possible. Toward grid boundary, quadratic interpolation is used. Linear interpolation is used if only two grid points are present.
* If one of the grid dimensions has only 1 grid point, this dimension is effectively ignored.
* Each dimension can be treated as either periodic or not.
* Periodic dimensions will always use cubic interpolation, even if only two grid points are present, because the grid is effectively infinite in this direction.
* For non-periodic dimensions, an out-of-bound query point will be snapped back to the grid boundary. A warning will be given at the first occurrence.
* Note that for periodic dimensions, the subroutines assume the first and last grid points are NOT duplicates of each other.
* Supports a more efficient way to compute surface slopes and normal vectors, replacing the previous method based on central differencing.
* Are now used by SeaState and HydroDyn for interpolating wave kinematics and potential-flow wave excitation.

The new grid interpolation tools are developed with SeaState and HydroDyn in mind, which previously relied on linear interpolation. The present cubic interpolation is expected to be significantly more accurate even on coarser grids and with larger time steps, saving memory. The new interpolation tools are also generic and can potentially be used by other modules in the future.

There are merge conflicts with `dev` due to some recent changes by @andrew-platt that need to be resolved. 
This PR will also create merge conflicts with `dev-tc`.

nwtc-library, seastate, hydrodyn, moordyn

With the switch from linear to cubic interpolation, some r-test cases involving SeaState and HydroDyn now have slightly different results. These were updated.

**To-Dos:**
- [x] Resolve conflict with `dev`
- [x] Update r-test results


andrew-platt:
- [x] Visual Studio updates

andrew-platt:
- [x] merge r-test

andrew-platt (approved):
Looks good!  I'll merge `dev` in and test with VS before merging.

modules/nwtc-library/src/GridInterp.f90
```diff
@@ -0,0 +1,885 @@
...
+   real(ReKi),           intent(in   )  :: isopc       ! isoparametric coordinates
+   integer(IntKi),       intent(in   )  :: support
+   real(ReKi),           intent(inout)  :: N1D(4)
+   real(ReKi)                           :: isopc2
+
+   if ( Support == 3 ) then  ! Cubic interpolation
+
+      isopc2 = isopc*isopc
+
+      N1D(1) = -1.5*isopc2 + 2.0*isopc - 0.5
```
deslaughter: These floating point values need the `_ReKi` type specifier to ensure that they're the correct type. I know the compiler should do the right thing based on how `DOUBLEPRECISION` is specified, but I've gotten burned by this before.

modules/nwtc-library/src/GridInterp.f90
```diff
@@ -0,0 +1,885 @@
...
+   ! Compute normalized coordinate
+   p = (pIn-pZero) / delta
+
+   if (isPeriodic) then
+
+      ! Calculate normalized coordinate between 0 and 1
+      isopc = p - floor(p,ReKi)
+
+      ! Get the normalized coordinates of the two nearest nodes to the left and to the right
+      Indx = floor( p, IntKi ) + (/-1,0,1,2/)
```
deslaughter: I recommend using `[ ]` to enclose arrays in new code.

modules/nwtc-library/src/GridInterp.f90
```diff
@@ -0,0 +1,885 @@
...
+   character(*), parameter                      :: RoutineName = 'GridInterp3DR4'
+   real(SiKi)                                   :: GridInterp3DR4
+   integer(IntKi)                               :: i,j,k
+
+   ! interpolate
+   GridInterp3DR4 = 0.0_SiKi
+   do k = 1,4
+      do j = 1,4
+         do i = 1,4
+            if (.not.EqualRealNos(m%N3D(i,j,k),0.0_ReKi)) then
```
deslaughter: What is this check trying to prevent? Having an `if` statement in a kernel like this is going to be bad for performance. It would be better to multiply by zero than to perform this check 64 times.

I'm not sure that I understand how this is checking for the index being out of bounds.

modules/nwtc-library/src/GridInterp.f90
```diff
@@ -0,0 +1,885 @@
...
+End Subroutine SetIndex
+
+Subroutine GetN1D(isopc, support, N1D)
+
+   real(ReKi),           intent(in   )  :: isopc       ! isoparametric coordinates
+   integer(IntKi),       intent(in   )  :: support
+   real(ReKi),           intent(inout)  :: N1D(4)
+   real(ReKi)                           :: isopc2,isopc3
+
+   if ( Support == 3 ) then  ! Cubic interpolation
```
deslaughter: I recommend using a `select case(Support)` construct here as it typically avoids evaluating multiple `if` statements and just jumps to the appropriate code.

deslaughter (changes requested):

modules/nwtc-library/src/GridInterp.f90
andrew-platt: I agree.  We can just remove the check and it should do exactly the same thing, but will optimize better.

Also applies to `GridInterp3DR8`

luwang00: Hi @deslaughter and @andrew-platt, if the weight `N3D` entry is zero, it indicates that the corresponding point in `data` is out of the grid. In each dimension, there are 4 possible configurations: query point next to lower boundary, next to the upper boundary, caught between two boundary nodes (grid only has two nodes in that dimension), and away from boundary (or periodic dimension), leading to 4^4 possible combinations.

There could be room for optimization, but the current implementation at least keeps the logic simple and less prone to bugs. When a neighboring point is out of bound in a dimension, the `GetN1D` and `GetN1Ddx` function leaves a zero for that position. This leads to `GridInterpSetup3D` and `GridInterpSetup4D` marking all nodes in the 3D or 4D grids that are out of bounds with a zero weight through multiplication.

modules/nwtc-library/src/GridInterp.f90
luwang00: Done!

modules/nwtc-library/src/GridInterp.f90
luwang00: Done!

modules/nwtc-library/src/GridInterp.f90
luwang00: Following @deslaughter's suggestion,  updated the `SetIndex` subroutine to clamp all `Indx` to the grid boundary if out-of-bound. These nodal values will be multiplied by zero during the interpolation, so this change will not affect the results. With this change, we can remove the check for zero weight in all interpolation subroutines for performance.

modules/nwtc-library/src/GridInterp.f90
luwang00: Done!

PR #3009, opened 2025-09-25
Linearization: Fix pointer to dUdy block for SD inputs to MAP outputs

bjonkman:
Linearization with MAP++ connected to SubDyn could have placed a block matrix in the wrong place in the dUdy matrix if ElastoDyn's PlatformPtMesh and SubDyn's y3Mesh don't have the same number of nodes. 

None that I know of. 

Linearization with MAP++ and SubDyn modules enabled.

I don't think any of the current regression tests have a model for this scenario.


andrew-platt:
- [ ] backport to 4.1.3, if we do such a release.

Issue #3018, opened 2025-10-01
Turbsim uses up all available threads when not compiled with OpenMP


mayankchetan:
**Bug description**
<A clear and concise description of the bug.>
When running `turbsim` on a system that has OpenMP ( or OpenMP loaded in case of HPC), but is not compiled with `-DOPENMP=ON`, `turbsim` runs on all available threads. This can bog down the system, or in some cases crashes HPC nodes.

Current workaround is to use `export OMP_NUM_THREADS=1`, but this can affect other processes that rely on OpenMP like OLAF.

**To Reproduce**
<Update the following list with your specific information.>
1. Compile with `-DOPENMP=OFF`
2. have or load OpenMP on system.
3. Run `turbsim` with `r-test/glue-codes/openfast/5MW_Baseline/Wind/90m_12mps_twr.inp`
4. Use top/htop to view resources used by turbsim.
5. Also see ` ####### CPU seconds used.` after `turbsim` completes running.

**Expected behavior**
<A clear and concise description of what you expected to happen.>
`turbsim` does not use all threads when compiled with `-DOPENMP=OFF` & system has/loaded OpenMP.

**Screenshots, if applicable**
<Add screenshots to help explain your problem.>

1. When OpenMP is not loaded on an HPC node:

<img width="1068" height="285" alt="Image" src="https://github.com/user-attachments/assets/849e5f97-5cb9-4c4c-aff1-8708ef9c9f56" />

```
 **************************************************************************************************
 TurbSim

 Copyright (C) 2025 National Renewable Energy Laboratory
 Copyright (C) 2025 Envision Energy USA LTD

 This program is licensed under Apache License Version 2.0 and comes with ABSOLUTELY NO WARRANTY.
 See the "LICENSE" file distributed with this software for details.
 **************************************************************************************************

  Running TurbSim (from OpenFAST-v4.1.2)


  Reading the input file "90m_12mps_twr.inp".
  Calculating the spectral and transfer function matrices:
     u-component matrices (IEC coherence method)
     v-component matrices (identity coherence)
     w-component matrices (identity coherence)
  Generating time series for all points:
     u-component
     v-component
     w-component
  Computing hub-height statistics
  Writing statistics to summary file
  Generating AeroDyn binary time-series file "90m_12mps_twr.bts"

  Processing complete.  38.71 CPU seconds used.

  TurbSim terminated normally.
```

2. When OpenMP is loaded on an HPC node:

[...]

PR #3020, opened 2025-10-02
TurbSim: OpenMP parallelization in factorization and FFT

andrew-platt:
Needs some review and testing.

_TurbSim_ has been single-threaded forever.  This PR adds parallelization through OpenMP and should in theory run faster as a result.

If _OpenMP_ is enabled at compile time, parallelization for MKL will be automatically turned off.  We saw this the MKL parallelization is an issue in #3018 when the MKL would try to parallelize some tiny matrix manipulation incurring massive overhead costs.

Still need to extend to the following routines:
- [ ] `CalcFourierCoeffs_API`
- [ ] `CalcFourierCoeffs_General`
- [x] `CalcFourierCoeffs_None`

#3018

_TurbSim_ only

* ~A `PRIVATE(<allocatable_array>)` in a `!$OMP PARALLEL DO` pragma doesn't always allocate the copy of that array.  This leads to segfaults.  I have messy workarounds for this.  There may be a better way to do it, but I have no idea if the _ROCM_ _Flang_ compiler will be able to handle anything more advanced.~
* Using `THREADPRIVATE` with `COPYIN` works better than the original solution

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,6 +125,11 @@ if (OPENMP)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
     link_libraries("${OpenMP_CXX_LIBRARIES}")
   endif()
+elseif (NOT BLA_VENDOR)
+  # If we're not using OpenMP, and a specific BLAS vendor has not been set, 
+  # set MKL threading to sequential to avoid potential issues with 
+  # small calculations taking longer due to threading overhead (turbsim).
+  set(MKL_THREADING "sequential")
 endif()
 
 #-------------------------------------------------------------------------------


modules/turbsim/src/TSsubs.f90
```diff
@@ -195,41 +156,18 @@ SUBROUTINE CalcFourierCoeffs_IEC( p, U, PhaseAngles, S, V, TRH_in, ErrStat, ErrM
...
          ! -----------------------------------------------
          
-         CALL Coh2H(    p, IVec, IFreq, TRH, S, ErrStat2, ErrMsg2 )
-         !$OMP CRITICAL 
-         CALL SetErrStat(ErrStat2, ErrMsg2, ErrStat, ErrMsg, 'CalcFourierCoeffs_IEC')
-         !$OMP END CRITICAL
+         CALL Coh2H(p, IVec, IFreq, TRH, S, ErrStat2, ErrMsg2)
          if (ErrStat2 >= AbortErrLev) then
-            OMPerr = .true.
+            !$OMP CRITICAL
```
andrew-platt, line 161: Nice!  that's a good way to handle it -- reduces the number OMP CRITICAL usages dramatically

andrew-platt (commented):
@deslaughter's changes are great - a much better method than what I proposed.

modules/turbsim/src/TSsubs.f90
```diff
@@ -149,27 +156,18 @@ SUBROUTINE CalcFourierCoeffs_IEC( p, U, PhaseAngles, S, V, TRH, ErrStat, ErrMsg
...
    END DO !IVec   
                
-   CALL Cleanup()
-   RETURN
-   
-!............................................
-CONTAINS
-   SUBROUTINE Cleanup()
-
-      IF ( ALLOCATED( Dist      ) ) DEALLOCATE( Dist      )
```
bjonkman, line 170: In this subroutine, we are allocating memory, but with removing the cleanup() reoutine, we are not explicitly deallocating it anymore. I know this routine is called only once, so maybe isn't a big concern, but I've seen this cause memory leaks or allocation errors because the array is already allocated when the routine gets called a second time.  Thoughts?

deslaughter: These arrays will be automatically deallocated when they go out of scope when the function returns. If the function is called a second time then the local array should start as deallocated unless marked with the`SAVE` attribute.

bjonkman: I've had issues with some compilers not actually following that behavior. Maybe it was a compiler bug or maybe they were trying to optimize something, I'm not sure.

andrew-platt:
Merging without `CalcFourierCoeffs_General` or `CalcFourierCoeffs_API` routines.  These don't appear to be as commonly used.

PR #3025, opened 2025-10-06
[BugFix] FF awae: set v_plane=0 when center of wake plane leaves domain

andrew-platt:
Ready to merge, pending test results.

If a wake plane center leaves the domain, set the velocity of the plane to zero.  In low wind speed cases where the center of a skewed wake plane was leaving domain, the remaining sampled region of the plane could get very small and have negative weighting.  If this happens with a bit of flow reversal, the plane could suddenly jump backwards by a substantial distance.  This was leading to some strange looking results and odd interactions with turbines near that edge of the domain.

Issue reported internally

Wake velocity calculations in the `AWAE` module only.

Screenshot 2025-10-06 at 5 17 26 PM
Wakes at `T=t`

Screenshot 2025-10-06 at 5 17 44 PM
Wakes at `T=t+1`.  The wake jumps backwards as a result of an unusual weighting calculation on a partial wake region with flow reversal at the far right low resolution boundary.

No test results change.

### DIFF
--- a/modules/awae/src/AWAE.f90
+++ b/modules/awae/src/AWAE.f90
@@ -380,6 +380,7 @@ subroutine LowResGridCalcOutput(n, u, p, xd, y, m, errStat, errMsg)
    integer(IntKi)      :: maxN_wake
    integer(IntKi)      :: WAT_iT,WAT_iY,WAT_iZ  !< indexes for WAT point (Time interchangeable with X)
    integer(IntKi)      :: errStat2
+   character(ErrMsgLen):: errMsg2
    character(*), parameter   :: RoutineName = 'LowResGridCalcOutput'
    logical             :: within
    real(ReKi)     :: yHat_plane(3), zHat_plane(3)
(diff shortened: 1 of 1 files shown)


ptrbortolotti:
and thank you @andrew-platt and @jjonkman !

jjonkman:
@ptrbortolotti -- Thanks for verifying that the algorithm works as intended.  @andrew-platt and I discussed a small change whereby the wake planes get pushed outside of boundary of the low-resolution domain once their wake centers cross the boundary; @andrew-platt will implement this change before this PR is finalized.

jjonkman:
Hi @andrew-platt,

Thanks for making this change.  I agree this is the approach we discussed.  However, I can still foresee a problem if one wake plane gets pushed out of the domain and those around it get close to the boundary, but don't get pushed out (e.g. when flow is nominally along X and a plane gets pushed out from the Y or Z boundaries).  This could stretch the wake volumes between the plane that gets pushed out and those that remain in the domain.  Getting around this would require adding some nominal free-stream velocity to the wake plan that gets pushed out, e.g., as is calculated to move the WAT box.

andrew-platt:
Could we run into an issue with the free-stream velocity at the wake plane where the velocity is low so the plane only partially exits the domain?  Could that land us back in the scenario that started this issue?

Or are you thinking we could apply this over multiple timesteps until the entire wake plane has fully left the domain?

ptrbortolotti:
I've run a comparison between the two recent commits and I see some minor differences in some signals. Here I plot rotor speed for T3
image

T2 also shows these minor differences, whereas T1s are identical

jjonkman:
@andrew-platt -- I'm thinking we should sum the nominal free-stream velocity with the velocity normal to the boundary that you already added.  Likewise, use the nominal free-stream-velocity for any wake plane that has already left the domain.  In this way, any wake plane that crosses a boundary will be pushed out, but once it is out, it will continue propagating downstream to prevent wakes from "bunching" up" near the boundaries where they exited the domain.

ptrbortolotti:
comparison1
comparison2


everything looks good to me both in openfast outputs and flow from the vtk files

ptrbortolotti:
Screenshot 2026-01-14 at 3 27 46 PM

modules/awae/src/AWAE.f90
```diff
@@ -631,6 +631,103 @@ subroutine LowResGridCalcOutput(n, u, p, xd, y, m, errStat, errMsg)
...
+            V_plane(2) = (D_tgt - p_plane(2)) / (2.0_ReKi * real(p%dt_low,ReKi))    ! push towards (-Y_bound - D_wake)
+         case (-3_IntKi)         ! Crossed -Z
+            D_tgt = p%Grid_Low(3,1) - D_wake
+            V_plane(3) = (D_tgt - p_plane(3)) / (2.0_ReKi * real(p%dt_low,ReKi))    ! push towards (-Z_bound - D_wake)
+         case ( 3_IntKi)         ! Crossed +Z
+            D_tgt = p%Grid_Low(3,p%NumGrid_low) + D_wake
+            V_plane(3) = (D_tgt - p_plane(3)) / (2.0_ReKi * real(p%dt_low,ReKi))    ! push towards (+Z_bound + D_wake)
+      end select
+
+      ! Step 3: add background XY plane flow to keep plane drifting (will have already returned on any planes still in bounds)
```
jjonkman: Why is this XY only, and not 3D?  I wouldn't expect much mean velocity in Z, but I'm not sure why you chose this.

jjonkman (commented):
The overall approach makes sense to me and matches what we discussed.  Just one comment below.

modules/awae/src/AWAE.f90
andrew-platt: Initially I was only including the main wind component assuming it lies along `X`.  Since the dominant wind direction could theoretically lie along `Y`, I added that in.  My underlying assumption was that only the XY plane was important to consider since the `Z` component is almost always small, so I simply ignored it.

However, since that `Z` component is almost always small, there is no reason to not include it.  I'll add it in.

modules/awae/src/AWAE.f90
```diff
@@ -725,7 +725,7 @@ subroutine PlaneOutOfDomain(D_Wake,p_plane,V_plane,planeDomainExit,ErrStat3,ErrM
       end select
 
       ! Step 3: add background XY plane flow to keep plane drifting (will have already returned on any planes still in bounds)
```
jjonkman: Comment needs to be updated. Background flow is now full XYZ.

jjonkman (approved):

Issue #2027, opened 2024-02-10
FAST Farm T1 initialize Error


godsentaustine:
i am running FAST Farm with the OC3 Hywind configuration. and I'm getting this error after updating the ''--- AMBIENT WIND: INFLOWWIND MODULE'' parameters. I've tried to figure out where this is coming from. any help will be appreciated as i'm quite new using FAST Farm.

![image](https://github.com/OpenFAST/openfast/assets/126730108/8cc405af-c3ef-404a-bf65-655894c31a93)

![image](https://github.com/OpenFAST/openfast/assets/126730108/17b17ab1-ff82-42cd-9152-26262d755e60)

jjonkman:
Dear @godsentaustine,

Similar questions were asked and answered in the following OpenFAST issues: https://github.com/OpenFAST/openfast/issues/1613 and https://github.com/OpenFAST/openfast/issues/1044.

Best regards,

godsentaustine:
Dear @jjonkman ,
Thnaks for your reply. I've been able to follow the [python-toolbox](https://github.com/OpenFAST/python-toolbox/tree/main) to generate FAST.Farm input. (The Farm layout is a 3x2 and are 5D from each other). The error i was getting was still coming up not until i changed dr = cmax. that appeared to have resolved that error. 
The error below is what i'm currently encountering. Does this have to do with the number of wake planes? because i did try to increase it which did'nt work. Where can I resolve this?
![image](https://github.com/OpenFAST/openfast/assets/126730108/b1ef901e-a58d-4643-880c-8c83940b158a)

![image](https://github.com/OpenFAST/openfast/assets/126730108/e21f413c-9986-46f9-9738-be5b48365b4b)

![image](https://github.com/OpenFAST/openfast/assets/126730108/e90c7528-00b3-425d-91f6-3a5795fac8f7)

godsentaustine:
The Y0_Low parameter sets the origin of the Y boundary to -80.0 m, and with NY_Low being 8 and dY_Low being 18.608 m, this gives a total coverage in the Y direction of only 8 * 18.608 m = 148.864 m, which doesn't seem sufficient to cover turbine #4's Y position (630 m). my thinking is i'd need to adjust Y0_Low and potentially NY_Low or dY_Low to ensure the low-resolution grid covers these turbines. Is this reasonable? since i'd be altering the calculations from the modelling guidance which i presume the toolbox follows.

jjonkman:
[...]

Issue #1613, opened 2023-06-08
FAST.Farm Mesh Mapping: Node 1 does not project onto any line2 element.


lusoalman:
I want to do a simple steady FAST.Farm simulation with two wind turbine models with D=1.2 m. When I run FASTFarm v3.5.0 on my windows 10 I get this error for both turbines: Farm_Initialize:T1:Farm_InitFAST:FWrap_Init:MeshMapCreate:CreateLoadMap_L2_to_L2:CreateMapping_Pr jectToLine2:Node 1 does not project onto any line2 element. Closest distance is 0.12987 m.

According to previous issues in the community my nodal points from the .fstf the aerodyn and the elastodyn files are not matching. But I can't find the problem. 

[WTM_D1d2_Aerodyn_T1.txt](https://github.com/OpenFAST/openfast/files/11687162/WTM_D1d2_Aerodyn_T1.txt)
[WTM_D1d2_ElastoDyn_T1.txt](https://github.com/OpenFAST/openfast/files/11687163/WTM_D1d2_ElastoDyn_T1.txt)
[WTM_D1d2_2T5D.txt](https://github.com/OpenFAST/openfast/files/11687164/WTM_D1d2_2T5D.txt)
[WTM_D1d2_T1_6mps.txt](https://github.com/OpenFAST/openfast/files/11687165/WTM_D1d2_T1_6mps.txt)

jjonkman:
Dear @lusoalman,

Actually, the `CreateLoadMap_L2_to_L2` error is regarding the mesh-mapping search in the OpenFAST wrapper within FAST.Farm, which is the mesh-mapping search between the AeroDyn Line2 blade load output mesh and the radial finite-difference grid that is stored on a mesh in the OpenFAST wrapper.  I would guess the issue is in your AeroDyn blade input file, _WTM_D1d2_Aerodyn_Blade.dat_, which you have not shared.  A similar question was asked and answered in the following issue: https://github.com/OpenFAST/openfast/issues/1044.

Best regards,

lusoalman:
Thank you for the quick answer! 
Here is the missing file:
[WTM_D1d2_Aerodyn_Blade.txt](https://github.com/OpenFAST/openfast/files/11688735/WTM_D1d2_Aerodyn_Blade.txt)
What is the AeroDyn Line2?

lusoalman:
Thank you again for the help! I will look into it and check everything again step by step. 

All the best, 
Sara

Issue #1044, opened 2022-03-15
Error while running Fast Farm


Tommi1198:
Hello evryone, I encountered an error while running the primary input file of Fast Farm...is there someone who can tell me how to fix it and what to do? Thank you very much!
The error is the following: "forrtl: severe (157): Program Exception - access violation"

Tommi1198:
Sorry but I cannot understand what to do to solve my problem, since I cannot found the solution for my type of error

andrew-platt:
Hi @Tommi1198,

Which version of FAST.Farm are you running? In addition to the issue @jjonkman mentioned, there may be another possibility depending on which version you are using.

Regards

Tommi1198:
Dear @jjonkman , I tried to change the code noted in PR #769 , but I still have the same kind of error: "forrtl: severe (157): Program Exception - access violation"
Dear @andrew-platt , my version is OpenFAST v3.0.0, since my professor told me to use that one. 
How can I fix this error?

Tommi1198:
Dear @jjonkman , the change that you refer to is only in IfW_FFWind_Base.F90 right? Only a line of change with respect to the original one? Am I correct?

jjonkman:
Dear @Tommi1198,

Yes, that is correct.   Can you share the full set of messages written to the command window after executing FAST.Farm?

Best regards,

Tommi1198:
Dear @jjonkman , here what you asked me:

FAST.Farm

Copyright (C) 2021 National Renewable Energy Laboratory
Copyright (C) 2021 Envision Energy USA LTD

This program is licensed under Apache License Version 2.0 and comes with ABSOLUTELY NO WARRANTY.
See the "LICENSE" file distributed with this software for details.
**************************************************************************************************

FAST.Farm-v3.0.0
Compile Info:
 - Compiler: Intel(R) Fortran Compiler 1900
 - Architecture: 64 bit
 - Precision: single
 - OpenMP: No
 - Date: Jun 23 2021
 - Time: 21:41:28
Execution Info:
 - Date: 03/17/2022
 - Time: 14:17:59+0100

 Heading of the FAST.Farm input file:
     Sample FAST.Farm input file
Running AWAE.
Running InflowWind.
forrtl: severe (157): Program Exception - access violation
Image              PC                Routine            Line        Source

Tommi1198:
Dear @jjonkman , this problem is shown if I want to use the supercontroller, maybe because I have not the SuperController file...where can I find it?
[...]

PR #3031, opened 2025-10-10
FF: align initial wake grid with the blades

andrew-platt:
Ready to merge

We have seen occasional mesh mapping errors during initialization of _FAST.Farm_.  This may be due to the initial position of the `m%ADRotorDisk` meshes getting set based only on the wake grid sitting on the ground.  However, the wake grid should be centered on the hub as it is in later calculations.  If the grid is less than the distance from the ground to the top of whatever blade is pointed vertically (blade 1 typically), a mesh mapping error was occurring.  This change now aligns the wake grid meshes with the blade axis during init - not technically correct, but will avoid mesh mapping issues.

#2027 
#1613 
#1044

_FAST.Farm_ initialization in the _AWAE_ module only.

This PR came about after some internal discussions about an issue with a tall model where the number of wake radii was set too low.

No test results are affected.

### DIFF
--- a/glue-codes/fast-farm/src/FASTWrapper.f90
+++ b/glue-codes/fast-farm/src/FASTWrapper.f90
@@ -77,6 +77,7 @@ SUBROUTINE FWrap_Init( InitInp, u, p, x, xd, z, OtherState, y, m, Interval, Init
    TYPE(FAST_ExternInitType)                       :: ExternInitData 
    INTEGER(IntKi)                                  :: j,k,nb      
    REAL(ReKi)                                      :: p0(3)       ! hub location (in FAST with 0,0,0 as turbine reference)
+   REAL(R8Ki)                                      :: orientation(3,3)  ! temp orientation array
    
    INTEGER(IntKi)                                  :: ErrStat2    ! local error status
    CHARACTER(ErrMsgLen)                            :: ErrMsg2     ! local error message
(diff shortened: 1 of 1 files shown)


glue-codes/fast-farm/src/FASTWrapper.f90
```diff
@@ -196,10 +197,14 @@ SUBROUTINE FWrap_Init( InitInp, u, p, x, xd, z, OtherState, y, m, Interval, Init
...
-         ! shortcut for 
-         ! call MeshPositionNode(m%ADRotorDisk(k), j, [0,0,r(j)], errStat2, errMsg2)
-         m%ADRotorDisk(k)%Position(3,:) = p%r ! this will get overwritten later, but we check that we have no zero-length elements in MeshCommit()
+         ! set node initial position/orientation
+         ! NOTE:  the mesh data for ADRotorDisk gets overwritten later, but should be aligned somewhat with
+         !        the blade during initialization to avoid mesh mapping errors. This puts it aligned with
+         !        the blade pitch axis now, but no calculations will be done until after it is reset properly
+         !        later.
+         do j=1,size(p%r)
+            m%ADRotorDisk(k)%Position(1:3,j) = m%Turbine%AD%u%rotors(1)%BladeRootMotion(k)%Position(1:3,1) + m%Turbine%AD%u%rotors(1)%BladeRootMotion(k)%TranslationDisp(1:3,1) + m%Turbine%AD%u%rotors(1)%BladeRootMotion(k)%Orientation(3,1:3,1) * p%r(j)
```
jjonkman: Is there a reason you are using the blade motion mesh here instead of the hub motion mesh used by the similar calculation in FWrap_CalcOutput()?

glue-codes/fast-farm/src/FASTWrapper.f90
```diff
@@ -76,6 +76,7 @@ SUBROUTINE FWrap_Init( InitInp, u, p, x, xd, z, OtherState, y, m, Interval, Init
 
    TYPE(FAST_ExternInitType)                       :: ExternInitData 
    INTEGER(IntKi)                                  :: j,k,nb      
+   REAL(ReKi)                                      :: p0(3)       ! hub location (in FAST with 0,0,0 as turbine reference)
```
jjonkman, line 79: It looks like you added this local variable but never use it.

glue-codes/fast-farm/src/FASTWrapper.f90
andrew-platt: Initially I did this for simplicity.  However, one of the meshes I'm using must not exist yet, so there is a segfault.  I'll have to rework it (should have tested first)

glue-codes/fast-farm/src/FASTWrapper.f90
andrew-platt: Ended up using in the revision.

glue-codes/fast-farm/src/FASTWrapper.f90
andrew-platt: I updated the note on why the method used elsewhere can't be used here.  We end up with a cone now instead of a flat disk, but it should work for initialization (we never calculate anything with this - just need to get the mapping initialized).

Issue #3022, opened 2025-10-02
MoorDyn: support a larger number of points being attached to a body


mattEhall:
MoorDyn currently limits the number of points connected to a body to 30. This prevents modeling floating platforms with larger numbers of mooring line attachment points.

We should raise this limit to a much larger number. The limit is currently dictated by the size of arrays initialized in MoorDyn, so this can easily be changed. Recommend a limit in the 100-500 range if we don't see other constraints.

This suggestion raised by Inaki Zabala at noted in https://forums.nrel.gov/t/limit-in-the-number-of-lines-conected-to-a-platform/9342

andrew-platt:
Is that an arbitrary limit in the code easily changed, or are structural changes to the code required?

mattEhall:
If I remember right, it's just a matter of array sizes. Likely some "30"s in the code that could be changed to a larger number.

andrew-platt:
Sounds like a pretty straightforward change.  We could make that a parameter so it only needs to be changed in one place.

mattEhall:
That sounds perfect Andy. We just realized this is also an issue for how many anchors are supported (because anchors are attached to a "ground" body). So some users will hit this limit quickly in FAST.Farm applications. It would be great to prioritize a fix.

andrew-platt:
completed in #3034

PR #3034, opened 2025-10-14
MoorDyn: increase number of attachment points possible for bodies/anchors/points

andrew-platt:
Ready to merge... if it works

Using _MoorDyn_ was artificially limited to 30 attachments to bodies, which was an issue for very large shared mooring _FAST.Farm_ simulations.  This PR parameterizes these limits in the registry so that it is easy to change later (this PR also bumps up the limits for now).

#3022

Very large _MoorDyn_ models with many interconnects only.

This is not the ideal way to deal with this.  A better solution is to set the limits on these arrays while reading in the model.  However the memory cost with doing it this way is not large compared to the complexity the code could end up with.


mattEhall:
Thanks Andy!  I didn't realize that something like MD_MaxBdAtch could be available within subroutines like Body_AddPoint without extra parameters being passed in.  This will be great.

PR #3065, opened 2025-11-10
New wavetank library for HIL simulations with AD+MD

andrew-platt:
Ready for merging

The _WaveTank_ glue-code is an experimental code for coupling hardware-in-the-loop MHK models in a wavetank to software simulating the MHK turbine loads that cannot be physically modeled in the wave tank. The _OpenFAST_ modules _SeaState_, _AeroDyn_, _MoorDyn_, and _InflowWind_ are statically linked into a single dynamic library (`cmake` target `wavetanktesting_c_binding`) with a c-binding based interface. This library can be called from _LabView_ or another code.

#2720 
#2735

Updates to following modules and interfaces:

- `AeroDyn_Inflow_C_Binding` library
   - Improved internal memory handling during _Init_
- `HydroDyn`
   - fix broken error handling in `HydroDyn_CalcOutput` routine
- `MoorDyn_C_Binding` library
- `SeaState_C_Binding` library
   - use `TMax` instead of number of timesteps
   - revised argument order in _Init_
   - `WaveTimeShift` for advancing wavefield in time (cannot delay)
   - direct string passing of some path/filenames instead of pointers
- `pyOpenFAST`
   - Updates to match _SeaState_ changes
   - new `tdmslib.py` for reading `tdms` files from _LabView_
- `NWTC_Library`
   - ability to set `CU` to a different value for writing to file instead of output to screen (used with LabView)

**Incomplete items and limitations**
- _LabView_ example is not functional
- No Visual Studio build process exists yet

This code is still a work in progress and should not be considered usable for generic cases.

New test cases:
- `glue-codes/openfast/MHK_RM1_Floating_Tank-scaled` - 1/49th scale version of the `MHK_RM1_Floating` case.  This is a reference case.
- `glue-codes/other/py_wavetank_test1` - 1/49th scale version of the `MHK_RM1_Floating` case using Python to call the wavetank library.  Prescribed motion from a wave tank simulation of the RM1 is used as the motion input.  This case can be used as a template

### DIFF
--- a/docs/source/user/index.rst
+++ b/docs/source/user/index.rst
@@ -32,6 +32,7 @@ This section contains documentation for the OpenFAST module-coupling environment
    TurbSim <turbsim/index.rst>
    FAST.Farm <fast.farm/index.rst>
    C++ API <cppapi/index.rst>
+   WaveTank <other/index.rst>
 
 
 Additional module documentation
--- a/docs/source/user/other/index.rst
+++ b/docs/source/user/other/index.rst
@@ -1,0 +1,40 @@
+.. _WaveTank:
+
+WaveTank
+========
+
+The WaveTank glue-code is an experimental code for coupling hardware-in-the-loop
+MHK models in a wavetank to software simulating the MHK turbine loads that
+cannot be physically modeled in the wave tank.  The  *OpenFAST* modules
+*SeaState*, *AeroDyn*, *MoorDyn*, and *InflowWind* are statically linked into a
+single dynamic library (``cmake`` target ``wavetanktesting_c_binding``) with a
+c-binding based interface.  This library can be called from *LabView* or another
+code.
+
+Inputs to the library include the time and motions, including the velocities and
+accelerations, located at a single reference poitn at each time step.  The
+resulting forces and moments are returned to the calling code.
+
+Restrictions
+~~~~~~~~~~~~
+The current setup WaveTank library has several restrictions:
+
+- rigid structure including platform, tower, and nacelle
+- no yaw DOF
+- rigid rotor
+- constant rotor RPM for entire simulation
+- no option for controller interfacing at present
+- visualization limitted to *AeroDyn* and *SeaState*
+- Current implementation only supports floating MHK turbines (``MHK = 2``). Other modes are present but not fully implemented.
+
+
+
+
+Input File
+~~~~~~~~~~
+
+
+.. toctree::
+   :maxdepth: 2
+   
+   wavetank_input.rst
--- a/docs/source/user/other/wavetank_input.rst
+++ b/docs/source/user/other/wavetank_input.rst
@@ -1,0 +1,242 @@
+.. _WaveTank-Input:
+
+Input File
+----------
+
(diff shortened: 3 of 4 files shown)


glue-codes/python/pyOpenFAST/wavetanktesting.py
```diff
@@ -12,187 +12,523 @@
...
+
+    #--------------------------------------
+    # Constants
+    #--------------------------------------
+    # NOTE: The length of the error message in Fortran is determined by the
+    #       ErrMsgLen variable in the NWTC_Base.f90 file. If ErrMsgLen is modified,
+    #       the corresponding size here must also be updated to match.
+    ERROR_MESSAGE_LENGTH: int = 8197
+    DEFAULT_STRING_LENGTH: int = 1025
+
```
deslaughter, line 206: Are these included in the `OpenFASTInterfaceType`?

modules/seastate/CMakeLists.txt
```diff
@@ -39,22 +39,39 @@ add_library(seastlib STATIC
...
-add_library(seastate_c_binding SHARED 
-  src/SeaState_C_Binding.f90
-)
-target_link_libraries(seastate_c_binding seastlib versioninfolib)
+#   create object instead of directly linking into shared and static -- causes issues in parallel builds
+#   This is only required because we are static linking the library for wavetank
+#   NOTE: target linking at the object, static, and shared libraries.  Different CMake versions handle this
+#         slightly differently with unpredictable results if I don't.
+add_library(seastate_c_binding_object OBJECT src/SeaState_C_Binding.f90)
+target_link_libraries(seastate_c_binding_object seastlib nwtclibs versioninfolib)
```
deslaughter, line 49: If you add `PUBLIC` then you shouldn't need to add `target_link_libraries` for the static and shared builds.

```suggestion
target_link_libraries(seastate_c_binding_object PUBLIC seastlib nwtclibs versioninfolib)
```

andrew-platt: I'll give that a try on all of them.

glue-codes/python/pyOpenFAST/wavetanktesting.py
andrew-platt: Good catch.
There is an inconsistency in the names used that I had not sorted out (`ERROR_MSG_C_LEN` in `OpenFASTInterfaceType` vs. `ERROR_MESSAGE_LENGTH`).  Several of the other python libraries have this inconsistency.
I'll make an attempt to revise this, but it might take a bit of time.  Alternatively, we can add it as an action item on the 5.0.0 release.

modules/seastate/CMakeLists.txt
andrew-platt: I tried this locally, but results in `undefined symbols` in the shared library.  So I'll leave this as is for now

glue-codes/python/pyOpenFAST/wavetanktesting.py
andrew-platt: Opened #3124.  Will deal with this in 5.0

deslaughter (approved):
I think this looks good, impressive work!

PR #3066, opened 2025-11-11
Use k-d tree to speed up wake application to high-res grid in AWAE

deslaughter:

This PR adds a k-d tree implementation (https://en.wikipedia.org/wiki/K-d_tree) to NWTC Library and uses it in the AWAE module to more efficiently apply the wake effects to each turbine's high-resolution inflow grid and to the overall low-resolution grid. The previous algorithm in AWAE checked every combination of every wake region of every turbine for each point in each turbine's high-res grid and for every point in the low-res grid. This was a slow process due to the large number of wind grid points, wake points, and turbines, especially for large farms. 

The new algorithm loads the wake points from all turbines into a k-d tree, finds the center of each turbine's high-res grid, calculates a search radius from that center point to 1) cover the destination turbine's high-res grid, 2) include the largest wake diameter, 3) be large enough to capture a wake which may span an entire grid. Then, the wake points that are near the high-res grid are processed and added to the grid. This significantly reduces the amount of time spent searching for which wake points are applicable to a given high-res grid point.

A similar is approach is applied to the low-res grid which is now divided into roughly equal sized chunks along the horizontal dimensions. Each chunk has a center, search radius, and grid indices which are used for separate k-d tree searches and allow for selectively updating parts of the low-res grid.

NWTC Library and AWAE module (used by FAST.Farm)

Existing FAST.Farm tests are unchanged.

### DIFF
--- a/modules/awae/src/AWAE.f90
+++ b/modules/awae/src/AWAE.f90
@@ -56,11 +56,6 @@ module AWAE
    public :: AWAE_TEST_CalcOutput
    public :: AWAE_TEST_Interp2D
 
-   real(ReKi), parameter, private :: I33(3,3) = &
-      reshape([1.0_ReKi, 0.0_ReKi, 0.0_ReKi, &
-               0.0_ReKi, 1.0_ReKi, 0.0_ReKi, &
-               0.0_ReKi, 0.0_ReKi, 1.0_ReKi], [3,3])
-
    contains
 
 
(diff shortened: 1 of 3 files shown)


modules/awae/src/AWAE.f90
```diff
@@ -54,6 +56,10 @@ module AWAE
    public :: AWAE_TEST_CalcOutput
    public :: AWAE_TEST_Interp2D
 
+   real(ReKi), parameter, private :: I33(3,3) = &
```
andrew-platt: this is a little nicer than calling the `Eye` function to generate the identity.

modules/awae/src/AWAE.f90
```diff
@@ -352,284 +374,399 @@ subroutine LowResGridCalcOutput(n, u, p, xd, y, m, errStat, errMsg)
...
+            ! Update disturbed velocity fields from local values
+            m%Vdist_low(:,ix,iy,iz) = Vdist_low
+            m%Vdist_low_full(:,ix,iy,iz) = Vdist_low_full
+            
+         end if  ! (n_wake > 0)
+      end do ! iXYZ, loop NumGrid_low points
+   end do
+   
+   !----------------------------------------------------------------------------
+   ! What is this?
```
andrew-platt: This calculates the plane velocity vector.

modules/awae/src/AWAE.f90
deslaughter: Yeah, I wouldn't do it everywhere, but it's helpful for this. Actually, it might could be removed all together since I'm calling the FlowField functions directly so we don't need the hub orientation. What do you think?

modules/awae/src/AWAE.f90
deslaughter:
```suggestion
   ! Calculate plane velocity vector
```

modules/awae/src/AWAE.f90
andrew-platt: Let's remove that.  We can also address the comment on line 1203 about the hub and disk avg vel - we don't call `IfW_CalcOutput` anymore.  Although, I'm a little unsure if that will cause issues in IfW initialization if that is random junk for the hub orientation (I don't think we check the dcm, so it might be fine).

PR #3104, opened 2025-12-11
FF WAT: change default WAT_Scalebox to True

andrew-platt:
Almost ready to merge.  See list below

This PR sets `WAT_Scalebox = .true.`, and updates the method for setting the `WAT_dXdYdZ` parameter with `WAT = 1`.

- The recommendation for wake added turbulence (WAT) is to rescale by default with.  However the code and test cases were setup with a default for `WAT_Scalebox = .false.`.  This PR changes this default to `.true.`.  Without the rescaling, cases were occasionally appearing where the velocity from the WAT was causing wakes to propagate in the negative direction.
- When `WAT = 1`, the discretization in `WAT_dXdYdZ` is now set to match the high resolution grid if all turbine hi-res domains are identical, otherwise it will fall back to calculating from recommended practices.

N/A

This affects the wake added turbulence in _FAST.Farm_ only.

Tests will change slightly

**Finish before merge**
- [x] update results (as needed)
- [x] update docs

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,6 +125,11 @@ if (OPENMP)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
     link_libraries("${OpenMP_CXX_LIBRARIES}")
   endif()
+elseif (NOT BLA_VENDOR)
+  # If we're not using OpenMP, and a specific BLAS vendor has not been set, 
+  # set MKL threading to sequential to avoid potential issues with 
+  # small calculations taking longer due to threading overhead (turbsim).
+  set(MKL_THREADING "sequential")
 endif()
 
 #-------------------------------------------------------------------------------
--- a/docs/source/user/servodyn-stc/StC_Theory.rst
+++ b/docs/source/user/servodyn-stc/StC_Theory.rst
@@ -473,7 +473,7 @@ Therefore :math:`\ddot{z}_{_{TMD_Z/P_N}}` is governed by the equations
 
 
 
-The forces :math:`F_{X_{_{TMD_Z/O_N}}}` and :math:`F_{Z_{_{TMD_Z/O_N}}}`
+The forces :math:`F_{X_{_{TMD_Z/O_N}}}` and :math:`F_{Y_{_{TMD_Z/O_N}}}`
 are solved noting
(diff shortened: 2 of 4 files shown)


andrew-platt:
There are changes in the results as the wakes evolve differently.

## _TSinflow_
At time 90 seconds.  `XY` slice at 90 m.

Screenshot 2025-12-30 at 7 11 21 AM

Before: `WAT_dXdYdZ = [5,5,5]`

Screenshot 2025-12-30 at 7 12 23 AM

After: `WAT: DxDyDz set to:   10.170   10.000   10.000 (using high res grid resolution)`

Note that the `WAT` is causing a larger amount of wake deflection in this case than would be expected indicating the default values for `k_Def` and `k_Grad` are not well suited to this case. Also note that there are other artifacts in the wake that should be investigated further.

andrew-platt:
## _TSinflow\_curl_
At time 58 seconds.  `XY` slice at 90 m.

Screenshot 2025-12-30 at 7 26 12 AM

Before: `WAT_DxDyDz = 5.0, 5.0, 5.0`

Screenshot 2025-12-30 at 7 29 05 AM

After: `WAT: DxDyDz set to:   10.170   10.000   10.000 (using high res grid resolution)`

Very minor differences in the wakes (almost imperceptible after the wake from T1 arrives at T2).

Screenshot 2025-12-30 at 7 30 17 AM

andrew-platt:
## _LESinflow_
At 48 seconds.  `XY` slice at 90 m.

Screenshot 2025-12-30 at 7 15 21 PM

Before: `WAT_DxDyDz     5.0, 5.0, 5.0`


Screenshot 2025-12-30 at 7 16 46 PM

After: `WAT_DxDyDz     5.0, 5.0, 5.0` (user defined).  `ScaleBox = .true.` introduces very minor changes to results.

andrew-platt:
## _Mod\_Amb3_

Very minor differences appear in the wake velocities (2nd decimal place, values like `W1VDisY`).

andrew-platt:
## _TSinflowADskSED_

At T=90 seconds, 90 m height.

Screenshot 2025-12-30 at 8 04 56 PM

Before: `WAT_DxDyDz    5.0, 5.0, 5.0`

Screenshot 2025-12-30 at 8 05 56 PM

After: `WAT_DxDyDz:      10.170   10.000   10.000` and `WAT_ScaleBox: .TRUE.`

andrew-platt:
The `py_seastate_1` regression test is failing.  This is fixed in #3065, which has not been merged yet.

PR #3115, opened 2025-12-23
Add loose coupling option to dev-tc, update regression tests

deslaughter:
This PR is ready to merge

This PR adds the `ModCoupling` switch to the main input file which allows the user to select between:
- `LooseCoupling`: traditional loose coupling scheme (could use more testing and benchmarking)
- `TightCoupilngFixed`: tight-coupling algorithm with a fixed Jacobian update time. Simulation will exit if the simulation doesn't converge
- `TightCouplingAdapative`: tight-coupling algorithm where the Jacobian is automatically updated and a correction step is performed if solution doesn't converge in a time step.

This PR also renames the new `RotorDir` flag to `MirrorRotor` which is more descriptive of how the flag will be used. This flag currently isn't used, but it will be for multi-rotor turbines.

Glue code, regression tests

The `AWT_WSt_StartUp_HighSpShutDown` regression test is enabled once again and uses the loose coupling scheme for the solve which is required for the high-speed brake.


andrew-platt:
- [ ] merge r-test/dev-tc (contains updates from #3117)

andrew-platt:
Looks good!

PR #3130, opened 2026-01-06
Fix FAST_Reader bug when ElastoDyn tower file path is absolute

louis-papillon:
Fix FAST_Reader bug when ElastoDyn tower file path is absolute because an else condition was missing. Works if path is relative.

No known issue.


andrew-platt (approved):

deslaughter:
@mayankchetan We should probably have a general function for doing this for any path in the input files.

andrew-platt:
I agree we should have a general function for that - sometime in the future.

PR #3133, opened 2026-01-08
[Bug Fix] Fix the element size calculation in AeroAcoustics module

bjonkman:
On last node (at the tip), the element size for aeroacoustics calculations was calculated as half the difference in span locations between the last two nodes. This should not be divided by 2. 

AeroAcoustics only

The AeroAcoustics regression test, `IEA_LB_RWT-AeroAcoustics` fails. Here are some of the failing channels:

image

image

image

image

### DIFF
--- a/modules/aerodyn/src/AeroAcoustics.f90
+++ b/modules/aerodyn/src/AeroAcoustics.f90
@@ -635,7 +635,8 @@ subroutine AA_UpdateStates( t, n, m, u, p,  xd, OtherState, errStat, errMsg )
                if (EqualRealNos(abs_le_x, 0.0_ReKi)) then
                   angletemp = 0.0_ReKi
                else
-                  angletemp = ATAN2(m%LE_Location(2,j,i), abs_le_x) * R2D_D
+                  angletemp = ATAN2(m%LE_Location(2,j,i), abs_le_x) * R2D ! returns angles in the range [-180, 180] degrees
+                  if (angletemp<0.) angletemp = angletemp + 360. ! in calculation for rco_minus1 below, we compare angles in the range [0, 360] degrees
                end if
 
                k_minus1 = 0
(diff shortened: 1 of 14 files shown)


andrew-platt:
This has a relatively large impact on the measurements for`Obs1`.  I'm not familiar enough with this to know if this is an important change or not.  @ptrbortolotti, can you give your thoughts on this?  Does this better reflect real world data, or is this within the uncertainty of measurements?

modules/aerodyn/src/AeroAcoustics.f90
bjonkman, line 674: Another question for @ptrbortolotti:
when TICalcMethod = 2, it computes this TIVx in line 673 as the _standard deviation_ of the Inflow vector (in each bin). This is NOT the TI. Should we be dividing this `xd%TIVx(j,i)` by `tempmean` to get TI here?

It gets used in the `InflowNoise()` calculation, where it assumes it's TI.

andrew-platt:
- [x] update r-test results before merge

modules/aerodyn/src/AeroAcoustics.f90
ptrbortolotti: another good catch @bjonkman. frankly I've never used `TICalcMethod =2`. back at TUM we saw that the method, even when implemented correctly, took a long time to converge and we always relied on user-defined input TI

andrew-platt: Addressed in #3149

ptrbortolotti:
hi @bjonkman and @andrew-platt. thank you for your patience. I've finally had the time to run some checks, rerunning the Case 1.1 from this this [conference paper](https://docs.nrel.gov/docs/fy23osti/85983.pdf), see Figure 7, and this [technical report](https://iea-wind.org/wp-content/uploads/2021/05/report_WTNCBenchmark.pdf), see Figure 41.

The summary is that this PR changes the results minimally.

These are the results from the conference paper. I generated those in 2021 running OpenFAST v3.0.0 (or possibly a dev branch around that time)
Screenshot 2026-01-29 at 1 55 09 PM

This is the comparison today between v3.0.0, v4.1.2, and v4.2.0. The first two match exactly, the second is slightly different

comparison_spectra

Please let me know if you'd like to see more checks

PR #3134, opened 2026-01-08
[BUG FIX] AeroAcoustics module angle range

bjonkman:
If using `TICalcMeth = 2`, the angle calculated is in the range [-180,180] degrees, but the logic lower in the UpdateStates routine assumes the angle is in the range [0, 360] degrees. 

AeroAcoustics module

The regression tests for aeroacoustics uses `TICalcMeth = 1`, so this will not affect any current tests.

### DIFF
--- a/modules/aerodyn/src/AeroAcoustics.f90
+++ b/modules/aerodyn/src/AeroAcoustics.f90
@@ -635,7 +635,8 @@ subroutine AA_UpdateStates( t, n, m, u, p,  xd, OtherState, errStat, errMsg )
                if (EqualRealNos(abs_le_x, 0.0_ReKi)) then
                   angletemp = 0.0_ReKi
                else
-                  angletemp = ATAN2(m%LE_Location(2,j,i), abs_le_x) * R2D_D
+                  angletemp = ATAN2(m%LE_Location(2,j,i), abs_le_x) * R2D ! returns angles in the range [-180, 180] degrees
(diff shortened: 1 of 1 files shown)


modules/aerodyn/src/AeroAcoustics.f90
bjonkman, line 419: Does anyone know why line 419 exists? 

`p%rotorregionlimitsrad(size(p%rotorregionlimitsrad)-1)=p%rotorregionlimitsrad(size(p%rotorregionlimitsrad)-1)+3`

I'm trying to figure out why they would add 3 meters to the second-to-last bin here. I could *maybe* see this getting modified on the last one--especially since the limit is based on only 1 blade but applied to all blades--but I'm not sure why the second-to-last region limit would be altered. @ptrbortolotti?

ptrbortolotti: oh boy no I've never realized that the code I inherited was doing that...

ptrbortolotti (approved):
I've never truly used `TICalcMeth = 2`, thanks for debugging this

modules/aerodyn/src/AeroAcoustics.f90
bjonkman: Thanks, @ptrbortolotti! 

Unless we find a reason to keep it there, I'm going to remove this logic in the near future. I'm trying to simplify some of this code so it's easier to read (and modify).

PR #3147, opened 2026-01-16
Modal damping

JustinPorter88:

Implementation of modal damping in BeamDyn.

New features are added to BeamDyn and input decks are modified. Code for the eigenanalysis in SubDyn is moved so that it can be reused in the BeamDyn calculation.


deslaughter:
@andrew-platt This PR is ready for review if you wouldn't mind giving it a quick look. I'm afraid the `IEA22MW_ModalDamping` regression test may be a little sensitive as I get slightly different results on my laptop and in Github Actions, but I think it's just numerical noise. Regardless, we may get test failures if the runners change hardware.

PR #3149, opened 2026-01-16
AeroAcoustics cleanup + added driver code

bjonkman:
This pull request cleans up some of the AeroAcoustics code. It also includes 
- a fix of how `TICalcMeth=2` calculates the TI for the inflow noise calculations,
- a new check that the `BldPrcnt` parameter in the input file is entered between 0 and 100%,
- a fix in case there are fewer than 2 blade nodes (would have caused a segmentation fault otherwise).

These updates should make the code run a bit faster, though I haven't profiled it to check.

I have also included source code for a new AeroAcoustics driver, but I've not added it to any build scripts.

This includes the element-size fix at the tip node that is introduced in https://github.com/OpenFAST/openfast/pull/3133. So, that should be merged first (and if not, we should add a division-by-two to this code).

AeroAcoustics module of AeroDyn

This should give the same results as #3133 for the aeroacoustics regression test here. 

When `TICalcMeth=2` and `TIMod /= 0`, this code will give different results because the calculated value of TI will be different in the InflowNoise routine.


andrew-platt:
I'll do a separate PR for the AA driver build through CMake.

Issue #857, opened 2021-10-05
Bug Report: If NLinTimes = 1, Linearization always occurs on first timestep


bjonkman:
**Bug description**
When NLinTimes > 1 and CalcSteady = true, OpenFAST will wait for steady state to begin linearization. However, when NLinTimes = 1 (and RotSpeed is non-zero), the first timestep is linearized and the simulation ends without ever reaching a steady state.

PR #3156, opened 2026-01-19
Linear Trim Solution: update range calculation when NLinTimes=1

bjonkman:
When linearizing with a trim solution and NLinTimes=1, the range of the reference outputs was always 0.01. Now it uses the channel value. This could probably be modified further, but for now, it gives a slightly better range for the convergence calculation.

Possibly related to https://github.com/OpenFAST/openfast/issues/857, though I don't think this will fix that issue. 

Linearizing with Trim Solution and NLinTimes = 1.

Due to the check in https://github.com/OpenFAST/openfast/pull/1707 that makes NLinTimes at least 2, this fix won't actually do anything in the current code.

This doesn't affect any current regression tests.


deslaughter:
As a note for the future, this won't apply to OpenFAST 5 because it always uses at least two rotor azimuth positions for linearization with CalcSteady (effectively at least NLintimes = 2) and then only save one of them if NLinTimes = 1.

PR #3163, opened 2026-01-21
AA driver: add driver compile to cmake

andrew-platt:
Ready to merge

Added the `aeroacoustics_driver` target to `cmake`.  This will now get built for checking against incidental API changes that might break it.  However, there are no test cases with it at present - these can be added later.

Also, minor correction to `AeroAcoustics_driver_subs.f90`: changed `storage` array to `ReKi` so it will work with the binary writer

#3149 added an aero acoustics driver, but did not include and build process for it.

No test cases exist for this driver at present.

### DIFF
--- a/glue-codes/simulink/CMakeLists.txt
+++ b/glue-codes/simulink/CMakeLists.txt
@@ -19,6 +19,7 @@
 set(MEX_LIBS
   $<TARGET_FILE:openfast_prelib>
   $<TARGET_FILE:basicaerolib>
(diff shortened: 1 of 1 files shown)


deslaughter (approved):

bjonkman:
@andrew-platt, thanks for making those changes. Sorry I didn't add the changes to the build script myself. :) 

FYI, I am intending to change that driver in the future because right now it just runs with constant inputs, which make for constant outputs.

PR #3169, opened 2026-01-23
Update Visual Studio projects in dev-tc

deslaughter:
This PR is ready to merge

This PR updates the Visual Studio project files in `vs-build` so OpenFAST and FAST.Farm build correctly. Another update will be required after merging `dev` into `dev-tc`. This migrates the Fortran compiler to IFX, which has been out for a while and Intel doesn't make it easy to get back to IFORT. It also updates to Visual Studio 2022.

Visual Studio solution and project files in `vs-build`


modules/inflowwind/src/IfW_FlowField.f90
```diff
@@ -72,7 +72,6 @@ subroutine IfW_FlowField_GetVelAcc(FF, IStart, Time, PositionXYZ, VelocityUVW, A
    logical                                   :: GridExceedAllow   ! is this point allowed to exceed bounds of wind grid
 
    ErrStat = ErrID_None
-   ErrMsg = ""
```
andrew-platt, line 75: Really interesting that this was a performance hit.

PR #3210, opened 2026-02-18
Hydroelastic super-element capability for OpenFAST

luwang00:
This PR is not ready to be merged. 

This PR adds hydroelastic super-element capabilities to OpenFAST.

* HydroDyn has been upgraded to support potential-flow bodies with additional generalized DOF beyond the 6 rigid-body modes. Currently, the generalized DOF is limited to `NBody=1` and can only be used with the revamped ExtPtfm module. However, internally, HydroDyn supports generalized DOF for multiple bodies and any of the `NBodyMod` options.

* The ExtPtfm module has been extensively revamped. Support for large rigid-body motion similar to SubDyn was implemented. The ability to specify connection points on the structure allowing coupling to mooring was also added. In addition to prescribed modal forcing, users can also prescribe force time series to be applied at these connection points (currently, only force is allowed, not moment). To support these new capabilities, the input files for ExtPtfm have been redesigned.

* At the glue-code level, coupling between ExtPtfm and HydroDyn and between ExtPtfm and the mooring modules have been added. ExtPtfm is now moved to tight coupling, and the ability to perform linearization with ExtPtfm is also enabled. 

Glue-code, HydroDyn, ExtPtfm

No change to existing test results, except for the only case involving ExtPtfm, `5MW_OC4Jckt_ExtPtfm`. The new results are noticeably different but are still generally in agreement with the old results. The changes in the results might be caused by updating ExtPtfm to use tight coupling. Another change to ExtPtfm is to compute the Jacobians based on central differencing instead of analytically. This might also influence the results. Also, the input files to HydroDyn and ExtPtfm need to be updated with the new inputs.

**To-dos**
- [x] Add check to ensure the number of generalized DOF in HydroDyn matches the number of elastic modes in ExtPtfm.
- [x] Update documentation.

### DIFF
--- a/reg_tests/executeOpenfastRegressionCase.py
+++ b/reg_tests/executeOpenfastRegressionCase.py
@@ -96,28 +96,10 @@ if not os.path.isdir(inputsDirectory):
 
 # create the local output directory if it does not already exist
 # and initialize it with input files for all test cases
-for data in ["AOC", "AWT27", "_DummyTurbineData", "SWRT", "UAE_VI", "WP_Baseline"]:
+for data in ["5MW_Baseline", "AOC", "AWT27", "_DummyTurbineData", "SWRT", "UAE_VI", "WP_Baseline"]:
     dataDir = os.path.join(buildDirectory, data)
     if not os.path.isdir(dataDir):
         rtl.copyTree(os.path.join(moduleDirectory, data), dataDir, excludeExt=excludeExt)
-
-# Special copy for the 5MW_Baseline folder because the Windows python-only workflow may have already created data in the subfolder ServoData
-dst = os.path.join(buildDirectory, "5MW_Baseline")
-src = os.path.join(moduleDirectory, "5MW_Baseline")
-if not os.path.isdir(dst):
-    rtl.copyTree(src, dst, excludeExt=excludeExt)
-else:
-    names = os.listdir(src)
-    for name in names:
-        if name == "ServoData":
-            continue
-        srcname = os.path.join(src, name)
-        dstname = os.path.join(dst, name)
-        if os.path.isdir(srcname):
-            if not os.path.isdir(dstname):
-                rtl.copyTree(srcname, dstname, excludeExt=excludeExt)
-        else:
-            shutil.copy2(srcname, dstname)
 
 if not os.path.isdir(testBuildDirectory):
     rtl.copyTree(inputsDirectory, testBuildDirectory, excludeExt=excludeExt, 


modules/extptfm/src/ExtPtfm_MCKF.f90
```diff
@@ -274,24 +332,53 @@ subroutine ExtPtfm_InitVars(u, p, x, y, m, Vars, InputFileData, Linearize, ErrSt
...
+    call MV_AddMeshVar(Vars%y, "Rigid-body motion mesh", MotionFields, &
+                       DL=DatLoc(ExtPtfm_y_FBMesh), &
+                       Mesh=y%FBMesh)
+
+    if ( p%nConn > 0_IntKi ) then
+       call MV_AddMeshVar(Vars%y, 'Connection point mesh', MotionFields, &
+                         DL=DatLoc(ExtPtfm_y_ConnMesh), &
+                         Mesh=y%ConnMesh)
+    end if
+
```
deslaughter: I recommend using the Active flag for enabling/disabling the variables, but the if statements are fine.

```suggestion
   call MV_AddMeshVar(Vars%y, 'Connection point mesh', MotionFields, &
                      DL=DatLoc(ExtPtfm_y_ConnMesh), &
                      Mesh=y%ConnMesh, &
                      Active=p%nConn > 0_IntKi)
```

modules/extptfm/src/ExtPtfm_MCKF.f90
```diff
@@ -274,24 +332,53 @@ subroutine ExtPtfm_InitVars(u, p, x, y, m, Vars, InputFileData, Linearize, ErrSt
...
 
     call MV_InitVarsJac(Vars, m%Jac, Linearize, ErrStat2, ErrMsg2); if (Failed()) return
 
-    if (Linearize) then
-       call ExtPtfm_CopyContState(x, m%x_perturb, MESH_NEWCOPY, ErrStat2, ErrMsg2); if (Failed()) return
-       call ExtPtfm_CopyContState(x, m%dxdt_lin, MESH_NEWCOPY, ErrStat2, ErrMsg2); if (Failed()) return
-       call ExtPtfm_CopyInput(u, m%u_perturb, MESH_NEWCOPY, ErrStat2, ErrMsg2); if (Failed()) return
-       call ExtPtfm_CopyOutput(y, m%y_lin, MESH_NEWCOPY, ErrStat2, ErrMsg2); if (Failed()) return
-    end if
+    ! if (Linearize) then
```
deslaughter: You can remove these comments. I've been leaning towards always initializing these variables in all modules even when not doing linearization. They don't take up a lot of space and it prevents issues if linearization is added in the future.

modules/hydrodyn/src/HydroDyn.f90
```diff
@@ -1021,6 +1036,29 @@ subroutine HydroDyn_InitVars(Vars, u, p, x, y, m, InitOut, InputFileData, Linear
    call MV_AddMeshVar(Vars%u, "Platform-RefPt", MotionFields, DatLoc(HydroDyn_u_PRPMesh), u%PRPMesh, &
                       Perturbs=Perturbs)
 
+   if ( p%HasAddDOF ) then
```
deslaughter: Maybe use the `Active` argument?

modules/hydrodyn/src/WAMIT_Interp.f90
```diff
@@ -687,5 +688,26 @@ function WAMIT_ForceWaves_Interp_4D_vec6(Time, pos, pKinXX, WF_p, WF_m, ErrStat3
...
+   type(GridInterp_ParameterType),        intent(in   ) :: WF_p               !< wavefield parameters
+   type(GridInterp_MiscVarType),          intent(inout) :: WF_m               !< wavefield misc/optimization variables
+   integer(IntKi),                        intent(  out) :: ErrStat3
+   character(*),                          intent(  out) :: ErrMsg3
+
+   real(SiKi)                             :: WAMIT_ForceWaves_Interp_4D_vecN(N)
+
+   ! get the bounding indices from the WaveField info (same indexing used in WAMIT)
+   call GridInterpSetup4D( (/Real(Time,ReKi),pos(1),pos(2),pos(3)/), WF_p, WF_m, ErrStat3, ErrMsg3 )
+   WAMIT_ForceWaves_Interp_4D_vecN = GridInterp4DVecN( N, pKinXX, WF_m )
```
deslaughter, line 707: It just occurred to me that the `WF_m` will prevent these calls from being parallelized, which could significantly speed up HydroDyn's calculations. It might be worth figuring out how to break out the relevant parts into their own structure that could be returned by `GridInterpSetup*` and then passed into the various `GridInterp` routines. Just a thought, nothing that needs to be done here.

modules/openfast-library/src/FAST_Solver.f90
```diff
@@ -106,21 +106,23 @@ subroutine FAST_SolverInit(p_FAST, p, m, GlueModData, GlueModMaps, Turbine, ErrS
...
+                 pack(modInds, ModIDs == Module_ExtPtfm), &
                  pack(modInds, ModIDs == Module_IfW), &
                  pack(modInds, ModIDs == Module_ExtLd)]
 
    ! Indices of tight coupling modules
    p%iModTC = [pack(modInds, ModIDs == Module_ED .and. p%ModCoupling /= LooseCoupling), &
                pack(modInds, ModIDs == Module_BD .and. p%ModCoupling /= LooseCoupling), &
-               pack(modInds, ModIDs == Module_SD .and. p%ModCoupling /= LooseCoupling)]
+               pack(modInds, ModIDs == Module_SD .and. p%ModCoupling /= LooseCoupling), &
+               pack(modInds, ModIDs == Module_ExtPtfm .and. p%ModCoupling /= LooseCoupling)]
```
deslaughter, line 129: There should probably just be an if statement around this that does `allocate(p%iModTC(0)` if `p%ModCoupling == LooseCoupling` so that we don't have to check each module. For loose coupling, that array should be empty.

modules/extptfm/src/ExtPtfm_MCKF.f90
andrew-platt: I support this change also.  It better matches the rest of the glue code.

luwang00: Done!

modules/hydrodyn/src/HydroDyn.f90
luwang00: Done!

modules/extptfm/src/ExtPtfm_MCKF.f90
luwang00: Done!

andrew-platt:
Verified docs build.  Rendering here: https://ap-openfast.readthedocs.io/en/f-superelement/source/user/extptfm/usage.html

Issue #3189, opened 2026-02-05
Issues with TMD in ServoDyn-Stc


RBergua:
I'm trying to model a wave energy converter (WEC) in OpenFAST as follows:
<img width="318" height="300" alt="Image" src="https://github.com/user-attachments/assets/5eca401e-13b4-458b-8182-818ae4ee80af" />

I tried modeling the hull as the platform in ElastoDyn and the internal reaction mass (IRM) as a TMD in ServoDyn as part of the structural control module (StC).

For reference, the hull represents 80% of the total system mas, and the IRM represents 20% of the mass.

I started doing some tests in still water conditions without mooring lines.

When modeling the system without IRM, I'm obtaining the expected behavior in heave (e.g., in this case the device moves upwards because it is missing the weight of the IRM).
<img width="670" height="453" alt="Image" src="https://github.com/user-attachments/assets/d25adf43-334a-42a9-b80e-6c2c84f51343" />

As part of this verification phase, I also locked the 6 rigid body platform motions in ElastoDyn and looked at the behavior of the IRM. As expected, the ElastoDyn platform stays in place and there is a constant hydrostatic force upwards due to the buoyancy. Looking at he IRM, we can see that it moves 1.63 m downwards due to the gravity acceleration. This behavior is the expected one. For reference, I also include the velocity and the force that the spring of the TMD does in the vertical direction.
<img width="1018" height="817" alt="Image" src="https://github.com/user-attachments/assets/47f8eb52-ba68-44fd-83a0-750c6dd98f40" />

Unfortunately, if I try to combine the platform with heave motion and the TMD, the solver experiences a numerical instability. For reference, I'm using OpenFAST dev-tc (https://github.com/OpenFAST/openfast/tree/dev-tc).
<img width="835" height="255" alt="Image" src="https://github.com/user-attachments/assets/6f780f15-75a5-44d4-a190-ec470c7d4064" />
<img width="1006" height="824" alt="Image" src="https://github.com/user-attachments/assets/e512a295-9a41-4650-844b-45f93bdf6b9d" />

As can be observed, the solver does not converge and I get a heave displacement of 3E12 m. I tried using several correction iterations (`NumCrctn` = 3) and dropping the time step (dt = 1E-3 s). But the results didn't improve.

I understand that this could be a limitation of the loose coupling of ServoDyn within OpenFAST. I tried reducing the IRM mass ratio from 20% to 10%. In this case, I can make it run if I only enable the heave DOF for the platform.
[...]

PR #3213, opened 2026-02-19
Include ServoDyn in Option 1 solve if structural controllers are enabled

deslaughter:
This PR is ready to merge pending testing from @RBergua. 

This PR changes the glue code to add ServoDyn to the Option 1 implicit input-output solve for modules whose forces and accelerations significantly affect the structural dynamics of the model if any structural controllers are enabled in the module. As part of this improvement, the original ServoDyn Jacobian calculation routines were replaced with central difference routines to match the other modules which were revised during tight-coupling development. This change highlighted a couple of differences in the Jacobians, related to structural controllers, and several test baselines were updated as a result.

While processing this PR it was found that the Simulink build was no longer working for the Visual Studio projects, so they were modified to address this separate issue.

This feature was added as a result of discussion in #3189 where modeling a Wave Energy Converter failed because the model would quickly go unstable.

- `ServoDyn.f90`
- `FAST_Solver.f90`
- Visual Studio Projects


RBergua (approved):
I confirm that ServoDyn is working as expected based on the TMD tests that I was performing.

Thanks, @deslaughter!

Issue #3216, opened 2026-02-23
Linear vs nonlinear hydrostatics in a hybrid model (potential-flow + viscous drag)


RBergua:
As part of the OC7 Phase II project, I am analyzing a floating structure (VolturnUS-S design) as follows:
<img width="383" height="378" alt="Image" src="https://github.com/user-attachments/assets/76e3bad4-2ff2-40c4-8b4d-f3759c77284d" />

The incident waves travel along the x-direction. As part of the project, amongst other things, we are looking at the torque loads in the downstream pontoon 2. This is the torque along the pontoon longitudinal axis (_z_p2_)

I'm using 3 numerical models in OpenFAST:
NLR1: strip-theory model using rectangular elements,
NLR2: potential-flow model.
NLR3: strip-theory model using virtual cylinders as a work around when rectangular elements are not available.

The 3 models use the same nonlinear hydrostatics formulation.

Below you can see the torque observed in the pontoon 2 during a regular wave condition. The window of time corresponds to 3 wave periods approximately.
<img width="599" height="753" alt="Image" src="https://github.com/user-attachments/assets/0bf8edec-980f-417b-9c39-e5f103ae73fb" />

As observed, the NLR1 and NLR3 results are more or less consistent with other participants in the project using strip-theory models. However, the NLR2 model behaves different compared to other potential-flow models. A phase shift and a hump in the lower part of the torque signal can be observed.

The differences in terms of amplitudes between the strip-theory and the potential-flow models are due to the physics included in these models and will be covered in a publication.

In any case, something seems to be off with the NLR2 model. And such behavior was observed regadless of using the displaced or undisplaced platform position.

I performed a sensitity analysis to try to better understand what was going on:
<img width="644" height="521" alt="Image" src="https://github.com/user-attachments/assets/9e26ab50-3189-4ebf-b071-ac407b54e71e" />

The original results shown in the comparison against other participants corresponds to the vertical wave stretching with nonlinear hydrostatics (shown in this plot in blue). When not using the wave stretching capability (red solution) or when using wave stretching with linear hydostatics (yellow solution), the results are quite different and aligned with other participants in the project.

It seems that the **nonlinear hydrostatics may be incosistent with potential-flow hydrodynamic loads based on linear or second-order theory**.

[...]

PR #3219, opened 2026-02-24
Added flexibility for strip-theory hydrostatic load calculation

luwang00:

This PR adds a new input switch `HstMod` under `Strip Theory Options` to HydroDyn. `HstMod` controls the computation of distributed hydrostatic loads on strip-theory members. If `HstMod = 0`, the hydrostatic pressure is always integrated on the instantaneous wetted surface up to the SWL, even if wave stretching is enabled. If `HstMod = 1` and one of the wave stretching model is selected with `WaveStMod > 0`, the hydrostatic pressure will be integrated up to the instantaneous wave free surface, considering both wave elevation and wave slope. If wave stretching is not used, `HstMod` will always be set to zero internally, disregarding the user input. Irrespective of `HstMod`, the exact displaced position of the structure is always used when computing hydrostatic loads to obtain the correct hydrostatic restoring effect.

Generally, when strip theory is used for both hydrodynamic and hydrostatic loads, wave stretching with `HstMod = 1` can be used. However, for hydroelastic simulations, a mixed approach is sometimes preferred. With this approach, the hydrodynamic loads are computed using multiple potential-flow bodies, while the hydrostatic loads are computed using the nonlinear strip-theory approach (with `PotMod = false` but all hydrodynamic coefficients zeroed out). This approach provides the correct mean loads in the structure from hydrostatic effects that cannot be easily obtained with displaced volumes and hydrostatic stiffness matrices. In this case, inconsistencies can arise between the hydrodynamic loads based on perturbation theory and the fully nonlinear hydrostatic loads up to the instantaneous free surface when simulating large incident waves (see Issue #3216). To address this problem, `HstMod = 0` can be used, allowing wave stretching to be used when computing the distributed drag force only.

Issue #3216

HydroDyn, openfast_io

No change to existing test results. All HydroDyn input files, openfast_io, and user documentation have been updated with the new `HstMod` input.


andrew-platt (approved):

jjonkman (approved):
Thanks for adding this feature!

PR #3221, opened 2026-02-26
Enhancements to the structural control submodule of ServoDyn

luwang00:
This PR is ready to merge. Need to update r-test first.

This PR contains two enhancements to the structural control submodule of ServoDyn:

* A new option `StC_DOF_MODE=3` has been added. This option provides a 3DOF omni-directional TMD to supplement the existing 2DOF (XY) omni-directional TMD. To make room for the new option, existing options of `StC_DOF_MODE` have been renumbered to: 
0 - no StC;
1 - three independent StC DOF;
2 - 2DOF XY omni-directional StC;
3 - 3DOF XYZ omni-directional StC;
4 - reserved and currently unused;
5 - TLCD;
6 - prescribed force/moment time series;
7 - force determined by external DLL.

* A new option `StC_CMODE=3` has been added. This option enables active control through a user-defined subroutine. A template subroutine `UserStC` has been added to `modules/servodyn/src/UserSubs.f90`.

* ServoDyn StC user docs, openfast_io, and the comments in the relevant r-test input files have all been updated to be consistent with the new input options.

ServoDyn, openfast_io, docs

The StC reaction moment with `StC_DOF_MODE=2` has been updated for robustness, leading to negligible changes in the test results of `StC_test_OC4Semi`.
- [x] Requires merging r-test


jjonkman (commented):
Overall, this change makes sense to me.

Would it make more sense to renumber the `StC_DOF_MODE` options, assigning the 3DOF omni-directional TMD to 3, and shifting the other up by 1 each?

The readthedocs documentation should also be updated accordingly.

luwang00:
@jjonkman Thanks for reviewing the code changes. I agree with the suggestions. I'll move the 3DOF option to 3, but I'd like to shift the rest up by 2, leaving 4 open in case we want to implement a 6DOF option in the future. I'll also get the docs updated.

@andrew-platt I think I can get all this done today, including r-test. If so, maybe we can move this up to 5.0? I think the features are quite useful to have.

RBergua (approved):
Thank you so much for putting this together, @luwang00! This is very powerful!

I add some comments below to provide some context. This notes will be useful for me and may be also benefitial for others down the road.

The first enhancement, `StC_DOF_MODE` = 3, is useful to include a TMD in the structural control (StC) in ServoDyn that can move in the 3 translational directions (omni-directional). Note that till now, StC allowed to introduce up to 3 **independent** TMD DOFs by means of `StC_DOF_MODE` = 1 or a 2DOF omni-directional TMD in the horizontal XY plane by means of `StC_DOF_MODE` = 2. Now with `StC_DOF_MODE` = 3, we have the option to have one mass that moves in the x, y, z directions (3DOF omni-directional TMD). This overcomes the limitation that we were commenting here: https://github.com/OpenFAST/openfast/issues/3189#issuecomment-3934683849

The second enhacemenet, `StC_CMODE` = 3, is a user-defined subroutine. When defining for example a Wave Energy Converter (https://github.com/OpenFAST/openfast/issues/3189#issue-3902067649), we wanted to have one mass that moved in 3 translational directions. Thanks to this PR, we can do this now by selecting `StC_DOF_MODE` = 3. But in the real system, the stiffness and damping of the internal reaction ball mechanism is not linear (there is a nonlinearity coming from the geometry when the ball moves in the space). What we do in this user-defined subroutine is to pass the displacement vector (x, y, z) and velocity vector (x, y, z) for this mass to compute internally the corresponding forces (x, y, z). Note that this displacement, velocity, and force vector are expressed in the local coordinate system of ServoDyn (e.g., if the hull in ElastoDyn rotates, this local coordinate system follows those rotations). Inside the subroutine (empty template in this PR) we hard-coded the attachment points at the hull side and at the internal reaction ball side and placed 6 spring-dashpot elements (Power Take-Off [PTO] devices). At every time step, we have a given displacement and velocity of the mass that we pass to the subroutine to compute the corresponding elastic force (Kx) and damping force (Cv) for each PTO. Then we sum the 6 contributions and return the total force. This allows us to include a fully nonlinear TMD in the 3 translational directions within ServoDyn.

To use this subroutine, we modified the template UserStC at the end of modules/servodyn/src/UserSubs.f90 and compiled OpenFAST.

For reference, when studying the Wave Energy Converter commented above, the agreement between WEC-Sim (using multibody elements for the PTOs) and OpenFAST (using ServoDyn) is very good.
image

The above results were obtained compiling OpenFAST with GCC version 15.2.0 with 64-bit architecture and double precision. Equivalent results can also be obtained with GCC using single precision or using Intel(R) Fortran Compiler 20250302 in double precision. However, using Intel(R) Fortran Compiler 20250302 in single precision returns slightly different results:
image

andrew-platt (approved):

PR #3231, opened 2026-03-03
SeaState: Small change in WaveTp logic

RBergua:

In SeaState, the peak-spectral period parameter (`WaveTp`) should only play a role when using `WaveMod` = 1 or 2 (regular or irregular waves).
image

However, nowadays, there is a check regardless of the wave option selected. For example, when using `WaveMod` = 5 (externally generated wave-elevation time series), I get:
image

This modification checks for a valid `WaveTp` only if `WaveMod` = 1 or 2.

### DIFF
--- a/modules/seastate/src/SeaState_Input.f90
+++ b/modules/seastate/src/SeaState_Input.f90
@@ -697,11 +697,16 @@ subroutine SeaStateInput_ProcessInitData( InitInp, p, InputFileData, ErrStat, Er
 
 
       ! WaveTp - Peak spectral period.
-   if ( InputFileData%Waves%WaveTp <= 0.0 )  then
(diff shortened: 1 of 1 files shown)


andrew-platt:
Does this apply to 4.2 as well?

RBergua:
Yes, it applies to any SeaState. This condition should have been always there.

andrew-platt:
Backport to 4.2.1

luwang00:
@RBergua Thanks for fixing this. It looks good to me. Maybe change to `0.0_ReKi` as @deslaughter suggested before.

RBergua:
Good point! I just updated it to `0.0_SiKi` to be aligned with the Registry. Thanks, @luwang00!

andrew-platt (approved):

luwang00 (approved):

PR #3233, opened 2026-03-04
NWTC Library: Adjust subroutines for converting between DCM and small angles

bjonkman:
Adjusted subroutines for computing small angles from DCM and the inverse routines to compute DCMs from small angles. These equations are accurate for larger angles.

- [ ] r-test branch merging required?


modules/nwtc-library/src/NWTC_Num.f90
```diff
@@ -2688,45 +2688,58 @@ FUNCTION GetSmllRotAngsD ( DCMat, ErrStat, ErrMsg )
...
          IF (PRESENT(ErrMsg)) THEN
-            ErrMsg = ' Angles in GetSmllRotAngs() are larger than '//TRIM(Num2LStr(LrgAngle))//' radians.'
+            ErrMsg = ' Denominator is zero in GetSmllRotAngs().'
          ELSE
-            CALL ProgWarn( ' Angles in GetSmllRotAngs() are larger than '//TRIM(Num2LStr(LrgAngle))//' radians.' )
+            CALL ProgAbort( ' Denominator is zero in GetSmllRotAngs().', TrapErrors = .TRUE. )
          END IF
+      
+      ELSE
+         GetSmllRotAngsD = GetSmllRotAngsD / denom
```
jjonkman, line 2728: Does this need this need to be?

GetSmllRotAngsD = GetSmllRotAngsD * 0.5_DbKi / denom

modules/nwtc-library/src/NWTC_Num.f90
```diff
@@ -2753,41 +2766,56 @@ FUNCTION GetSmllRotAngsR ( DCMat, ErrStat, ErrMsg )
...
          IF (PRESENT(ErrMsg)) THEN
-            ErrMsg = ' Angles in GetSmllRotAngs() are larger than '//TRIM(Num2LStr(LrgAngle))//' radians.'
+            ErrMsg = ' Denominator is zero in GetSmllRotAngs().'
          ELSE
-            CALL ProgWarn( ' Angles in GetSmllRotAngs() are larger than '//TRIM(Num2LStr(LrgAngle))//' radians.' )
+            CALL ProgAbort( ' Denominator is zero in GetSmllRotAngs().', TrapErrors = .TRUE. )
          END IF
+      
+      ELSE
+         GetSmllRotAngsR = GetSmllRotAngsR / denom
```
jjonkman, line 2803: Does this need this need to be?

GetSmllRotAngsR = GetSmllRotAngsR * 0.5_SiKi / denom

jjonkman (commented):
This matches what I would expect, with one change.

jjonkman (approved):
Never mind.  I see from the plan that the 2's cancel.  This implementation looks correct.  Thanks!

bjonkman:
I looped through different angles, converting them to DCMs and then calling this code to get the "small" angles from of the DCM. I then plotted the difference between the starting angle and the calculated angles. As you can see, the error is small (taking into account for all 3 dimensions, the error was no larger than 1.1e-13 degrees).
image

For reference, this is the plot with the original algorithm that we have now replaced:
image

andrew-platt:
I'll look review the regression test changes and update them in the next few days for merging.  Thanks @bjonkman!

deslaughter (approved):

deslaughter:
I think this is ready to merge. The ATLAS blas seemed to give clean regression tests without having to rerun any. I say we try it and see.

PR #3240, opened 2026-03-08
UA: adding quasi steady outputs to UA driver outputs

ebranlard:

Adding quasi-steady airfoil coefficients (Cl, Cd, Cm) to UA DEBUG outputs at the quarter point and three quarter point.
These outputs will show up for the UA driver or if the user ask for debug UA outputs during compilation of OpenFAST.

AeroDyn / UA / UA Driver

Test shouldn't be affected


modules/aerodyn/src/UnsteadyAero.f90
```diff
@@ -1423,10 +1429,19 @@ subroutine UA_Init_Outputs(InitInp, p, y, InitOut, errStat, errMsg)
             InitOut%WriteOutputUnt(iOffset+43) ='(-)'
             InitOut%WriteOutputUnt(iOffset+44) ='(-)'
             InitOut%WriteOutputUnt(iOffset+45)  ='(deg)'
+            iOffAcc = iOffset+45
 
          else
             call SetErrStat( ErrID_Fatal, 'Programming error UAmod case not accounted for.', ErrStat, ErrMsg, RoutineName ); return
          end if
+
```
bjonkman, line 1437: Would it be easier if you just set `iOffAcc = iOffset + p%NumOuts - 6` here? Instead of setting it separately in each UAMod case above?

andrew-platt: It might be worth modifying the logic and math a bit in the future (case statements perhaps).  Given the time constraints on 5.0.0, I'm inclined to leave it as is for now since we haven't heard back in a bit.

Issue #3293, opened 2026-03-31
SoilDyn issues in OpenFAST v5.0.0


RBergua:
I am woking with the SoilDyn module and the REDWIN DLL. I'm using the recently released OpenFAST v5.0.0 and comparing the results against OpenFAST v2.x that was used in the OC6 Phase II project (https://docs.nlr.gov/docs/fy21osti/79938.pdf).

For reference, the OpenFAST model accounts for a rigid RNA and a flexible tower (in ElastoDyn), a flexible monopile (in SubDyn) and soil-structure interaction (in SoilDyn).

I'm running a case with irregular waves and imposing a time serie of Fx, Fy, Fz, Mx, My, Mz loads at the nacelle through the structural control.

When comparing both numerical models, I can see that most results at the tower base location are very close between versions (as expected). See below for reference:
<img width="1418" height="756" alt="Image" src="https://github.com/user-attachments/assets/98c7a382-e0f3-4ee8-a5b4-2beb83176d2f" />

However, when I look at the seabed location [-30 m] (SoilDyn sensor), I can see **significant differences between both numerical models**:
<img width="1427" height="753" alt="Image" src="https://github.com/user-attachments/assets/b59b22b5-a88d-4b1e-992d-15965b354f34" />

Based on the data at the tower base location, I can perform some simple verifications. The first one, the force in the x direction at the seabed should be aligned to the tower base one, but containing more dynamics. For example, the seabed will be influenced by the waves loading the monopile. Basically, I would expect the same moving average between Fx at the tower base and seabed and different dynamics. In addition, the moving average should be pretty well aligned with the Fx that I prescribe at the nacelle.

Looking at OpenFAST v2.x, I can see the expected  behavior:
<img width="999" height="633" alt="Image" src="https://github.com/user-attachments/assets/998eb82f-d2f7-4d0e-9169-f36bdbf9de19" />

However, OpenFAST v5.0.0 shows an unexpected behavior with the moving average being off.
<img width="994" height="631" alt="Image" src="https://github.com/user-attachments/assets/27ba7209-727a-42e8-8630-383f2e046ca3" />

It's also quite interesting to compare the Fz loading at the seabed since we saw an offset in the mean value in the above plot comparing OpenFAST v2.x vs OpenFAST v5.0.0. I extracted the loading from SubDyn and SoilDyn for the Z direction.

In OpenFAST v2.x these two outputs are aligned:
[...]

PR #3297, opened 2026-04-02
Fix SubDyn reaction load output

luwang00:

Previously, the reaction loads from SubDyn is missing the self-weight of the reaction joint(s). This contribution is now added back. The cable loads on the reaction joint(s) are also added just in case.

Issue https://github.com/OpenFAST/openfast/issues/3293

SubDyn

Likely need to update some of the r-tests involving SubDyn


jjonkman:
@luwang00 -- Thanks for fixing this.  You mention "added back".  I thought this term was previously already included in SubDyn.  Was it removed at some point (accidentally?) and now added back?

luwang00:
@jjonkman Sorry, I wasn't that careful with the wording. I'm not sure when this bug was introduced. It might be that the self-weight was part of `u%LMesh` in the past but subsequently got separated out into `p%FG` at some point, but the output code is not updated to match.

andrew-platt (approved):

PR #3304, opened 2026-04-10
SubDyn: Y3 mesh

RBergua:
I believe this PR is ready to be merged. 

r-tests using SoilDyn are expected to return different results.

In PR https://github.com/OpenFAST/openfast/pull/1526, the static improvement method (SIM) was removed from the elastic displacement output in SubDyn's Y3 mesh. The rationale was that SIM-corrected displacements were not consistent with the elastic velocities and accelerations, which do not account for SIM contributions from truncated higher modes. This change was introduced to avoid issues with MoorDyn in floating systems.

However, this change introduced an inconsistency for fixed-bottom systems: the full elastic displacement at the reaction node in SubDyn (Guyan + CB + SIM) differs from the displacement passed to SoilDyn via the Y3 mesh (Guyan + CB only, no SIM).

To overcome this limitation, the Y3 mesh for fixed-bottom systems has been modified. The proposed code **modifies only the displacements of the reaction node(s) to include the full elastic displacements (Guyan+CB+SIM)**.  All other nodes in the Y3 mesh retain their existing Guyan + CB displacements, preserving consistency between displacements and their time derivatives (velocities and accelerations). This should also ensure correctness for MoorDyn connections to non-reaction nodes in fixed-bottom systems. The only unaddressed edge case is a MoorDyn connection directly to a SubDyn reaction node, which is not an expected configuration.

Extensive documentation and testing can be found here: https://github.com/OpenFAST/openfast/issues/3293

SubDyn Y3 mesh output for fixed-bottom systems. I believe this mesh is currently used by SoilDyn and MoorDyn.

### DIFF
--- a/modules/subdyn/src/SubDyn.f90
+++ b/modules/subdyn/src/SubDyn.f90
@@ -887,7 +887,7 @@ SUBROUTINE SD_CalcOutput( t, u, p, x, xd, z, OtherState, y, m, ErrStat, ErrMsg )
             ! Construct the direction cosine matrix given the output angles
             CALL SmllRotTrans( 'UR_bar input angles', m%U_full_NS(DOFList(4)), m%U_full_NS(DOFList(5)), m%U_full_NS(DOFList(6)), DCM, '', ErrStat2, ErrMsg2); if(Failed()) return
             y%Y2mesh%Orientation     (:,:,iSDNode)   = DCM
-            y%Y2mesh%TranslationDisp (:,iSDNode)     = m%U_full_NS     (DOFList(1:3)) !Y2: Guyan+CB (but no SIM) displacements
+            y%Y2mesh%TranslationDisp (:,iSDNode)     = m%U_full        (DOFList(1:3)) !Y2: Guyan+CB (with SIM) displacements
             y%Y2mesh%TranslationVel  (:,iSDNode)     = m%U_full_dot    (DOFList(1:3))
             y%Y2mesh%TranslationAcc  (:,iSDNode)     = m%U_full_dotdot (DOFList(1:3))
             y%Y2mesh%RotationVel     (:,iSDNode)     = m%U_full_dot    (DOFList(4:6))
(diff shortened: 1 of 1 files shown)


modules/subdyn/src/SubDyn.f90
```diff
@@ -892,10 +894,18 @@ SUBROUTINE SD_CalcOutput( t, u, p, x, xd, z, OtherState, y, m, ErrStat, ErrMsg )
             y%Y2mesh%TranslationAcc  (:,iSDNode)     = m%U_full_dotdot (DOFList(1:3))
             y%Y2mesh%RotationVel     (:,iSDNode)     = m%U_full_dot    (DOFList(4:6))
             y%Y2mesh%RotationAcc     (:,iSDNode)     = m%U_full_dotdot (DOFList(4:6))
+            if (any(p%Nodes_C(:,1) == iSDNode)) then
```
luwang00: For performance, I think it would be better to move the `if` statement out of this loop, and simply iterate through the reaction nodes only in a new loop after the `Y2` to `Y3` displacement copy. So restore the codes up to the `Y2` to `Y3` copy and add after that something like
```
do i = 1, p%nNodes_C ! loop on constrained nodes
   iSDNode = p%Nodes_C(i,1)
   associate(DOFList => p%NodesDOF(iSDNode)%List)  ! Alias to shorten notations
   ! Reaction node: Y3 gets Guyan+CB+SIM displacements (for SoilDyn)
   CALL SmllRotTrans( 'UR_bar input angles', m%U_full(DOFList(4)), m%U_full(DOFList(5)), m%U_full(DOFList(6)), DCM, '', ErrStat2, ErrMsg2); if(Failed()) return
   y%Y3mesh%Orientation     (:,:,iSDNode)   = DCM
   y%Y3mesh%TranslationDisp (:,iSDNode)     = m%U_full        (DOFList(1:3)) ! Y3: Guyan+CB+SIM displacements
end do
```
This also avoids the `any` check, and all the new codes are confined in one block.

luwang00 (changes requested):

modules/subdyn/src/SubDyn.f90
RBergua: That's a good point, @luwang00. I didn't notice any computational time penalty. But definitely your approach is cleaner!

andrew-platt: I like @luwang00's proposed modification.

luwang00 (approved):

PR #3310, opened 2026-04-16
SubDyn: JSON output file with local x_e axis for circular beams

RBergua:

The proposed code modifies the JSON output file from SubDyn. Now, the local _x_e_ local axis for circular beams is also written down in the output file (similar to what we do with the rectangular beams using _SideA_dir_). This allows to visualize the local coordinate system in viz3Danim (see for reference: https://github.com/ebranlard/viz3Danim/pull/9).
image

With this modification, the user can orientate the local _x_e_ and _y_e_ axes for circular beams using the spin angle (`MSpin`) in SubDyn. Till now, we were computing the default _x_e_ and _y_e_ axes in circular beams within viz3Danim without accounting for the spin angle because we didn't have visibility about it. Now, this information is available for viz3Danim and it can be parsed.

https://github.com/ebranlard/viz3Danim/issues/10

JSON output file from SubDyn.


andrew-platt (approved):
Thanks for fixing this!

andrew-platt:
Adding this to 5.0.1 (`rc-5.0.1` branch)

Issue #3314, opened 2026-04-19
Bug report Question about the multi-rotor case


TianyuanWangi:
**Bug description**
<A clear and concise description of the bug.>
Hi, 
I encountered some confusion when studying the multi-rotor case in OpenFAST v5.0.0, specifically regarding the coupling reference point used by MoorDyn.

For a single turbine case, the coupling reference point is fully determined by the ElastoDyn RPR. However, in the multi-rotor case, I found the following in `SUBROUTINE FAST_InitializeAll`:

```
p_FAST%PlatformPosInit = Init%OutData_ED(iRot)%PlatformPos
Init%InData_MD%PtfmInit(:,1)  = p_FAST%PlatformPosInit ! initial position of the platform (when a FAST module, MoorDyn just takes one row in this matrix)
```
This seems to indicate that the MoorDyn coupling point is defined using the RPR from one rotor (depending on iRot, possibly the second rotor).

When I looked into `SUBROUTINE MD_Init`, I found:
```
DO l = 1,p%nCpldPoints(iTurb)   ! keeping this one simple for now, positioning at whatever is specified by glue code <<<
   J = J + 1
   
   ! set reference position as per input file  <<< what about turbine positions in array?
   rRef(1:3) = m%PointList(m%CpldPointIs(l,iTurb))%r                           

   ! set absolute initial positions in MoorDyn
   CALL MeshPositionNode(u%CoupledKinematics(iTurb), J, rRef(1:3), ErrStat2, ErrMsg2)  

   ! calculate initial point relative position, adjusted due to initial platform rotations and translations
   u%CoupledKinematics(iTurb)%TranslationDisp(1,J) = InitInp%PtfmInit(1,iTurb) + OrMat(1,1)*rRef(1) + OrMat(2,1)*rRef(2) + OrMat(3,1)*rRef(3) - rRef(1)
   u%CoupledKinematics(iTurb)%TranslationDisp(2,J) = InitInp%PtfmInit(2,iTurb) + OrMat(1,2)*rRef(1) + OrMat(2,2)*rRef(2) + OrMat(3,2)*rRef(3) - rRef(2)
   u%CoupledKinematics(iTurb)%TranslationDisp(3,J) = InitInp%PtfmInit(3,iTurb) + OrMat(1,3)*rRef(1) + OrMat(2,3)*rRef(2) + OrMat(3,3)*rRef(3) - rRef(3)   

[...]

PR #3316, opened 2026-04-22
Fix MoorDyn initial fairlead positions when SubDyn is present

luwang00:

Previously, the initial offsets of MoorDyn fairleads during dynamic relaxation were set based on ElastoDyn even if SubDyn is present. With this PR, the fairlead initial positions will be set based on SubDyn initial offset when appropriate.

Issue #3314 

Glue-code

No change to r-test

### DIFF
--- a/modules/openfast-library/src/FAST_Subs.f90
+++ b/modules/openfast-library/src/FAST_Subs.f90
@@ -373,7 +373,7 @@ SUBROUTINE FAST_InitializeAll( t_initial, m_Glue, p_FAST, y_FAST, m_FAST, ED, SE
          ! Save the GearBox index
          p_FAST%GearBox_index = Init%OutData_ED(iRot)%GearBox_index
 
-         ! Assign the inital positions for use by MoorDyn initalization
+         ! Assign the initial positions for use by MoorDyn initialization
          p_FAST%PlatformPosInit = Init%OutData_ED(iRot)%PlatformPos
       
          ! If steady state calculation is enabled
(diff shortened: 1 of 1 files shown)


modules/openfast-library/src/FAST_Subs.f90
```diff
@@ -992,6 +992,15 @@ SUBROUTINE FAST_InitializeAll( t_initial, m_Glue, p_FAST, y_FAST, m_FAST, ED, SE
                         Init%OutData_SD%Vars, p_FAST%Linearize, ErrStat2, ErrMsg2)
       if (Failed()) return
 
+      ! Assign the initial positions for use by MoorDyn initialization (this overwrites the PlatformPosInit from ED above)
+      if (Init%OutData_SD%SDHasRBDoF) then
+         p_FAST%PlatformPosInit = Init%OutData_SD%PlatformPos
+      else if (.not.Init%OutData_SD%IsFloating) then
+         p_FAST%PlatformPosInit = 0.0_DbKi
```
andrew-platt, line 999: does SD not have the initial platform surge from ED?

andrew-platt (approved):

modules/openfast-library/src/FAST_Subs.f90
luwang00: For fixed-bottom structures, it makes the most sense to just zero any initial offset. Also, mooring is unlikely to be involved in this case.

Issue #3318, opened 2026-04-27
SoilDyn: Misleading console message


RBergua:
**Bug description**
SoilDyn includes two available `CalcOptions` at the moment:
`Calculation option  {1: Stiffness / Damping matrices, 2: P-Y curves [unavailable], 3: coupled REDWIN DLL}`

Currently, as soon as SoilDyn is being called, a message in the console is written down:
`SoilDyn returning only non-linear portion of reaction forces` 

However, this may be _slightly_ misleading for the users because non-linear forces are only computed and send from SoilDyn to SubDyn if REDWIN (`CalcOption` = 3) is used. 

`CalcOption`: 1 uses a linear stiffness matrix. This is equivalent to use SSI in SubDyn directly.
<img width="733" height="346" alt="Image" src="https://github.com/user-attachments/assets/8376a32a-82d2-4f9a-ba70-200393d88b3b" />

`CalcOption`: 3 uses a linear stiffness matrix AND non-linear forces computed by REDWIN are applied to SubDyn.
<img width="835" height="404" alt="Image" src="https://github.com/user-attachments/assets/c181c2f2-d57a-4428-9c7d-f2ed394bc48f" />

I believe it would be better to display the message about non-linear forces only when REDWIN is used. I'm fixing some bugs and will address this message there as well. I will open a PR soon.

jjonkman:
Perhaps change the message to state that the linear portion of the stiffness is sent directly to SubDyn as a stiffness matrix, rather than as a load, and other terms are transferred as loads.  Such a message would apply to all `CalcOptions` of SoilDyn and it is the stiffness matrix transfer that that is "special" compared to other module couplings within OpenFAST.

Best regards,

Issue #3317, opened 2026-04-24
SubDyn: Wrong reaction loads when SoilDyn includes nonlinear loads


RBergua:
**Bug description**
When modeling soil-structure interaction (SSI) through a linear stiffness matrix in SubDyn or SoilDyn (i.e., `CalcOption` = 1), the results are the expected ones and SoilDyn and SubDyn return consistent results. See for reference:
<img width="793" height="520" alt="Image" src="https://github.com/user-attachments/assets/fab83c9b-b7fd-4506-a681-d9efe5aef999" />

However, once the user includes a nonlinear SSI (i.e., SoilDyn `CalcOption` = 3), the reaction loads provided by SubDyn (e.g., `ReactFXss`) are not consistent. See for reference:
<img width="762" height="554" alt="Image" src="https://github.com/user-attachments/assets/395b2c8f-737a-45b5-990e-fd1aa72c7241" />

I was wondering what these reaction loads from SubDyn were actually representing. After doing some numbers, I can confirm that the reaction loads from SubDyn account for the linear part of the SSI and disregards the nonlinear part. Since the (linear) stiffness matrix is an input that I provide and SubDyn returns the displacements and rotations at the seabed, I'm able to compute analytically the loading that comes from the SSI linear part. For example:
Stiffness matrix:
6.336198E9   0.0000e+00   0.0000e+00   0.0000e+00  -5.015421E10  0.0000e+00
   0.0000e+00   6.336198E9   0.0000e+00   5.015421E10  0.0000e+00   0.0000e+00
   0.0000e+00   0.0000e+00   1.119691E10  0.0000e+00   0.0000e+00   0.0000e+00
   0.0000e+00   5.015421E10  0.0000e+00   8.111942E11  0.0000e+00   0.0000e+00
  -5.015421E10  0.0000e+00   0.0000e+00   0.0000e+00   8.111942E11  0.0000e+00
   0.0000e+00   0.0000e+00   0.0000e+00   0.0000e+00   0.0000e+00   2.552673E11

Linear force at the seabed computed analytically:
{M1N1TDxss}·6.336198E9 + {M1N1RDye}·(-5.015421E10)

By doing this, I can see an excellent agreement between the linear part of the SSI and the reaction load from SubDyn:
<img width="730" height="588" alt="Image" src="https://github.com/user-attachments/assets/fc64dd43-c22b-453a-b9f6-a2bb0b2bb376" />

[...]

Issue #3301, opened 2026-04-10
SoilDyn: Compiler issues


RBergua:
**Bug description**
For some reason that I don't really understand, SoilDyn seems to have issues depending on the compiler used. 

For example, when compiling OpenFAST using GCC version 15.2.0 in 64-bit (with or without OpenMP), the program aborts when executing SoilDyn. No errors or exceptions are received. OpenFAST just stops. For reference, this is using a stiffness matrix in SoilDyn (`calcOption = 1`). There is no REDWIN DLL involved.
<img width="1071" height="624" alt="Image" src="https://github.com/user-attachments/assets/13c2f458-9c2f-4a70-9dc8-576e7db2034f" />

When compiling using Intel(R) Fortran Compiler 20250302 in 64-bit, everything works as expected for the same case:
<img width="939" height="1105" alt="Image" src="https://github.com/user-attachments/assets/63518fbc-d3a0-4825-bbf9-17f43033641f" />

andrew-platt:
This is peculiar.  I suspect the issue is some memory that is either not allocated correctly, or an uninitialized index that IFX initializes to zero, but GCC does not.

PR #3319, opened 2026-04-27
SoilDyn/SubDyn: reaction loads when REDWIN is used

RBergua:

The proposed code introduces a new flag (`SlDNonLinear`) to pass from SoilDyn to SubDyn in order to correctly identify when SoilDyn is computing nonlinear soil reaction forces (REDWIN DLL, `CalcOption`=3).

The new flag `SlDNonLinear` is set internally in SoilDyn based on `CalcOption`, ensuring it is only TRUE when the REDWIN DLL is active. When nonlinear forces are active, SubDyn reaction outputs (`ReactSS`) are set to NaN to prevent the user from using reaction loads that only have the linear part. Users should instead use the SoilDyn output sensors (e.g., `Sld1Fxg`, `Sld1Fyg`) which correctly account for the total soil reaction at each node.

Additionally, the linear soil stiffness transfer from SoilDyn to SubDyn is now better documented in the console (`SoilDyn: linear stiffness passed directly to SubDyn as a stiffness matrix`).

For reference, in the current code before these changes, `SlDNonLinearForcePortionOnly` seems to be defined as TRUE whenever SoilDyn is called, regardless of whether nonlinear forces are actually being computed. This seems a bit confusing. See at the bottom the "Future improvements?" part. 

This PR addresses issues https://github.com/OpenFAST/openfast/issues/3317 and https://github.com/OpenFAST/openfast/issues/3318

SubDyn reaction load sensor when REDWIN (`CalcOption` = 3 in SoilDyn) is used.
SoilDyn messages in the console at run time.

This PR impacts the SubDyn outputs `ReactFXss`, `ReactFYss`, `ReactFZss`, `ReactMXss`, `ReactMYss`, `ReactMZss`. However, there are no r-test cases with REDWIN. So, no differences are expected.

For verification purposes, you can find examples of the new messages and results below:
`CalcOption` = 1 uses a linear stiffness matrix.
image

In this case all sensors (SubDyn and SoilDyn) are available:
image

`CalcOption`: 3 uses a linear stiffness matrix AND non-linear forces computed by REDWIN.
image

In this case, the reaction loads from SubDyn return NaN. The proper results (including the linear and nonlinear part) are available in SoilDyn.
image

**Additional information: some drama compiling OpenFAST...**
I tried to compile OpenFAST in Visual Studio Code by means of GCC. Although I could compile without issues, when I executed the code I ended up with the issues reported here: https://github.com/OpenFAST/openfast/issues/3301

Then I decided to move to GitHub Actions and compile with Intel Fortran Compiler. @andrew-platt Somehow, when running the `deploy.yml`, GitHub was not generating from scratch the `SubDyn_Types.f90` file based on my updated `SubDyn_Registry.txt`. I'm not sure if this is a limitation of the `deploy.yml` approach. To overcome this, I took the updated `SubDyn_Types.f90` from my local run in Visual Studio Code and added it to my branch: 67db890217cd67fbb595e7feac9119d9e697ea03.

**Future improvements?**
When looking at the `Calc_StiffDamp` and `Calc_REDWIN` in the SoilDyn.f90 file, the linear and nonlinear loads are computed as expected. However, we can find the condition: `if (p%SlDNonLinearForcePortionOnly) then ` in several places. I do believe this is indeed always TRUE and could be removed. But I'm not sure if there was a reason for it during the development. 
image

### DIFF
--- a/modules/openfast-library/src/FAST_Subs.f90
+++ b/modules/openfast-library/src/FAST_Subs.f90
@@ -907,6 +907,9 @@ SUBROUTINE FAST_InitializeAll( t_initial, m_Glue, p_FAST, y_FAST, m_FAST, ED, SE
                     SlD%y, SlD%m, dt_module, Init%OutData_SlD, ErrStat2, ErrMsg2)
       if (Failed()) return
 
+      ! Pass nonlinear flag to SubDyn: true only when REDWIN DLL is active (CalcOption=3)
+      Init%InData_SD%SlDNonLinear = SlD%p%UseREDWINinterface ! REDWIN DLL returning nonlinear soil reaction forces      
+
       ! Add module to list of modules, return on error
       CALL MV_AddModule(m_Glue%ModData, Module_SlD, 'SlD', 1, dt_module, p_FAST%DT, &
                         Init%OutData_SlD%Vars, p_FAST%Linearize, ErrStat2, ErrMsg2)
--- a/modules/soildyn/src/SoilDyn.f90
+++ b/modules/soildyn/src/SoilDyn.f90
@@ -109,9 +109,11 @@ subroutine SlD_Init(InitInp, u, p, x, xd, z, OtherState, y, m, Interval, InitOut
    z%DummyConstrState = 0.0_ReKi
    OtherState%DummyOtherState = 0.0_ReKi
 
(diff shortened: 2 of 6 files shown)


andrew-platt (approved):

andrew-platt:
Oops. I meant to put this on rc-5.0.1. Will open a PR to backport it.

Issue #3277, opened 2026-03-20
FF: wake clipping issues


rthedin:
Using v5.0.0. Simple case setups and simple tests are showing unexpected behavior in wake clipping. With 5.0 deprecation of ` NumPlanes` and introduction of `numDFull`/`numDBuff`, I'm not seeing them behaving as I'd expect.

See animation attached.

https://github.com/user-attachments/assets/917adf16-d219-4fb5-b40d-ded0687c0728

andrew-platt:
Can you extend the low res box to the right?  The weirdness in wake development is really close to when the first plane from the last turbine exits the low res domain.  I'm wondering if something in how we handle that boundary crossing is messing up the upstream turbine wakes somehow (indexing or something peculiar in the wake handling).

rthedin:
I don't think that is the case. See animation below.

https://github.com/user-attachments/assets/1d3c5e8c-0736-4c5a-b8c3-43f0f1e25251

andrew-platt:
The root cause of the symptoms here were due to an error in the way wakes were getting dropped when one wake catches up to another.  This bug was introduced in #3071, and corrected in #3323.

andrew-platt:
Changed title since `NumDBuff` and `NumDFull` were not related to this issue.

PR #3071, opened 2025-11-17
Various changes to the handling of wake planes

luwang00:
**Feature or improvement description**
This PR changes the handling of wake planes in FAST.Farm, AWAE, and WD:
 * Removed `NumPlanes` as an input and replaced it with `NumDFull` and `NumDBuff` in the FAST.Farm input file.
 * Wake planes can propagate over a distance given by `NumDFull*RotorDiamRef` (the distance of full wake propagation).
 * Between a distance of `NumDFull*RotorDiamRef` and `(NumDFull+NumDBuff)*RotorDiamRef` (the buffer region), the wake deficits of the wake planes decay linearly with distance to zero.
 * All wake planes that propagated past the end of the buffer region are removed.
 * The number of wake planes of each rotor is now dynamic (up to a maximum allowed number estimated internally).
 * The coupling between WD instances and AWAE at each step is now based on the time-varying number of active wake planes.
 * If the number of wake planes exceeds the maximum allowed number, which should not happen in most cases, the excess planes downwind are removed with a warning.
 * Also added checking for wake planes that overtake their downwind neighbors. If found, a warning is given, and, if the simulation is allowed to continue, the offending wake plane is simply removed.

**Related issues**
This PR depends on PR #3066 of @deslaughter.

**Impacted areas of the software**
FAST.Farm, AWAE, WD

**Test results**
Updates to FAST.Farm r-tests are needed mainly due to input changes.

PR #3323, opened 2026-04-30
FAST.Farm: fix indexing on dropping wake planes that get dropped

andrew-platt:
Ready to merge

### Feature or improvement description
If a _FAST.Farm_ wake plane passes another, an error in the indexing was causing the wake plane getting passed to be dropped, but also causing the very last wake plane to get dropped as well.  This would result in the end of the wake jumping backwards whenever a wake was dropped from the middle of the sequence.

https://github.com/user-attachments/assets/6ad85fa8-2cc7-4208-bd20-b7a4748af6fd
Figure 1: single turbine case before bugfix.

https://github.com/user-attachments/assets/5a3ac9c8-aa62-4de2-b0b8-8cc58cc2ab46
Figure 2: single turbine case after bugfix.  Note the artifacts from the first wake plane - these will be addressed in a separate PR.

In the above simulations, the fast moving plane that is catching up to a slower moving plane is dropped.  ~This could be modified in a later PR.~ This method was dropped in favor of merging wake planes by averaging.

#### Solution: averaging of wake planes

As a fast moving plane catches up with a slow moving plane, the two planes are averaged together and saved.  Older planes are then shifted forward in indexing to account for the now missing plane.

https://github.com/user-attachments/assets/9386a301-e294-4994-badd-c49a42f364ff
Figure 3: A fast moving wake plane catching up to a slow moving wake plane is now merged together instead of simply dropping the fast moving plane.

#### Incidental changes

In addition to fixing the above issue, a couple other minor modifications are included in this PR:

* ~in calculating `MaxNumPlanes`, increased coefficient from 15 to 30.  One of the test cases run during diagnostics unexpectedly hit this limit.  The calculation itself should be looked at sometime as it doesn't appear to account for the low res timestep between planes, which is likely why this limit was hit.~. Will diagnose this later
* changed `NumDFull` and `NumDBuff` from integer to real.  There was no reason for these to be integers.

### Related issue, if one exists
#3277 -- this was due to this bug

### Impacted areas of the software
_FAST.Farm_ wake, but only in cases where the simulation is setup poorly and wake planes are passing each other and getting dropped as a result.

### Additional supporting information
Note that the are some strange artifacts in the first plane.  These will be corrected in different PR.

### Generative AI usage
Bug diagnostics assisted by Anthropic Claude <claude@anthropic.com>

### Test results, if applicable
Regression tests are properly setup, so they did not show this behavior,

### DIFF
--- a/modules/wakedynamics/src/WakeDynamics.f90
+++ b/modules/wakedynamics/src/WakeDynamics.f90
@@ -972,8 +972,9 @@ subroutine WD_UpdateStates( t, n, u, p, x, xd, z, OtherState, m, errStat, errMsg
 
       if ( xd%x_plane(i) > p%x_Buff ) then
 
-         xd%NumPlanes = max( xd%NumPlanes - 1.0, 2.0 )
+         xd%NumPlanes = max( xd%NumPlanes - 1.0, 2.0 )   ! Plane indexing includes 0, hence the -1.0
 
+      ! If a plane overtakes another plane, drop the plane that is doing the overtaking and shift all remaining planes forward. We skip checking the last plane since it can't overtake anything.
       else if ( i+1 < NINT(xd%NumPlanes) .and. xd%x_plane(i) >= xd%x_plane(i+1) ) then
 
          call SetErrStat(ErrID_Warn, ' Turbine '//trim(num2lstr(p%TurbNum))//' wake plane '//trim(num2lstr(i))//' (x_plane='//trim(num2lstr(xd%x_plane(i)))//') has overtaken wake plane '//trim(num2lstr(i+1))//' (x_plane='//trim(num2lstr(xd%x_plane(i+1)))//'). Offending wake plane removed. Reduce f_c to prevent planes from passing each other. ', errStat, errMsg, RoutineName)
@@ -982,14 +983,9 @@ subroutine WD_UpdateStates( t, n, u, p, x, xd, z, OtherState, m, errStat, errMsg
             return
          end if
 
-         ! Remove offending plane and shift everything behind up
-
-         xd%NumPlanes = xd%NumPlanes - 1.0
-
-         ! Didn't check xd%NumPlanes >= 2 here. The first wake plane is unlikely to move upwind of the rotor.
-
+         ! Overwrite the i plane that is passing with the i+1 plane, and shift all following planes back by one plane.
+         ! Didn't check xd%NumPlanes >= 2 here. The first wake plane is unlikely to pass the second plane.
          do j = i+1,NINT(xd%NumPlanes)-1
-
              xd%Vx_wind_disk_filt(j-1) = xd%Vx_wind_disk_filt(j)
              xd%x_plane      (    j-1) = xd%x_plane      (    j)
              xd%TI_amb_filt  (    j-1) = xd%TI_amb_filt  (    j)
@@ -1003,8 +999,10 @@ subroutine WD_UpdateStates( t, n, u, p, x, xd, z, OtherState, m, errStat, errMsg
              xd%Vx_wake2     (:,:,j-1) = xd%Vx_wake2     (:,:,j)
              xd%Vy_wake2     (:,:,j-1) = xd%Vy_wake2     (:,:,j)
              xd%Vz_wake2     (:,:,j-1) = xd%Vz_wake2     (:,:,j)
-
          end do
+
+         ! Now that we shifted the planes up, remove the last one
+         xd%NumPlanes = xd%NumPlanes - 1.0
 
       end if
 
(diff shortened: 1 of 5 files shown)


andrew-platt:
This bug was introduced in #3071

jjonkman:
I'm not sure I fully understand why 15 was changed to 30 when calculating `MaxNumPlanes`, but the derivation of the original equation is included in the implementation plan _FAST.Farm_PLan_WakeExtentAndBuffering_Rev2.docx_.  Effectively, `DT_Low` cancels out.  A safety factor of 1.25 was used originally, but increasing to 30 effectively doubles this to 2.5, which sounds excessive to me.

What is causing the issue with the initial plane?  How is this case set up to have step change in wake response?  The wake solution looks odd to me overall, but perhaps the turbine operation set-up is odd?

andrew-platt:
@luwang00 and I weren't entirely sure where the 1.25 came from.  I'll change the 30 back to 15 to keep in line with the original derivation.

I'm still tracking down the issue with the initial plane.  I suspect an uninitialized variable someplace - we see different results for certain vtk points depending on compilation with gcc, but no notable issues with intel (double precision giving `NaN` or 6 digit velocity values, and single precision giving closer to reasonable outlier garbage for the same points).  This appears only on the first timestep, and gradually clears up on later timesteps (some artifacts still persist for a long time due to wake interactions). Since that's a completely separate issue from this one, I'll track it down separately and address it with a separate PR.

andrew-platt:
There are also oddities in this particular setup.  But it is highlighting some issues we should fix anyhow.

modules/wakedynamics/src/WakeDynamics.f90
```diff
@@ -970,42 +970,61 @@ subroutine WD_UpdateStates( t, n, u, p, x, xd, z, OtherState, m, errStat, errMsg
...
 
-         end do
+            ! Merge the i and i+1 plane by averaging them together
+            xd%Vx_wind_disk_filt(i) = (xd%Vx_wind_disk_filt(i) + xd%Vx_wind_disk_filt(i+1)) / 2.0_ReKi
+            xd%x_plane      (    i) = (xd%x_plane      (    i) + xd%x_plane      (    i+1)) / 2.0_ReKi
+            xd%TI_amb_filt  (    i) = (xd%TI_amb_filt  (    i) + xd%TI_amb_filt  (    i+1)) / 2.0_ReKi
+            xd%D_rotor_filt (    i) = (xd%D_rotor_filt (    i) + xd%D_rotor_filt (    i+1)) / 2.0_ReKi
+            xd%YawErr_filt  (    i) = (xd%YawErr_filt  (    i) + xd%YawErr_filt  (    i+1)) / 2.0_ReKi
+            xd%p_plane      (  :,i) = (xd%p_plane      (  :,i) + xd%p_plane      (  :,i+1)) / 2.0_ReKi
+            xd%xhat_plane   (  :,i) = (xd%xhat_plane   (  :,i) + xd%xhat_plane   (  :,i+1)) / 2.0_ReKi
```
jjonkman: xhat_plane is a unit vector normal to the plane.  It should be renormalized after averaging.

andrew-platt: good point!

andrew-platt: fixed in e4068931b

modules/wakedynamics/src/WakeDynamics.f90
```diff
@@ -996,7 +996,7 @@ subroutine WD_UpdateStates( t, n, u, p, x, xd, z, OtherState, m, errStat, errMsg
             xd%D_rotor_filt (    i) = (xd%D_rotor_filt (    i) + xd%D_rotor_filt (    i+1)) / 2.0_ReKi
             xd%YawErr_filt  (    i) = (xd%YawErr_filt  (    i) + xd%YawErr_filt  (    i+1)) / 2.0_ReKi
             xd%p_plane      (  :,i) = (xd%p_plane      (  :,i) + xd%p_plane      (  :,i+1)) / 2.0_ReKi
-            xd%xhat_plane   (  :,i) = (xd%xhat_plane   (  :,i) + xd%xhat_plane   (  :,i+1)) / 2.0_ReKi
+            xd%xhat_plane   (  :,i) = TwoNorm(xd%xhat_plane   (  :,i) + xd%xhat_plane   (  :,i+1))       ! renormalize
```
jjonkman: You need compute xhat_plane as you had before (by averaging the two vectors), then divide that vector by its TwoNorm to keep it as a unit vector.

andrew-platt: The correct fix: b0e0ad536

luwang00 (approved):
@andrew-platt Thanks for fixing this!

Issue #3386, opened 2026-07-11
OLAF: vtk files naming


RBergua:
**Bug description**
I'm running OLAF by means of the AeroDyn standalone driver and requesting some velocity outputs. I'm requesting the wind data from `TSart` = 1.00 s.
```
--------------------------- OUTPUT OPTIONS (wake visualizations) ---------------------------------------------------
0       WrVTk              Outputs Visualization Toolkit (VTK) (independent of .fst option) {0: NoVTK, 1: Write VTK in the folder vtk_fvw at each time step defined in VTK_fps, 2: outputs at time steps defined by VTK_fps, but ensures that a file is written at the beginning and the end of the simulation (typically used with VTK_fps=0)} (flag)
0       nVTKBlades         Number of blades for which VTK files are exported {0: No VTK per blade, n: VTK for blade 1 to n} (-) 
1       VTKCoord           Coordinate system used for VTK export. {1: Global, 2: Hub, default: 1} 
0       VTK_fps            Frame rate for VTK output (frames per second) {"all" for all glue code timesteps (1/DTAero in a AeroDyn input file), default for all OLAF timesteps} [used only if WrVTK=1]. It has to be a multiple of 1/DTfvw. 
1       nGridOut           Number of grid outputs (the result is written down in the vtk_fvw folder using the name used in GridName below).
GridName    GridType (1 = velocity, 2 = velocity and vorticity) TStart TEnd           DTOut (all = AeroDyn time step, default: OLAF time step)     XStart    XEnd   nX    YStart   YEnd    nY    ZStart   ZEnd   nZ
(-)           (-)                                                (s)    (s)            (s)                                                          (m)      (m)    (-)    (m)     (m)     (-)    (m)     (m)    (-)
"2D"           1                                                 1.00   2.00          0.02                                                        4.6242   4.6242   1    -2.5     2.5     51   2.18843  2.18843   1
```

Since I'm using a time step that is 0.002 s, I would expect that the first file written down corresponds to the step 500 (1/0.002).

However, when reaching t = 1.00 s, OLAF writes the first file as 000000000.vtk (instead of 000000500.vtk) and the second file as 000000510.vtk. I understand that this second file is the proper one taking place at instant t = 1.02 s (step = 1.02/0.002 = 510). See below for reference:
<img width="599" height="174" alt="Image" src="https://github.com/user-attachments/assets/01d0a19f-5f04-4a1c-bdf1-4d9f6c2a81fd" />

[...]

PR #3389, opened 2026-07-13
OLAF: Fix incorrect VTK grid output filename on first write

RBergua:

Fixes an issue in `FVW_CalcOutput` where OLAF's VTK grid output filename on first write was always 0, regardless of when it was actually called. For example, previously, if `WrVTK = 0` and `TStart > 0`, the code would still name the first grid file 000000000.vtk instead of the correct step index.

With this change, there is a new logic: `m%FirstCall` is initialized `True` and changed to `False` unconditionally at the end of the first `FVW_UpdateStates` call. So, it properly identifies the true first `FVW_CalcOutput` call (t = 0) regardless of `WrVTK` or `nGridOut` settings.

This change is implemented in lines 1540–1541. Additionally, trailing whitespace previously present in the file has been removed.	

https://github.com/OpenFAST/openfast/issues/3386

OLAF output VTK filename.


andrew-platt (approved):
I haven't double checked all the logic, but your testing shows this now works as expected.

Issue #3395, opened 2026-07-18
Linearization (.lin) files are written with tabular OutFmt (default ES10.3E2, ~4 significant digits), corrupting wide-dynamic-range Jacobians on disk


mayankchetan:
**Bug description**

The glue code writes everything in linearization (`.lin`) files — the Jacobian matrices
(A/B/C/D, `dUdu`/`dUdy`, module blocks), the operating-point column, and the header scalars
(including *Rotor Speed*, which MBC3 postprocessors consume) — using the OpenFAST input's
tabular `OutFmt`, whose documented default `"ES10.3E2"` carries only ~4 significant digits.

For models whose state matrices span a wide dynamic range — e.g. BeamDyn blades with fully
populated sectional 6×6 mass/stiffness matrices, where the assembled A spans ~15 orders of
magnitude — 4 digits is not enough to round-trip the matrix through disk. Every downstream
eigenanalysis (openfast_toolbox / pyFAST, MBC3 / Campbell workflows, ACDC) then reports
physically impossible results for a parked, passive structure: spurious positive eigenvalues,
negative damping ratios, and merged or missing modes. The corruption is invisible in the
`.lin` file itself and is easily misread as a physics or solver defect (we initially did).

Evidence that this is purely a write-precision artifact:

1. Two decks differing **only** in `OutFmt` (`ES10.3E2` vs `ES18.9E3`) produce clean vs
   corrupt eigenanalyses from the same simulation.
2. Re-truncating the wide-format matrices to 4 significant digits in numpy reproduces the
   corrupt eigenvalues to the digit.
3. A patched writer (fixed wide format) on a default-`OutFmt` deck is clean.

**To Reproduce**

1. Compile `dev` (double precision).
2. Linearize a parked BeamDyn-blade case with fully populated sectional matrices (e.g. an
   IEA reference blade) using the default `OutFmt`.
3. Run any MBC3/eigenanalysis on the `.lin` file.
4. Observe positive real parts / negative damping for a passive structure; repeat with
   `OutFmt "ES18.9E3"` and observe a clean spectrum.

**Expected behavior**

`.lin` numeric output round-trips the computed matrices at full working precision, independent
of the *time-series* formatting choice in `OutFmt`.

**Fix**

PR #3393 (`f/lin-full-precision`): all numeric `.lin` output uses a fixed
`ES24.15E3` format; tabular time-series output continues to honor the user's `OutFmt`.
[...]

PR #3396, opened 2026-07-18
BeamDyn: rotating-frame-consistent linearization Jacobians when RotStates=True

### LINKED ISSUE #3394: BeamDyn linearization with RotStates=True returns frozen-frame Jacobians flagged RotatingFrame=T — spurious spin-softening in MBC3/Campbell results
**Bug description**

When BeamDyn is linearized with rotating states (`RotStates = True`, the default), the state
Jacobians it returns are expressed in a **frozen, root-aligned inertial snapshot** of the blade
frame, but the `.lin` file flags those rows/columns `RotatingFrame = T`. MBC3-based
postprocessors (openfast_toolbox / pyFAST Campbell workflows, ACDC) trust that flag and apply
their rotating-frame Ω-corrections, which injects **spurious spin-softening**: coupled mode
frequencies droop with rotor speed following `f² = f₀² − αΩ²` even for modes that should be
speed-independent.

Observable symptom: in a Campbell diagram of the NREL 5-MW with BeamDyn blades, the tower
side-side branch droops from 0.3221 Hz at standstill to 0.2915 Hz at 25 rpm (α ≈ 0.108,
fit residual 9e-4; 2.29 % droop already at 12 rpm). The ElastoDyn-blade reference is flat, and
a `RotStates = False` control shows only mild physical stiffening — no droop. The same defect
reproduces on the official IEA-15/3.4/10/22 reference models at v5.0.0 tags
(α = 0.143 / 0.267 / 0.124 / 0.234 respectively; raw droop up to ~8 % on IEA-22).

Mechanism: two frame effects are missing from the returned Jacobians —

1. the states are anchored to `OtherState%GlbRot` (frozen at initialization) rather than the
   rotating root frame the `RotatingFrame = T` flag promises, and
2. the transport (Coriolis-type) term of the frame change is absent.

The rotation-only part of this transform existed once (PR #2063) but was removed in
`5fbd380f8`; note the fix below is **not a revert** — the deleted code was ≈identity at the
operating point and lacked the transport term, so it never fixed the droop either.

The consistent similarity transform, applied per paired `(q, dq/dt)` state triplet
(pairing offset Nx/2), is

```
A_r = S·A·S⁻¹ + blkdiag(−ω̃)     with  S = [[I, 0], [−ω̃, I]],
B_r = S·B,   C_r = C·S⁻¹
```

where ω is the **full 3-vector** root `RotationVel` mapped into the `GlbRot` basis (scalar Ω is
not sufficient — e.g. precone puts a z-component of −0.0548 rad/s at 12 rpm on the 5-MW). The
transform is exactly identity at standstill.

**To Reproduce**

1. Compile `dev` (double precision) at or near `0160d29a0`.
2. Run a NREL 5-MW `CompElast = 2` linearization sweep, 0–25 rpm (`CalcSteady`, 1 azimuth
   snapshot suffices), with `OutFmt "ES18.9E3"` so the `.lin` write precision does not add a
   second artifact (see the companion OutFmt issue).
3. Post-process with any MBC3/Campbell tool and track the tower side-side branch.
4. The branch follows `f² = f₀² − αΩ²` instead of staying flat; repeating with ElastoDyn
   blades or `RotStates = False` removes the droop.

**Expected behavior**

Jacobians flagged `RotatingFrame = T` are frame-consistent with that flag, so
speed-independent branches stay flat: with the fix below, 5-MW tower-SS flatness across
0–25 rpm improves from 2.29 % (at 12 rpm) to 0.015 %, and the IEA fleet flattens to
0.111 % / 0.029 % / 0.026 % / 0.379 % (IEA-15/3.4/10/22) — the IEA-22 residual is also present
in the independent python-side reference transform (0.361 %) and is a known second-order
effect (large-prebend blade / Wiener-Milenković first-order limit), not part of this defect.

**Fix**

Implemented and fleet-verified on branch
[`mayankchetan/openfast:f/bd-rotstates-framefix`](https://github.com/mayankchetan/openfast/tree/f/bd-rotstates-framefix)
— PR to follow. The transform is applied only at linearization snapshots (a new optional
`IsLin` flag threaded from the glue), so the tight-coupling solver and AeroMap keep the raw
frozen-frame Jacobians their Newton iterations require. Standstill `.lin` output is
bit-identical to unpatched.

Known limitation (documented, not addressed): floating platforms — platform/yaw rates fold
into the root ω, where MBC3's scalar-Ω assumption independently breaks down anyway.

**OpenFAST Version**

`dev` @ `0160d29a0` (also reproduced on v5.0.0 tags of the IEA reference models); double
precision, gfortran, RelWithDebInfo, macOS (arm64).

**Additional context**

Discovered while chasing anomalous tower frequencies in BD-blade Campbell diagrams. A second,
independent defect was isolated during the same investigation — `.lin` files are written with
the tabular `OutFmt` (~4 significant digits), which corrupts wide-dynamic-range BD Jacobians
on disk — filed separately with fix PR #3393. The 5-MW evidence above is immune to that
artifact by construction (diagonal sectional matrices) and was additionally re-baselined with
wide-format decks.

Developed with Anthropic Claude (Claude Code) under human direction.

mayankchetan:

Fixes #3394. With `RotStates = True` (the default), the BeamDyn linearization Jacobians are a
frozen, root-aligned snapshot in the inertial frame. But the `.lin` rows and columns are
flagged `RotatingFrame = T`. MBC3-based postprocessors thus inject spurious spin-softening
(`f² = f₀² − αΩ²`) into speed-independent branches. Example: the 5-MW tower side-side branch
droops 2.29 % at 12 rpm, and the IEA-22 droops ~8 % at rated speed.

This PR makes the returned Jacobians frame-consistent with the flag. A new subroutine
`BD_JacRotFrame` in `modules/beamdyn/src/BeamDyn.f90` applies, per paired `(q, dq/dt)` state
triplet (pairing offset Nx/2):

```
A_r = S·A·S⁻¹ + blkdiag(−ω̃)     with  S = [[I, 0], [−ω̃, I]]
B_r = S·B,   C_r = C·S⁻¹
```

to `dXdx` / `dXdu` / `dYdx`. Here ω is the full root `RotationVel` 3-vector, mapped into the
frozen `GlbRot` basis. A scalar Ω is not sufficient: precone alone puts −0.0548 rad/s in z at
12 rpm on the 5-MW. At standstill the transform is exactly the identity. This is **not** a
revert of the PR #2063 code that `5fbd380f8` removed. That code was ≈identity at the operating
point, and it lacked the transport (Coriolis) term. It therefore never produced rotating-frame
Jacobians.

Design choice — where the transform lives. A new **optional `IsLin` flag** gates the
transform. The flag is threaded `ModGlue_Linearize → FAST_JacobianPInput /
FAST_JacobianPContState (FAST_Funcs.f90) → BD`. The transform thus applies **only at
linearization snapshots**. The tight-coupling solver (`BuildJacobianTC`) and the AeroMap calls
are unflagged. They keep the raw frozen-frame Jacobians that their Newton iterations require.
All existing callers use keyword arguments. No C-binding, Simulink, or module-driver
interfaces are affected. No registry regeneration is needed. An alternative placement is also
workable: apply the transform glue-side in `FAST_ModGlue.f90` and keep the module API
flag-free. We are glad to rework to that form if the maintainers prefer it.

Runtime guards: fatal on a state-layout-invariant violation, and fatal on a gross
re-anchoring deviation (>1e-3). The code warns once per snapshot on a measurable re-anchoring
deviation (>1e-6), on ω̇ ≠ 0, and on |WM parameters| > 0.2 (the first-order
rotation-composition limit).

Known limitation (documented in #3394): floating platforms. Platform and yaw rates fold into
the root ω. There, the scalar-Ω assumption of MBC3 independently breaks down.

Fixes #3394. Companion defect: #3395, fixed by PR #3393 (the `.lin` full-precision writer).
That defect is independent, and it also corrupts BD Campbell work. The two PRs are
independent, but one investigation isolated both. The verification input files here
neutralize #3395 with `OutFmt "ES18.9E3"`.

- `modules/beamdyn/src/BeamDyn.f90` — new `BD_JacRotFrame`, applied in `BD_JacobianPInput` /
  `BD_JacobianPContState` when `p%RotStates .and. IsLin`. The dead `RotateStates`
  declarations are removed.
- `modules/openfast-library/src/FAST_Funcs.f90`, `FAST_ModGlue.f90` — these thread the
  optional `IsLin` flag from `ModGlue_Linearize` only.
- Time marching is byte-identical code. Standstill linearization output is bit-identical.

This branch is based on `rc-5.0.1` to target the upcoming patch release (rebased 2026-07-21
from dev @ `0160d29a0`; the three changed files are identical in both bases).
The acceptance gate was re-run on an `rc-5.0.1` build of this branch and passes: tower-SS
flatness 0.015 % over 0–12 rpm, all damping ratios positive.

Verification (double precision, dev @ `0160d29a0`, wide-format input files):

- **NREL 5-MW Campbell 0–25 rpm** (CompElast=2): the tower-SS flatness improves from 2.29 %
  to **0.015 %**. The patched results match an independent python-side transform of the
  unpatched matrices to ≤0.02 %. They match the ElastoDyn reference to ≤0.6 %. All damping
  ratios are positive.
- **IEA fleet** (29 cases; IEA-15/3.4/10/22 ported to dev): patched Fortran vs the python
  reference agrees to ≤0.022 % per point. The SS flatness is 0.111 % / 0.029 % / 0.026 % /
  0.379 %. The IEA-22 residual equals the python reference's own 0.361 % (a second-order
  WM/prebend effect, not introduced here). The raw unpatched droop there is ~8 %.
- **Standstill**: the `.lin` numbers are bit-identical, patched vs unpatched
  (timestamp-only diff).
- **ω mapping**: an instrumented print reproduced the per-blade target vector, with the
  −2.5° precone z-component included (the instrumentation is removed).

**Generative AI usage**

Developed with Anthropic Claude (Claude Code) under human direction. The whole branch was
independently reviewed before submission.

r-test impact: the baselines change for `5MW_Land_BD_Linear`, `5MW_Land_BD_Linear_Aero`, and
`Damped_Beam_Rotating`. This change is the fix at work: the rotating-case Jacobians change by
design. `Damped_Beam_Rotating` runs clean on the patched binary. The standstill beam cases
are identical. All `CompElast = 1` cases are unaffected. The companion r-test branch
(17 regenerated `.lin` baselines) is prepared on the same branch name and is rebased onto
r-test `rc-5.0.1`. This PR stays a **draft** until that branch is opened as an r-test PR. We
can regenerate baselines on request, or follow the maintainer process.


mayankchetan:
Follow-up commit `f230c16a4` from an additional adversarial review pass on the branch: two guard-ordering edge cases in `BD_JacRotFrame` (no change to the transform itself).

1. The standstill early-return now precedes the re-anchoring invariant guard — at ω = 0 every operation is an exact identity regardless of frame alignment, so the routine returns the Jacobian unchanged instead of potentially aborting on a mis-anchored root. This makes the advertised standstill no-op unconditional.
2. The zero-speed path no longer returns silently when the root is accelerating (ω = 0, ω̇ ≠ 0, e.g. a LinTimes snapshot during startup): it now emits the same omitted-ω̇-transport-term warning as the spinning path before returning.

Re-verified after the change: the 7-speed 5-MW acceptance checker passes with identical numbers (tower-SS flatness 0.015 %).

PR #3414, opened 2026-08-01
OLAF: VTK outputs at non-equidistant grid points (nGridOut)

RBergua:
Currently, OLAF allows to extract information at points that are equidistant. For example:
```
4       nGridOut           Number of grid outputs.
GridName    GridType (1 = velocity, 2 = velocity and vorticity) TStart TEnd  DTOut XStart    XEnd   nX    YStart   YEnd    nY    ZStart   ZEnd   nZ
(-)           (-)                                                (s)    (s)   (s)   (m)      (m)    (-)    (m)     (m)     (-)    (m)     (m)    (-)
"1D"           1                                                  1     2    0.1    50       50      1     -75      75     20    100     100    1
```
Basically, OLAF does: linspace(Start,End,n).

However, sometimes, the user is interested in a non-equidistant distribution of grid points. See for example the requested points in the OC7 Phase III WP 3.2 project:
image

In this case, the discretization within the area covered by the rotor is a fine discretization of 0.25R while the discretization outside this area is coarser (0.3R). With the original `nGridOut` in OLAF, the user would have to define three grids to request all the information of interest. For example:
Grid 1: `YStart`: -2.5R, `YEnd`: -R, `n`: 6. 
Grid 2: `YStart`: -0.75R, `YEnd`:0.75, `n`: 7. 
Grid 3: `YStart`: R, `YEnd`: 2.5R, `n`: 6.

Since the project requests 6 hotwire arrays at 1D, 2D, 4D, 6D, 8D, and 10D downstream from the rotor, this would translate into 6*3 = 18 grids requested in OLAF. This is less than ideal because it writes many files at every time step and the user has to postprocess and consolidate them in a single file per hotwire array.

To avoid this, I have modified the grid table: a `Start` cell can be a **number** (equidistant, unchanged behavior) or a **quoted filename** (explicit list, one coordinate per line, strictly ascending order, without comments). This filename can be defined at the `XStart` and/or `YStart` and/or `ZStart`. Such file(s), e.g., “Ypoints.dat”, must be available in the same location as the OLAF input file.

When using the quoted filename (list-based), `GridType` must be `1`. This list-based with non-equidistant points does not support the computation of vorticity (for reference, the vorticity is computed based on the gradient using the wind velocity and the distance between grid points).

All this logic is also included in the documentation: docs/source/user/aerodyn-olaf/InputFiles.rst

Below you can see an example requesting equidistant points (3 files as output + postprocessing) vs non-equidistant points (1 file as output without postprocessing):
image

For reference, when equidistant points are requested, the VTK file is written down using the `StructuredPoints` format. Example of the first file with the 6 equidistant grid points:
image

For non-equidistant points (new), the `RectilinearGrid` format is used. Example of the file with the 19 non-equidistant grid points:
image

Finally, in case the user tries to use the list-based approach and requests the vorticity, an error message will be printed:
image

OLAF VTK outputs (more locations available).

This new feature does not impact the r-test. However, several verifications have been performed to ensure the proper behavior (see above).

### DIFF
--- a/modules/aerodyn/src/FVW_IO.f90
+++ b/modules/aerodyn/src/FVW_IO.f90
@@ -106,16 +106,16 @@ SUBROUTINE FVW_ReadInputFile( FileName, p, m, Inp, ErrStat, ErrMsg )
          call ReadGridOut(sLine, m%GridOutputs(i)); if(Failed()) return
          ! Resolve each axis to an explicit coordinate array (regardless of whether it is a list file or a range)
          call ResolveGridAxis(m%GridOutputs(i)%xStart, m%GridOutputs(i)%xEnd, m%GridOutputs(i)%nx, &
-                               m%GridOutputs(i)%xListFile, m%GridOutputs(i)%xPts); if(Failed()) return
+                               m%GridOutputs(i)%xListFile, m%GridOutputs(i)%xPts, ErrStat2, ErrMsg2); if(Failed()) return
          call ResolveGridAxis(m%GridOutputs(i)%yStart, m%GridOutputs(i)%yEnd, m%GridOutputs(i)%ny, &
-                               m%GridOutputs(i)%yListFile, m%GridOutputs(i)%yPts); if(Failed()) return
+                               m%GridOutputs(i)%yListFile, m%GridOutputs(i)%yPts, ErrStat2, ErrMsg2); if(Failed()) return
          call ResolveGridAxis(m%GridOutputs(i)%zStart, m%GridOutputs(i)%zEnd, m%GridOutputs(i)%nz, &
-                               m%GridOutputs(i)%zListFile, m%GridOutputs(i)%zPts); if(Failed()) return
+                               m%GridOutputs(i)%zListFile, m%GridOutputs(i)%zPts, ErrStat2, ErrMsg2); if(Failed()) return
          ! Error checking
          if (Check(m%GridOutputs(i)%nx<1, 'Grid output nx needs to be >=1')) return
          if (Check(m%GridOutputs(i)%ny<1, 'Grid output ny needs to be >=1')) return
          if (Check(m%GridOutputs(i)%nz<1, 'Grid output nz needs to be >=1')) return
-         ! Vorticity requires equidistant spacing. GridType must be 1 when using list files.
+         ! Vorticity (GridType=2) requires equidistant spacing. GridType must be 1 when using list files.
          if (m%GridOutputs(i)%type==idGridVelVorticity) then
             if (Check( len_trim(m%GridOutputs(i)%xListFile)>0 .or. &
                        len_trim(m%GridOutputs(i)%yListFile)>0 .or. &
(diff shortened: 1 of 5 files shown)


modules/aerodyn/src/FVW_IO.f90
```diff
@@ -327,24 +342,112 @@ subroutine ReadGridOut(sLine, GridOut)
...
+
+      if (len_trim(ListFile) == 0) then
+         ! Equidistant points, expressed as an explicit array
+         allocate(Pts(max(n,1)))
+         do j = 1, n
+            Pts(j) = AStart + (AEnd - AStart) * real(j-1,ReKi) / real(max(n-1,1),ReKi)
+         enddo
+         ErrStat2 = ErrID_None; ErrMsg2 = ''
+         return
+      endif
```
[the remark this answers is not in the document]
andrew-platt: This is a bit safer.

modules/aerodyn/src/FVW_IO.f90
```diff
@@ -327,24 +342,112 @@ subroutine ReadGridOut(sLine, GridOut)
...
+         if (Pts(j) <= Pts(j-1)) then
+            ErrStat2 = ErrID_Fatal
+            ErrMsg2  = 'Grid point list file "'//trim(FullFile)//'" must be strictly increasing (no duplicates); '// &
+                    'value at line '//trim(Num2LStr(j))//' ('//trim(Num2LStr(Pts(j)))//') is not greater than '// &
+                    'the previous value ('//trim(Num2LStr(Pts(j-1)))//').'
+            return
+         endif
+      enddo
+
+      ErrStat2 = ErrID_None
```
andrew-platt: We usually put the errstat setting at the beginning of the routine instead of end.  This is ok as long as there is no way it could end prematurely.

modules/aerodyn/src/FVW_IO.f90
```diff
@@ -327,24 +342,112 @@ subroutine ReadGridOut(sLine, GridOut)
...
+      if (n < 1) then
+         ErrStat2 = ErrID_Fatal
+         ErrMsg2  = 'Grid point list file "'//trim(FullFile)//'" contains no valid values.'
+         close(UnList)
+         return
+      endif
+
+      ! Read values
+      rewind(UnList)
+      allocate(Pts(n))
```
andrew-platt: see above for capturing allocation errors.

modules/aerodyn/src/FVW_IO.f90
RBergua: I have added the `ErrStat` at the beginning and I have made the error reporting explicit in `ResolveGridAxis`.

RBergua:
**This problem is not introduced by this PR. See for reference: https://github.com/OpenFAST/openfast/pull/3419**

@andrew-platt: For a reason that I still don't understand, I get the expected behavior when I compile locally with GCC 15.2.0. For example, I can see all the expected messages in the console:
image

However, when I compile with GitHub Actions (Intel Fortran compiler), the messages do not appear.
image

For reference, I'm runninng the exact same input files and compiled the same source code.

The code always works and the logic is the expected one. But the messages are not in the console. What is surprising to me is that this is at a global level. For example, if I define `DTfvw` with a very small value, I should get this message: 
```
DTfvw must be >= DTaero from AD15.
AeroDyn Driver encountered simulation error level: FATAL ERROR
Aborting AeroDyn_driver. 
```

However, with the Intel Fortran compiler, I just get:
```
AeroDyn Driver encountered simulation error level: FATAL ERROR
Aborting AeroDyn_driver. 
```

The behavior observed here reminds me this one: https://github.com/OpenFAST/openfast/issues/3301

RBergua:
**This problem is not introduced by this PR. See for reference: https://github.com/OpenFAST/openfast/pull/3419**

I have been doing some testing and found something interesting.

Since the original issue appeared with the executable compiled by GitHub Actions (Intel Fortran Compiler), I branched rc-5.0.1 from scratch (commit fd51a0bc5279c9ab6bd548fb41d921dc3fbe0c70) and added a regression test to check the console output. The plan was to reintroduce my modifications one at a time and find the point at which the messages stopped appearing.

Surprisingly, rc-5.0.1 doesn't print the messages either when built through GitHub Actions.

Steps I took (for future reference):
1. In `reg_tests`, I created a new folder (`manualTestCase`) to test the AeroDyn standalone driver (`OLAF_LC11.dvr`) with a set of input files. In the OLAF input file, I set `DTfvw` smaller than `DT` in the driver to trigger the fatal check in FVW_IO.f90 line 183:
`if (Check( Inp%DTfvw < p%DTaero, 'DTfvw must be >= DTaero from AD15.')) return`

2. I added a batch file, `.github\scripts\test_console_prints.bat`, to run the test case and capture the console output to a log file:
```
@echo off
:: Runs the OLAF test case and saves the console output to a log file.
set "CASE_DIR=reg_tests\manualTestCase"
set "LOG_FILE=%CASE_DIR%\console.log"
build\bin\AeroDyn_Driver.exe "%CASE_DIR%\OLAF_LC11.dvr" > "%LOG_FILE%" 2>&1
echo.
echo ----- Captured console output -----
type "%LOG_FILE%"
echo ------------------------------------
```

3. I added this new test as part of the `.github\workflows\deploy.yml`:
```
      - name: Test version output
        run: .github/scripts/test_windows_executables.bat

      - name: Test AeroDyn printing messages on console
        run: .github/scripts/test_console_prints.bat

      - name: Build MATLAB Mex File
```

When I trigger the deploy manually as part of the GitHub Actions, I get:
image

Result: When I trigger the deploy workflow manually, AeroDyn aborts as expected (since `DTfvw` < `DT`), but the expected error/warning messages are still missing from the console output. Since I can reproduce this on OpenFAST rc-5.0.1, I understand that the console-printing issue is not introduced by this PR.

andrew-platt (approved):

Issue #3429, opened 2026-08-13
OLAF: Wake singularity? Impact on rotor loads


RBergua:
As part of the OC7 Phase III WP 3.1 project, I'm simulating a 1:70 scale rotor. The system rotates at 240 rpm and the rotor diameter is 2.38132 m. The mean wind speed considered is 4.19 m/s and I use a turbulent wind speed representative of the wind tunel measurements. 

I observe the expected rotor thrust force and torque, but somehow, after 32 seconds I observe a significant drop in the rotor loading.
<img width="694" height="797" alt="Image" src="https://github.com/user-attachments/assets/f950254b-950f-453a-95d4-2315d1a4d5da" />

I performed a large number of sensitivity analysis:
1. Changed the number of near-wake panels. 
2. Changed the distribution of free and frozen panels in the near wake.
3. Ran with and without far wake.
4. Changed the wind fom unsteady to steady.
5. Included/excluded the tower shadow.
6. Included/excluded the unsteady aerodynamics in the polars.
7. Used constant polars instead of interpolating between Reynold numbers.
8. Ran with/without parallel solver (OMP).
9. Ran with single and double precision solver.

However, the behavior was always the same: after 32 seconds, the rotor loads dropped. For reference, I always use a time step of 0.002 seconds that corresponds to 2.88 deg azimuth in the rotor. This is already pretty demanding from a computational point of view.

Looking at the wake behavior, I can see that the near-wake clearly behaves different after 32 seconds.
<img width="1074" height="646" alt="Image" src="https://github.com/user-attachments/assets/01fc7d87-1da7-4f12-b438-307c47f781c6" />
The interesting part is that if I reduce the time step in half (from 0.002 s to 0.001 s), the same phenomena appears but at 16 s instead of 32 s. This indicates that, for some reason, the phenomena is triggered when the **step count** is **close to 16,000** (32/0.002=16/0.001).

@ebranlard: do you know what could be triggering this?

For reference, I attach the comparison between OLAF (orange) and BEMT (blue) below:
<img width="543" height="532" alt="Image" src="https://github.com/user-attachments/assets/2186e7da-42c0-4bbe-a162-7adbf52d49fc" />

RBergua:
[...]

PR #3430, opened 2026-08-14
AD: fix OLAF treecode near-core regularization floor and TwrInfl OpenMP data race

luwang00:
FVW treecode: the wake path used `Tree%DistanceDirect = 2*mean(RegParam)`, averaged over an over-allocated array whose sentinel tail drove it negative during wake buildup, disabling the near-core direct-evaluation fallback. Replace it with a per-node `maxRegParam` (max eps over the cell's particles); each branch now uses `distDirect = BranchFactor*radius + 2*maxRegParam` so control points inside a regularization core fall back to direct evaluation. Applied to both the particle and segment trees.

`TwrInflArray`: make `FirstWarn_TowerStrike` `firstprivate` and `ErrStat2/ErrMsg2` private in the OpenMP loop to avoid a data race.

In `ui_part_nograd_11`, the `idRegExp` mollifier `(1-exp(-(r/rc)^3))/r^3` differs from `1/r^3` by `<exp(-8)~3.4e-4` once `r>2*rc`, which is exactly the accuracy the far-field multipole already accepts at its `BranchFactor*radius+2*maxRegParam` floor. Treat the mollifier as 1 beyond that boundary (new `PART_REG_NRAD`/`PART_REG_CUT3` params) so the near-field direct kernel and the far-field multipole share the same `2*rc` cutoff. Since most near-field tree pairs have `r>>rc`, this skips `exp()` for the majority of evaluations, giving ~25-32 percent serial speedup on the treecode path with the output unchanged to `~1.8e-5` relative. Also caches `r^2`/`r^3`/`rc^3` to drop redundant calculations; the compact-support branch is refactored identically (bit-for-bit).

**Potential impact**
This bug fix can potentially generate better behaved rotor wake. Below are two snapshots of OLAF wake generated from the same AeroDyn driver/OLAF case provided by @RBergua. The simulation uses steady wind and all free near-wake panels. There are no frozen wake or far-wake panels. The two images show the same time instant. The only difference is the regularization fix.

*Steady wind; without regularization fix*
image

*Steady wind; with regularization fix*
image

Notice that the wake is significantly more ordered with the regularization fix. This is expected with the large `CoreSpreadEddyVisc=1000` and age-based core spreading. At the same time, the simulation with regularization fix is considerably slower; however, this is expected because the large core radii, especially with the older wake elements further downstream, force the code to resort to the costly direct and exact evaluation of induction velocity more often to properly account for core regularization. Previously, the incorrect regularization floor effectively disables near-core direct-evaluation fallback in many cases, resulting in faster simulation but highly disordered wake. To speed up the simulation with the regularization fix, we can reduce `CoreSpreadEddyVisc` as appropriate.

Also, note that the two simulations provide very similar rotor loads. However, the core-regularization fix can be important if the far wake is of interest of if simulating multiple tandem rotors.

Two snapshots of the wake with turbulent wind are shown below. The turbulent inflow makes the wake far less organized. Nevertheless, the wake with regularization fix shows more structure than the one without.

*Turbulent wind; without regularization fix*
image

*Turbulent wind; with regularization fix*
image

The instantaneous wake streamwise velocity computed with the direct particle method and the particle tree method is compared below. The agreement is fairly good.
image

There are several reported issues with OLAF. Unclear if this fix addresses any of them.

OLAF

**Generative AI usage**
Co-authored-by: Microsoft Copilot <copilot@microsoft.com>

One r-test needs rebasing: `ad_BAR_OLAF`. Only two output channels did not pass the threshold, but the changes are small, barely exceeding the threshold.

image
image

### DIFF
--- a/modules/aerodyn/src/FVW_VortexTools.f90
+++ b/modules/aerodyn/src/FVW_VortexTools.f90
@@ -72,11 +72,10 @@ module FVW_VortexTools
 
    !> The type tree contains some basic data, a chained-list of nodes, and a pointer to the Particle data that were used
    type T_Tree
-      type(T_VPart)  :: Part            !< Storage for all particles
-      type(T_VSgmt)   :: Seg             !< Storage for all segments
+      type(T_VPart) :: Part            !< Storage for all particles
+      type(T_VSgmt) :: Seg             !< Storage for all segments
       integer       :: iStep =-1       !< Time step at which the tree was built
       logical       :: bGrown =.false. !< Is the tree build
-      real(ReKi)    :: DistanceDirect
       type(T_Node)  :: Root            !< Contains the chained-list of nodes
    end type T_Tree
 
@@ -627,8 +626,6 @@ contains
       endif
       Tree%iStep  = iStep
       Tree%bGrown = .true.
-      ! Regularization floor is now applied per node via node%maxRegParam in the tree traversal
-      Tree%DistanceDirect = 0.0_ReKi
    end subroutine grow_tree_part
 
    !> Recursive function to grow/setup a tree. 
@@ -979,8 +976,6 @@ contains
       endif
       Tree_Seg%iStep  = iStep
       Tree_Seg%bGrown = .true.
-      ! Regularization floor is now applied per node via node%maxRegParam in the tree traversal
-      Tree_Seg%DistanceDirect = 0.0_ReKi
    end subroutine grow_tree_segment
 
    !> Recursive function to grow/setup a tree.
--- a/modules/aerodyn/src/FVW_VortexTools.f90
+++ b/modules/aerodyn/src/FVW_VortexTools.f90
@@ -63,6 +63,7 @@ module FVW_VortexTools
       real(ReKi)                         :: radius !< Typical dimension of a cell (max of x,y,z extent)
       real(ReKi),dimension(3)            :: center !< Used to store first the geometric center, then the vorticity center
       real(ReKi),dimension(3,10)         :: Moments
+      real(ReKi)                         :: maxRegParam = 0.0_ReKi !< Max reg. param (eps) over all particles in this node
       integer,dimension(:),pointer       :: iPart=>null()  !< indexes of particles stored in this node
       integer,dimension(:),pointer       :: leaves=>null()  ! NOTE: leaves are introduced to save memory
       type(T_Node),dimension(:), pointer :: branches =>null()
@@ -587,6 +588,7 @@ contains
          node%radius=0
          node%center(1:3)=Part%P(1:3,1)
          node%Moments=0.0_ReKi
+         node%maxRegParam = Part%RegParam(1)
          nullify(node%iPart)
          nullify(node%branches)
          allocate(node%leaves(1:1))
@@ -625,7 +627,8 @@ contains
       endif
       Tree%iStep  = iStep
       Tree%bGrown = .true.
-      Tree%DistanceDirect = 2*sum(PartRegParam)/size(PartRegParam) ! 2*mean(eps), below that distance eps has a strong effect
+      ! Regularization floor is now applied per node via node%maxRegParam in the tree traversal
+      Tree%DistanceDirect = 0.0_ReKi
    end subroutine grow_tree_part
 
    !> Recursive function to grow/setup a tree. 
@@ -675,6 +678,7 @@ contains
       GeomC = node%center ! NOTE: we rely on the fact that our parent has set this to the Geometric value 
       VortC = 0.0_ReKi
       wTot = 0.0_ReKi
+      node%maxRegParam = 0.0_ReKi
       ! --- Barycenter of vorticity of the node 
       do i = 1,node%nPart
          PartPos   = Part%P(:,node%iPart(i))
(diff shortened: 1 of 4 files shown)


RBergua:
To better understand the implications of this modification, I took the OLAF model from OC7 Phase III WP 3.1 and ran it with the original code and the proposed code.

The proposed modifications do not change the rotor loading:
image

However, there is a significant impact on the wake behavior. The OLAF model tested uses 3,400 near-wake panels (all of them free). No far-wake is included in the model. Before the fix, the `CoreSpreadEddyVisc` played a limited role. However, after the fix, the wake behavior is sensitive to it. Below I show the wake behavior in a horizontal plane at an elevation equal to the hub height.

Before the fix (`CoreSpreadEddyVisc = 1000`): 
wake_previous_to_fix_EddyViscosity1000

Affter the fix (`CoreSpreadEddyVisc = 500`):
wake_after_fix_EddyViscosity500

As can be observed, the behavior is very different. Before the proposed changes, we can observe highly unstable flow in the wake (despite using `CoreSpreadEddyVisc = 1000`).

This can also be observed in the vertical plane at different downstream distances. The wake deficit shown below corresponds to an averaged timeframe of 10 s (from 20 to 30 s).

Before the fix:
image

After the fix:
image

It's also important to note that after the proposed fix, the computational time increases for higher `CoreSpreadEddyVisc` values. Howeevr, the new contact-support option (`RegFunctionPart = 2` in line 26 of OLAF input file) in https://github.com/OpenFAST/openfast/pull/3457, brings the computational time back to normal.

Finally, this modification does not improve the issue reported here: https://github.com/OpenFAST/openfast/issues/3429

luwang00:
@RBergua Thanks for sharing the results! Regarding the increased computing time, I'm looking at implementing an alternative particle regularization function with true compact support. Right now, for particle/particle tree, OLAF either uses no regularization or exponential regularization. The latter is quite expensive. There is already a particle regularization function called `idRegCompact` in the code, but it is neither truly compact nor currently selectable. The plan is to replace it with a genuinely compact-support function that avoids transcendental calls like `exp` and allows the treecode near-core cutoff to be tightened. Hopefully, this will increase the speed by a nontrivial amount. A new `RegFunctionPart` input will be added to OLAF to let the user select particle regularization independently of the segment regularization controlled by `RegFunction`. `RegFunctionPart` will be inserted right after `RegFunction`. The options are `1` for exponential and `2` for compact. `Default` is `1`. In case you want to test it, here is the work-in-progress: https://github.com/luwang00/openfast/tree/f/OLAF (PR #3457)

ebranlard:
A good check is to compare the wake shapes between VelocityMethod =2 and 3 (particles only). The "best reference should be VelocityMethod=1 (segment N^2), but there is no one-to-one equivalence between the regularization of segments and particles. A rough overview of the wake shape for those three cases would be informative.

modules/aerodyn/src/FVW_VortexTools.f90
```diff
@@ -973,7 +979,8 @@ subroutine grow_tree_segment(Tree_Seg, nSeg, SegPoints, SegConnct, SegGamma, Seg
       endif
       Tree_Seg%iStep  = iStep
       Tree_Seg%bGrown = .true.
-      Tree_Seg%DistanceDirect = 2*sum(SegRegParam)/size(SegRegParam) ! 2*mean(eps), below that distance eps has a strong effect ! TODO REMOVE
+      ! Regularization floor is now applied per node via node%maxRegParam in the tree traversal
```
ebranlard: The `TODO REMOVE` confirms that I wanted to get rid of this...

modules/aerodyn/src/FVW_VortexTools.f90
```diff
@@ -1501,7 +1510,7 @@ recursive subroutine ui_tree_part_11(node, CP, Uind)
                enddo
             endif
          else
-            distDirect = max(BranchFactor*node%radius, DistanceDirect) ! Under this distance-> Direct eval., Above it -> quadrupole calculation
+            distDirect = max(BranchFactor*node%radius + 2.0_ReKi*node%maxRegParam, DistanceDirect) ! Direct eval. below this distance; floor = branch radius + 2*max(eps) in cell
```
ebranlard, line 1508: `DistanceDirect` ends up not being used anymore (since 0), we could likely remove it.

ebranlard:
I agree with the changes, the mean(eps) previously introduced was indeed a temporary hack introduced as a compromise.
If we want to match the N^2 computation, we do need to include the effects within the 2 Reg Param region of each particle. But with this new implementation, for large epsilon of a particle (will occur in the far wake, especially with large CoreSpreadEddyVisc) computations will get quite expensive. 

As a quick workaround, we should probably introduce a maximum reg param (either as part of the core model, or part of the tree model).

luwang00:
Hi @ebranlard, thanks for reviewing the changes.

Yes, we can remove `Tree%DistanceDirect` and `Tree_Seg%DistanceDirect` since these will always be zero now, but maybe we should retain the local argument/variable `DistanceDirect` of `ui_tree_part` and `ui_tree_segment`? I think we can still get nonzero values assigned in `LiftingLineInducedVelocities`.

Regarding the increased computing time, it looks like the compact particle regularization (PR #3457) is helping somewhat. The same run that took >5 days is on track to complete in just over 3 days (won't get exact time until it's done). Max regularization param is a good idea. I think it should be part of the core model. Enforcing that in the tree code could cause discontinuous velocity field, and we are back to the same problem. Also, it would be better to have a dynamic bound based on minimum core overlap ratio to ensure wake smoothness.

modules/aerodyn/src/FVW_VortexTools.f90
luwang00: Done. Removed `DistanceDirect` from the `T_Tree` type.

modules/aerodyn/src/FVW_VortexTools.f90
luwang00: Kept the local `DistanceDirect` here because `LiftingLineInducedVelocities` still uses it.

andrew-platt (approved):

PR #3442, opened 2026-08-24
SoilDyn initialization crash: stack overflow with MinGW

RBergua:
OpenFAST crashes silently with no error message when using SoilDyn (`CalcOption=1`, stiffness matrix mode) compiled with GCC/MinGW on Windows.

The program stops after printing `SoilDyn: linear stiffness passed directly to SubDyn as a stiffness matrix` and before SubDyn initializes. The same build works fine with Intel Fortran (e.g., compiling with GitHub Actions).
image

This SoilDyn's initialization crash is due to a **stack overflow** error.

The project's CMake already has a logic to avoid this stack overflow in the file `OpenfastFortranOptions.cmake`. But it’s inside a condition that only includes `CYGWIN`. If the user tries to compile using `MinGW-w64`, there is no increase in stack size. I just modified the condition:
`if(CYGWIN OR MINGW)`

Behavior after the fix:

1. GCC 15.2.0:
image

2. Intel Fortran compiler:
image

Fixes https://github.com/OpenFAST/openfast/issues/3301


andrew-platt (approved):
Nice detective work to find that!

PR #3449, opened 2026-08-29
AeroAcoustics improvements

bjonkman:
This pull request primarily addresses potential numerical issues in the aeroacoustics module. There are now safety checks around negative Reynolds, Mach, and Strouhal Numbers, overflow with 10^x, and division-by-zero. Some variables that may have been uninitialized are now set. 

It also 
- removes the warning in the AA module about the blade percentage when only one element is used. This 1-node use case would probably happen only in the AA driver.
- adds the Aeroacoustics driver and AirfoilInfo driver to the VS build. The aeroacoustics module was already in the cmake build scripts.
- changes the VS project for NWTC library to stop checking that the correct data types are passed through interfaces. The Debug mode didn't build because of FFTpack 5.1

Aeroacoustics, VS build

We have noticed at least one instance where the AA module output a large spike, which disappeared when re-run with the same executable and input files. I can't yet confirm if these changes fixed the issue, but I am hopeful that it helped.

**Generative AI usage**
Assisted by: Kiro:claude-opus-5


modules/aerodyn/src/AeroAcoustics_TNO.f90
```diff
@@ -10,6 +10,12 @@ MODULE TNO
 
    INTEGER,       PARAMETER :: TNOKi = ReKi
 
+   ! NOTE (bjj): The variables declared at module scope below are effectively SAVEd state. They are how the integrand
+   ! functions f_int1/f_int2/Pressure receive their parameters, because slatec_qk61 only accepts a function of one
+   ! variable. SPL_integrate() sets all of them on entry, so sequential calls (including calls for different rotors or
+   ! different blade nodes) are safe. This module is NOT reentrant, however: it must not be called from more than one
+   ! thread at a time, or one thread will overwrite another thread's flow conditions mid-integration.
```
andrew-platt: Does this have implications on using AA with FAST.Farm?

andrew-platt (approved):

.gitignore
```diff
@@ -69,3 +69,12 @@ openfast_io/openfast_io/_version.py
 *.code-workspace
 docker-compose.yml
 .devcontainer/
+
+# AI steering files for specific users
+.kiro/*
+
+# vs-build files
```
andrew-platt, line 76: I transferred these from the now deleted vs-build/.gitignore

andrew-platt:
Thanks for addressing the problematic math @bjonkman!

modules/aerodyn/src/AeroAcoustics_TNO.f90
bjonkman: Possibly. It's global data that is used in place of sending variables through subroutine arguments. If two threads are using/changing it at the same time, that would be a problem. 

It looks similar to the way AeroDyn used to implement the sub_brent method. We later simplified the logic to avoid those global variables. This would probably be a fairly straightforward task for AI. If I get some time to come back to this, maybe I'll take a second look.

andrew-platt: If it is really easy, it's worth doing now.  However I suspect it isn't a feature that gets used with FF (separate OF instances won't be additive at a given location).

andrew-platt:
I'm very confused on why some linearization tests are failing.  I wouldn't expect any of this code to by used in those cases (`5MW_Land_BD_Linear` and `MHK_RM1_Floating_MR_Linear`)

edit: after multiple reruns it succeeded.  This is not due to this PR, but a systemic issue with those cases.  We'll have to look into that separately later.

Issue #3463, opened 2026-09-14
AeroDyn v5.0.0 propagation-direction Jacobian uses stale wind rotations


Mohammad-Salik:
**Title:** AeroDyn v5.0.0 propagation-direction Jacobian uses stale wind rotations

**Bug description**

During linearization, AeroDyn perturbs `PropagationDir` without refreshing the dependent wind rotation matrices and rotation-enable state before evaluating the perturbed velocity.

At the validated operating point, the original implementation produces a zero propagation-direction column in `B` and zero direction feedthrough to turbine outputs in `D` across all 36 azimuths. InflowWind’s velocity-output Jacobian nevertheless correctly gives:

$$
\frac{\partial V_y}{\partial \psi} = -U
$$

**To Reproduce**

1. Compile OpenFAST v5.0.0 in double precision.
2. Linearize an aeroelastic case using uniform wind with vertical shear, zero mean propagation direction and upflow angle, and extended wind inputs enabled. Collect 36 azimuth snapshots.
3. Open the resulting `.lin` files and locate the propagation-direction input columns in `B` and `D`.
4. In the validated case, the state-input and turbine-output direction derivatives are zero, while the InflowWind lateral-velocity derivative is correctly `−U`.

These observations were validated at one operating point; equivalent behavior has not been established for arbitrary configurations.

**Expected behavior**

A propagation-direction perturbation should update the wind velocity evaluated by AeroDyn, allowing the corresponding aeroelastic derivatives to be calculated consistently with InflowWind.

**Screenshots, if applicable**

Not applicable; evidence comes from numerical comparisons of the `.lin` matrices.

**OpenFAST Version**

- Base version: OpenFAST v5.0.0
- Base commit: `2895884d2be01862173c88d70f86b358d2f1a50a`
- Local validation build: `v5.0.0-psi-validation`
- Architecture: 64 bit
- Precision: double

The validation build supports comparison of the original perturbation behavior, the repaired implementation, and an independent finite-difference implementation.

**System Information (please complete the following information):**

- OS: Windows, 64 bit
- Compiler: Intel Fortran `ifx` 2026.0, build `20260331`
- Build system: CMake with Ninja
- Relevant settings:
  ```text
  -DDOUBLE_PRECISION=ON
  -DOPENMP=OFF
  -DBLA_VENDOR=Intel10_64lp_seq
  -DCMAKE_BUILD_TYPE=Release
  ```

**Additional context**

[...]

PR #3468, opened 2026-09-18
AD: refresh wind rotations for direction linearization

Mohammad-Salik:

Fix the AeroDyn Jacobian perturbation for the InflowWind propagation-direction extended input.

`PerturbFlowField` currently perturbs `PropagationDir` while leaving the dependent `RotToWind`, `RotFromWind`, and `RotateWindBox` values unchanged. As a result, AeroDyn evaluates the perturbed case with the base wind rotation and the aeroelastic propagation-direction sensitivity is lost.

This change resets the cached rotation quantities from the base flow field for each perturbation and recomputes them when `PropagationDir` is perturbed. No change is made to the finite-difference step size.

Closes #3463

- AeroDyn linearization
- Extended InflowWind propagation-direction input

No time-domain behavior is intentionally changed; the modification is confined to the flow-field copy used inside the input-Jacobian perturbation.

The production change was validated on clean OpenFAST v5.0.0 source revision `2895884d2be01862173c88d70f86b358d2f1a50a` at one operating point with 36 azimuth snapshots. At the maintainers' request, this PR targets `rc-5.0.1` for the v5.0.1 bugfix release.

With the original implementation, both propagation-direction state-input columns are zero and their feedthrough to turbine outputs is zero, while the InflowWind diagnostic velocity Jacobian still gives the expected lateral derivative. With the repair, the aeroelastic direction sensitivity is restored.

The repaired rotation perturbation was compared at the same perturbation size against an independent uniform-wind-angle finite-difference implementation. The direction `B` columns agreed exactly at saved precision and the maximum `D`-column difference was `2.7041e-13`. Saved operating states/inputs/outputs, `A`, `C`, and all non-direction `B`/`D` columns were unchanged at saved precision.

A reproducer containing the input files, 36 legacy/fixed `.lin` pairs, and a standalone comparison script is attached to [issue #3463](https://github.com/OpenFAST/openfast/issues/3463#issuecomment-5669627736).

Finite-difference step-size convergence was investigated separately and is not part of this patch.

**Generative AI usage**

OpenAI Codex was used to assist with implementation and validation tooling. OpenAI ChatGPT was used to assist with preparation of the reproducer and PR text.

Validated on the `rc-5.0.1` PR branch at commit `aa22ddca9`, built with
Intel ifx 2026.0.0 in double precision.

The direction-linearization reproducer was rerun with the PR executable at
36 azimuth snapshots. The repaired model restores nonzero propagation-direction
`B` and turbine-output `D` sensitivity. Saved operating points, `A`, `C`,
and all non-direction `B`/`D` columns remain unchanged at the exported precision.

Existing linearization regression tests were also run. `WP_Stationary_Linear`
passes. The remaining affected linearization cases differ only in the
propagation-direction Jacobian columns: the current r-test references contain
the pre-fix zero sensitivity, while the PR produces the corrected nonzero
response. Modal frequencies/damping and all non-direction entries remain
unchanged.

An r-test reference update is therefore required to reflect the intended
propagation-direction Jacobian change.


andrew-platt (approved):
Looks good!  Thanks!

andrew-platt:
The linearization results with AeroDyn do change, which is expected.  I will merge this PR, then add a commit on `rc-5.0.1` updated .lin results files.

So this merge will fail regression testing, but will be fixed immediately afterwards.

